Issue #2, opened 2023-07-24
Replace 'dog' with 'pet' throughout codebase


Thrillberg:
This codebase was originally intended only for dog adoption but at Ruby for Good, we're extending it to apply to any pets that can be adopted. The task here is to replace all instances of 'dog' with 'pet'. Global find might come in handy.

PR #6, opened 2023-07-24
Replace dog with pet

egemen-dev:
In this PR, I have replaced the word 'dog' with 'pet' throughout the codebase.

All instances of the word 'dog' in class names, file names, variable names, route names, and any other relevant places have been replaced with the word 'pet'.

I ran the tests, and they all passed successfully. Additionally, I browsed the app to check for any issues that the tests might have missed, and I found none.

### Link to the issue
https://github.com/rubyforgood/pet-rescue/issues/2


Thrillberg (approved):
This looks great, @egemen-dev ! I wonder if there will be any bits of language that don't make sense for all pets and were obviously suited for dogs specifically. But even if so I think this is fine to merge and we can change those spots later on, as we find them. I did a cursory look through the changes and nothing popped out at me.

PR #7, opened 2023-07-24
Annotate models to assist developers in understanding schema

edwinthinks:

This PR adds the [annotate_model](https://github.com/ctran/annotate_models) gem which adds annotations to models and model related files. I've found this to be very helpful to understand the structure of database tables attached to models without having to dig into the schema.

Note - Annotations are updated automatically on migrations!

### Link to the issue

N/A


.tool-versions
```diff
@@ -0,0 +1 @@
+ruby 3.1.1
```
edwinthinks, line 1: Added this so those asdf users to manage their ruby.

Thrillberg (commented):
Awesome! 🤩 This is a great idea, @edwinthinks! Do you mind pulling the latest `main` and running the gem again? We changed `dog` to `pet` so some tables got changed by that.

app/models/pet.rb
```diff
@@ -1,4 +1,31 @@
 class Pet < ApplicationRecord
+  # == Schema Information
+  #
+  # Table name: dogs
```
edwinthinks, line 4: Oddly for some reason its still picking up the old annotation with dogs. Should be pets

edwinthinks: I believe this is caused by the state of the database still holding dogs as a table. I think adjusting our migrations should fix this.

Thrillberg (approved):
Looks great to me, @edwinthinks ! I see CI is failing so I'm re-running it now. If it passes, let's merge this!

PR #9, opened 2023-07-25
Utilize Dartscss over sass-rails modernization of tooling

edwinthinks:

This PR replaces `sass-rails` with `dartsass-rails` so that we can use a more future-proof (up to date) way of transpiling scss in rails. In addition, this adds the use of `foreman` to handle the processes for starting the server and re-compile changes made on the sass files.

To test you should follow the updated README instructions:
- Run `gem install foreman`
- And use `bin/dev` instead of rails s to start your server

### Link to the issue

N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@
 
 # Ignore SimpleCov test coverage metrics
 /coverage
+
+/app/assets/builds/*
+!/app/assets/builds/.keep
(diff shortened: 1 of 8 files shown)


Thrillberg (approved):
This looks great to me! 👍 My only hesitation is using `bin/dev` rather than `rails s` to start the server. This might trip some folks up. Is there a way to keep `rails s` as the command to start up the server or is that not worth doing and we'll just have to flag for people that the command is `bin/dev`?

PR #38, opened 2023-07-28
Add the postgres version to asdf as according to the old README.md

albertchae:
https://github.com/rubyforgood/pet-rescue/blob/41bfb59ab26649b172b683a86a285785445d9465/README.md?plain=1#L45


Thrillberg (approved):
Love it!

PR #56, opened 2023-07-28
Update setup script to include seeding

edwinthinks:

Includes seeding inside the bin/setup command via using db:setup instead of db:prepare

### Link to the issue

N/A


Thrillberg (approved):
lol I'm kind of confused between `rails db:prepare` and `rails db:setup` but I _think_ this looks right. Thank you for taking this on @edwinthinks !

PR #65, opened 2023-07-28
Deployment to Cloud66

edwinthinks:
### Link to the issue


robbkidd:
It's working!

Mostly!

config/credentials/staging.yml.enc
```diff
@@ -0,0 +1 @@
+TqDtQ9IzBRPXSxeuPJg2eRz8S9BC7SfW0H2YfD4NQijvl6/Lbz+YUx5BpSOEDK0d1vrmPPrK9e+rnE384DKLpPHniEF4jAkBXWtqifbiEG/f5kfZwogAOJ5cK13BjjvzRge0ab8vdSDnUOhJeJveWzrPdJNPiXd/l7I5zfKJd8C7W5Fl00BrhcKpv2TWcEnnLubjgmme6Z75zfUXoGrzlS+2uHx6UmUa2Q==--KsXxjt7p3xQORgHK--E5BwlXWQLmeYxeEYRfcnMw==
```
edwinthinks, line 1: @robbkidd when you get a chance could we review this? We couldn't get the Azure blob working so we put it in the crendentials.yml.enc. Maybe we undid something you did on purpose, we going to merge this in so we can deploy changes on main.

Thrillberg (approved):
🚀

config/credentials/staging.yml.enc
edwinthinks: I think I know why you made this call now. If we have a nonspecific credentials, then it forces all our people to have the master.key which is no bueno.

robbkidd: Yea, that was my thinking. Environment specific credentials so that ...

1. dev and test _aren't_ encrypted
2. production and staging use different credentials and each decrypts with a different SECRET_MASTER_KEY

PR #67, opened 2023-07-28
Fix static_pages_test

albertchae:
Not sure why it's broken without this change, but this is the error we're seeing

```
<We rescue pets in Baja, Mexico and rehome them with adopters in
      Canada and America> expected but was
<We rescue pets in Baja, Mexico and rehome them with adopters in
      Canada and America>..
```

### Link to the issue


garyhtou (approved):
Passing for me now! yayy!!

image

PR #70, opened 2023-07-28
Run system tests in CI using cuprite to avoid dependency on selenium

edwinthinks:

To ensure that our tests run without issues and too much overhead, this PR adds https://github.com/KonnorRogers/evil_systems which uses sane defaults for managing system tests. The top level benefits are:

- Cuprite, should be faster in many cases than the selenium counter part
- You don't need to worry about webdrivers

### Link to the issue

N/A


README.md
```diff
@@ -53,6 +53,28 @@ Staff
...
+Run system tests only (Not-Headless)
+```
+CI=false ./bin/rails test:system
+```
+
+Run ALL tests:
+```
+./bin/rails test:all
+```
+
```
Thrillberg, line 77: Thanks for updating the README!

Thrillberg (approved):
I haven't used `cuprite` before so I don't understand all of this but if it works 👍 !

PR #78, opened 2023-07-29
Remove fixture files after tests are run

Eduardo06sp:

Those with a keen eye may have noticed this folder getting bigger & bigger as tests are run: `tmp/storage_fixtures`

In the integration tests `org_pets_test.rb`, there are tests that use these attachment fixtures regarding uploading/deleting images.

What this commit does is clean the files up after all the tests are run, essentially wiping the `storage_fixtures` folder altogether, instead of adding fixture files to that folder indefinitely.


Thrillberg (approved):
This looks good to me! I think we may want to move away from fixtures in general in the future but for now, this seems like a good idea. Thank you!!

Issue #76, opened 2023-07-28
New checklist related tables


nsiwnf:
1. Create a `checklist_template` table for dynamically creating a `checklist` for a `match` 
 see #73 

2. Create a `checklist_assignments` table for tracking the events associated with a `match` ex. vaccination reminder, adoption event reminder. 
 -`checklist_assignment`
 -- `description`
 -- `due_date`
 -- `status`

PR #86, opened 2023-07-29
scaffold checklist template

aileenchoi:
### Link to the issue [76](https://github.com/rubyforgood/pet-rescue/issues/76)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -430,6 +430,7 @@ GEM
     zeitwerk (2.6.9)
 
 PLATFORMS
+  arm64-darwin-21
   arm64-darwin-22
   x86_64-darwin-22
   x86_64-linux
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -430,7 +430,6 @@ GEM
     zeitwerk (2.6.9)
 
 PLATFORMS
-  arm64-darwin-21
   arm64-darwin-22
   x86_64-darwin-22
   x86_64-linux
--- a/app/controllers/checklist_templates_controller.rb
+++ b/app/controllers/checklist_templates_controller.rb
@@ -1,9 +1,7 @@
 class ChecklistTemplatesController < InheritedResources::Base
-
   private
 
-    def checklist_template_params
-      params.require(:checklist_template).permit(:name, :description)
-    end
-
+  def checklist_template_params
+    params.require(:checklist_template).permit(:name, :description)
+  end
 end
--- a/app/controllers/checklist_templates_controller.rb
+++ b/app/controllers/checklist_templates_controller.rb
@@ -3,7 +3,7 @@ class ChecklistTemplatesController < InheritedResources::Base
   private
 
     def checklist_template_params
-      params.require(:checklist_template).permit(:name, :string, :description, :text)
+      params.require(:checklist_template).permit(:name, :description)
     end
 
 end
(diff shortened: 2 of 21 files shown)


Gemfile.lock
```diff
@@ -386,6 +386,7 @@ GEM
     zeitwerk (2.6.0)
 
 PLATFORMS
+  arm64-darwin-21
```
Thrillberg, line 389: Do we need this? Maybe take it out?

app/controllers/checklist_templates_controller.rb
```diff
@@ -0,0 +1,9 @@
+class ChecklistTemplatesController < InheritedResources::Base
+
+  private
+
+    def checklist_template_params
+      params.require(:checklist_template).permit(:name, :string, :description, :text)
```
Thrillberg, line 6: I don't think `.permit` works this way, ie: you only need to list the field names (`:name` and `:description`) and not the types (`:string` and `:text`).

app/helpers/checklist_templates_helper.rb
```diff
@@ -0,0 +1,2 @@
+module ChecklistTemplatesHelper
+end
```
Thrillberg, line 2: Do we need this if it's empty?

app/models/checklist_template.rb
```diff
@@ -0,0 +1,2 @@
+class ChecklistTemplate < ApplicationRecord
+end
```
Thrillberg, line 2: I *think* if we add a table to the db for checklist templates (which we should), then our `annotate` gem will automatically populate the top of this file with some useful db field info.

app/views/checklist_templates/_checklist_template.html.erb
```diff
@@ -0,0 +1,22 @@
+<div id="<%= dom_id checklist_template %>">
+  <p>
+    <strong>Name:</strong>
+    <%= checklist_template.name %>
+  </p>
+
+  <p>
+    <strong>String:</strong>
+    <%= checklist_template.string %>
```
Thrillberg, line 9: Oh maybe I was incorrect above. Is `string` a field here? Same question below for `text`.

app/controllers/checklist_templates_controller.rb
```diff
@@ -0,0 +1,9 @@
+class ChecklistTemplatesController < InheritedResources::Base
```
Thrillberg, line 1: Is this piggybacking off active admin? I think our intention is to remove active admin.

config/routes.rb
```diff
@@ -1,4 +1,5 @@
 Rails.application.routes.draw do
+  resources :checklist_templates
```
Thrillberg, line 2: Do we want to be able to use all actions?

db/migrate/20230729134738_create_checklist_templates.rb
```diff
@@ -0,0 +1,9 @@
+class CreateChecklistTemplates < ActiveRecord::Migration[7.0]
+  def change
+    create_table :checklist_templates do |t|
+      t.string :name
+      t.string :description
```
Thrillberg, line 5: Did you run this migration? It should update our `schema.rb` file.

Thrillberg (commented):
This is looking great! I have a few questions inside and would be happy to chat further about it if that would be helpful!

Thrillberg (approved):

PR #92, opened 2023-07-29
Added uniqueness to the Dog::Name Faker data

heyapricot:
Db Seed was failing because of repeated Dog Names

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -218,7 +218,7 @@ path = Rails.root.join("app", "assets", "images", "hero.jpg")
 10.times do
   pet = Pet.create!(
     organization: Organization.all.sample,
-    name: Faker::Creature::Dog.name,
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -218,7 +218,7 @@
 10.times do
   pet = Pet.create!(
     organization: Organization.all.sample,
-    name: Faker::Creature::Dog.name,
+    name: Faker::Creature::Dog.unique.name,
```
edwinthinks, line 221: We may not want to use `.unique` here. It slows it down a bit and we really want to remove the uniqueness

garyhtou:
https://github.com/rubyforgood/pet-rescue/pull/97 was recently merged! maybe that might help with this issue?

Thrillberg:
I think @garyhtou is right that this is resolved by #97 . @heyapricot if this is still an issue, feel free to reopen! :)

PR #97, opened 2023-07-29
Remove unique index on pet name

albertchae:
We removed the ActiveRecord validation in https://github.com/rubyforgood/pet-rescue/pull/58 but not the database constraint. This PR removes the DB constraint too.

One consequence is database seeding could fail due to Faker generating the same name.

### DIFF
--- a/app/models/pet.rb
+++ b/app/models/pet.rb
@@ -18,7 +18,6 @@
 #
 # Indexes
 #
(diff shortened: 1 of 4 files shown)


garyhtou (approved):
that works! thank you!

The failing test seems to be unrelated to this PR

PR #102, opened 2023-07-29
Improve ease of testing with shoulda /w some new model tests!

edwinthinks:

Adds https://github.com/thoughtbot/shoulda-matchers which makes it really easy to generate common syntax for very common tests. This should help us add testing because it "SHOULD" be easier (catch the punn :)?)

### Link to the issue

N/A

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -127,8 +127,8 @@ group :test do
 
   # Adds really common matchers you can use in tests to add
   # test coverage easily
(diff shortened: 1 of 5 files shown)


Thrillberg (approved):
Looks great to me! And thanks for adding some more tests to demonstrate how we SHOULD be using these matchers 😉

PR #109, opened 2023-07-29
Removed ability to create staff acount

heyapricot:
- Removed account_select route
- Removed the account_select shared partial
- Removed the conditionals that required account type from the new users registration view
- Changed caption and route of Create account links to 'Adopt'. They point to the  new user registration path instead of the account select path
- Removed Staff emails from the send_email method in the registrations controller

### DIFF
--- a/app/views/adoptable_pets/show.html.erb
+++ b/app/views/adoptable_pets/show.html.erb
@@ -146,7 +146,7 @@
                   <h4 class="mb-4 mt-3">
                     Create an account to apply for this pet
                   </h4>
-                  <%= link_to 'Create Account', account_select_path , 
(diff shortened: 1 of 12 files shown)


app/views/adoptable_pets/show.html.erb
```diff
@@ -146,7 +146,7 @@
                   <h4 class="mb-4 mt-3">
                     Create an account to apply for this pet
                   </h4>
-                  <%= link_to 'Create Account', account_select_path , 
+                  <%= link_to 'Adopt', new_user_registration_path ,
```
Thrillberg, line 149: Love this! Thank you for changing the link text!

Thrillberg (approved):
Thanks! Looks great to me!

PR #110, opened 2023-07-29
Proof of concept - chatting between staff & adopters

edwinthinks:

Wanted to see what I whip up real quick based on our conversation with Danielle about maybe wanting to chat with adopters and staff. Maybe in the future include foster as well.

The main point of this is to see if we want to pursue more time to refining this feature.

https://www.loom.com/share/5ee3b313a00641cdb53412d5405ea9a7

### Link to the issue

N/A

### DIFF
--- a/app/controllers/adopter_applications/chats_controller.rb
+++ b/app/controllers/adopter_applications/chats_controller.rb
@@ -1,4 +1,6 @@
 class AdopterApplications::ChatsController < ApplicationController
   def show
+    @match = AdopterApplication.find(params[:adopter_application_id])
+    @messages = @match.messages.includes(:sender)
   end
 end
--- a/app/controllers/adopter_applications/chats_controller.rb
+++ b/app/controllers/adopter_applications/chats_controller.rb
@@ -1,0 +1,4 @@
+class AdopterApplications::ChatsController < ApplicationController
+  def show
+  end
+end
--- a/app/javascript/controllers/chat_message_controller.js
+++ b/app/javascript/controllers/chat_message_controller.js
@@ -1,0 +1,12 @@
+import { Controller } from "@hotwired/stimulus"
+
+/**
+ * Add mutation to move the scroll to the bottom of the chat when
+ * it is updated.
+ */
+
+export default class extends Controller {
+  connect() {
+    this.element.scrollIntoView()
+  }
+}
--- a/test/models/message_test.rb
+++ b/test/models/message_test.rb
@@ -1,0 +1,7 @@
+require "test_helper"
+
+class MessageTest < ActiveSupport::TestCase
+  # test "the truth" do
+  #   assert true
+  # end
+end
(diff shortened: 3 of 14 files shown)


app/controllers/adopter_applications/chats_controller.rb
```diff
@@ -0,0 +1,6 @@
+class AdopterApplications::ChatsController < ApplicationController
+  def show
+    @match = AdopterApplication.find(params[:adopter_application_id])
```
Thrillberg, line 3: Why are we using `AdopterApplication` instead of `Adoption` here? Seems intuitive to use `Adoption`.

app/javascript/controllers/chat_message_controller.js
```diff
@@ -0,0 +1,12 @@
+import { Controller } from "@hotwired/stimulus"
+
+/**
+ * Add mutation to move the scroll to the bottom of the chat when
+ * it is updated.
+ */
+
+export default class extends Controller {
+  connect() {
+    this.element.scrollIntoView()
```
Thrillberg, line 10: This is rad!

test/models/message_test.rb
```diff
@@ -0,0 +1,7 @@
+require "test_helper"
+
+class MessageTest < ActiveSupport::TestCase
+  # test "the truth" do
+  #   assert true
```
Thrillberg, line 5: yeah but is the truth really true 🤔

Thrillberg (approved):
Looks great! Where in the logic does it determine who can see the messages?

app/controllers/adopter_applications/chats_controller.rb
edwinthinks: Good question, I think `Adoption` makes more sense as you said. I don't have any strong ties to the choice to use AdopterApplication. Whatever we use, it just needs to be related to the Adopter & Staff.

app/javascript/controllers/chat_message_controller.js
edwinthinks: Jeje it is good, but actually has some issues -- which it scrolls down the entire page rather than just the internal element. I'd probably use something like -- https://github.com/stimulus-use/stimulus-use/blob/main/docs/use-mutation.md to control both the scrolling down and moving the element to the left or right depending on who sent the message.

test/models/message_test.rb
edwinthinks: Who knows, in real life this assertion is actually flaky lol

edwinthinks:
> Looks great! Where in the logic does it determine who can see the messages?

There is no logic yet for this since its a proof of concept. However, the way to do it later would be to put authorization at the controller level to prevent access to the page. Within the page adding `turbo_stream_from` handles broadcasting/subscribing.

edwinthinks:
@Thrillberg thanks for the review! I'am not actually planning on merging this until we have some sort of UI figured out. Then we can re-open and maybe let another use this as a reference to implement. Great for any gooders who want to learn hotwire in a live setting!

PR #111, opened 2023-07-29
Scaffold Checklist Template Items

aileenchoi:
### [76](https://github.com/rubyforgood/pet-rescue/issues/76)

### DIFF
--- a/app/models/checklist_template.rb
+++ b/app/models/checklist_template.rb
@@ -9,5 +9,5 @@
 #  updated_at  :datetime         not null
 #
 class ChecklistTemplate < ApplicationRecord
-    has_many :items, class_name: "ChecklistTemplateItem", dependent: :destroy
+  has_many :items, class_name: "ChecklistTemplateItem", dependent: :destroy
 end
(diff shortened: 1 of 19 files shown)


app/models/checklist_template_item.rb
```diff
@@ -0,0 +1,25 @@
...
+#
+#  index_checklist_template_items_on_checklist_template_id  (checklist_template_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (checklist_template_id => checklist_templates.id)
+#
+class ChecklistTemplateItem < ApplicationRecord
+  belongs_to :checklist_template
+  validates :name, :expected_duration_days, presence: true
```
albertchae, line 24: should we check that `expected_duration_days` is positive?

app/models/checklist_template.rb
```diff
@@ -9,4 +9,5 @@
 #  updated_at  :datetime         not null
 #
 class ChecklistTemplate < ApplicationRecord
+  has_many :items, class_name: "ChecklistTemplateItem", dependent: :destroy
```
albertchae, line 12: minor but when comparing to class names, I like to do `ChecklistTemplateItem.name` because it guarantees the class exists and is loadable before getting the string name out of it. It's technically less efficient than a frozen string but I think is a bit more defensive against typos. I have heard arguments against this too though, so not a big deal either way.

albertchae (approved):
looks good, just some non blocking comments

PR #113, opened 2023-07-29
Denormalize organization_id onto Match

albertchae:
I wanted to add `acts_as_tenant(:organization, through: :pet)` to the
Match model, but it looks like `through` is only supported for HABTM
associations and not belongs_to/has_one. This seems to be intentional
design in acts_as_tenant https://github.com/ErwinM/acts_as_tenant/issues/293#issuecomment-1238567771

So I added `organization_id` to `Match`, but a risk of doing that is
when there's a discrepancy between the denormalized tenant_id and the
tenant_id reachable through the object graph. In this case that would be
when `match.organization_id` does not match `match.pet.organization_id`.

My preferred way of dealing with that is adding a composite foreign key
including the tenant_id so that you can use the database to establish and
guarantee "chain of trust" up to the tenant. Here that would look like
```
ALTER TABLE matches
ADD CONSTRAINT fk_pet_id_organization_id_to_pets
FOREIGN KEY (pet_id, organization_id)
REFERENCES pets (id, organization_id);
```

(N.B. we also need `add_index :pets, [:id, :organization_id], unique: true` because
composite foreign keys require uniqueness on the set of columns they're
pointing to).

We have to use raw DDL because rails currently doesn't support composite foreign keys
in their migration helpers. However, an issue with this approach is that
the `schema.rb` dump doesn't support foreign keys. We'd get something
like `add_foreign_key "matches", "pets", name: "fk_pet_id_organization_id_to_pets"` which
has no mention of the columns that make up the composite key. We could
get around this by changing this project to use a `structure.sql` schema
dump
https://blog.appsignal.com/2020/01/15/the-pros-and-cons-of-using-structure-sql-in-your-ruby-on-rails-application.html,
but I don't think we want to make this change now. So I instead just
added an ActiveRecord validation to guarantee the organization_id
matches.

P.S. it looks like native rails support for composite keys is coming https://github.com/rails/rails/issues/47593

### DIFF
--- a/app/models/match.rb
+++ b/app/models/match.rb
@@ -6,11 +6,13 @@
 #  created_at         :datetime         not null
 #  updated_at         :datetime         not null
 #  adopter_account_id :bigint           not null
+#  organization_id    :bigint           not null
 #  pet_id             :bigint           not null
 #
 # Indexes
 #
 #  index_matches_on_adopter_account_id  (adopter_account_id)
+#  index_matches_on_organization_id     (organization_id)
 #  index_matches_on_pet_id              (pet_id)
 #
 # Foreign Keys
@@ -21,10 +23,21 @@
 class Match < ApplicationRecord
   belongs_to :pet
   belongs_to :adopter_account
+  belongs_to :organization
+
+  validate :belongs_to_same_organization_as_pet
 
   after_create_commit :send_checklist_reminder
 
   def send_checklist_reminder
     MatchMailer.checklist_reminder(self).deliver_later
+  end
+
+  private
+
+  def belongs_to_same_organization_as_pet
+    if organization_id != pet.organization_id
+      errors.add(:organization_id, "must belong to same organization as pet")
+    end
   end
 end
--- a/test/models/match_test.rb
+++ b/test/models/match_test.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class MatchTest < ActiveSupport::TestCase
+  def test_match_must_belong_to_same_organization_as_pet
+    org1 = Organization.create!
+    org2 = Organization.create!
+    pet = Pet.create!(organization: org1,
+      name: "Harry", age: 5, size: "small", breed: "corgi", sex: "m", description: "test pet")
+    match = Match.create(pet: pet, organization_id: org2.id,
+      adopter_account: adopter_accounts(:adopter_account_one))
+    assert !match.valid?
+    assert match.errors[:organization_id].present?
+  end
+end
--- a/app/controllers/matches_controller.rb
+++ b/app/controllers/matches_controller.rb
@@ -2,7 +2,8 @@ class MatchesController < ApplicationController
   before_action :verified_staff, :same_organization?
 
   def create
-    @match = Match.new(adoption_params)
+    @pet = Pet.find(adoption_params[:pet_id])
+    @match = Match.new(adoption_params.merge(organization_id: @pet.organization_id))
 
     if @match.save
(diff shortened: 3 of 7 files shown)


test/models/match_test.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class MatchTest < ActiveSupport::TestCase
```
albertchae, line 5: I don't use minitest so not sure what best practices are, please comment away

test/models/match_test.rb
```diff
@@ -0,0 +1,16 @@
...
+  def test_match_must_belong_to_same_organization_as_pet
+    org1 = Organization.create!
+    org2 = Organization.create!
+    pet = Pet.create!(organization: org1,
+      name: "Harry", age: 5, size: "small", breed: "corgi", sex: "m", description: "test pet")
+    match = Match.create(pet: pet, organization_id: org2.id,
+      adopter_account: adopter_accounts(:adopter_account_one))
```
albertchae, line 12: I prefer to create test data in the test instead of using fixtures and I think we discussed doing that project wide. Ideally we could use something like factory_bot to fill in the irrelevant details of the model creation that aren't related to the test (in this case everything that isn't organization related). Unfortunately I didn't want to create all the stuff around `adopter_account` so I just reused the fixture in that case

app/models/match.rb
```diff
@@ -21,10 +23,21 @@
...
   after_create_commit :send_checklist_reminder
 
   def send_checklist_reminder
     MatchMailer.checklist_reminder(self).deliver_later
   end
+
+  private
+
+  def belongs_to_same_organization_as_pet
+    if organization_id != pet.organization_id
```
Thrillberg, line 39: Nice! I like this "hack" for ensuring that the data lines up.

test/models/match_test.rb
Thrillberg: I agree that this kind of setup for tests is preferable.

test/models/match_test.rb
Thrillberg: Nor do I but this looks great to me!

test/models/match_test.rb
```diff
@@ -0,0 +1,16 @@
...
+    assert !match.valid?
```
Thrillberg, line 13: Feels like there should be something more idiomatic for asserting a falsy value here but I'm not finding it on a quick google search. Oh well! This looks great as-is.

Thrillberg (approved):
Fabulous work! Thanks for doing all that digging into the composite foreign key business.

test/models/match_test.rb
albertchae: I agree and same experience here. I'll merge for now and ask if anyone knows better

robbkidd: `refute()` [is the opposite of](https://ruby-doc.org/3.1.2/gems/minitest/Minitest/Assertions.html#method-i-refute) `assert()`, but I'll confess I don't see it used all that often so I can't argue for its idiomacity†.

```suggestion
    refute match.valid?
```

† Which is totally a word I just made up.

robbkidd: [Minitest::Expectations](http://docs.seattlerb.org/minitest/Minitest/Expectations.html#method-i-wont_be) are also available.

```suggestion
    _(match).wont_be :valid?
```

Thrillberg: Sounds like `refute()` is about as idiomatic to Ruby as `idiomacity` is idiomatic to English.

robbkidd: I cannot refute that conclusion.

Issue #90, opened 2023-07-29
Enter birthdate for pets instead of saving age as a string


Thrillberg:
Currently we save a string for a pet's age. This will get out of date. Please prompt the user to enter a birthdate instead but be sure to add verbiage about letting them be approximate (not precise to the day since many people might not know the date itself).

PR #121, opened 2023-07-30
Add birth_date to Pet and compute age dynamically

albertchae:
Closes https://github.com/rubyforgood/pet-rescue/issues/90

Currently age is a static form field, so if a month passes in real time, the pet's age stays the same.

Ruby (https://bugs.ruby-lang.org/issues/14877) nor ActiveSupport have a
way to compute the age given a birthday and most answers online are
concerned with human ages which tend to only be in years. ~~Since it seems
like we want to include months, especially if the pet is less than 1
year old, I added a custom function for computing and display that. It
isn't technically accurate because we compute it off the duration from
today and the birthdate, but should be acceptable because birthdates in
the system will typically be approximate anyway.~~

Switched this to use https://api.rubyonrails.org/classes/ActionView/Helpers/DateHelper.html#method-i-time_ago_in_words, which is still approximate but should serve our purposes. The only difference is that instead of always saying "exact" years and months, it will sometimes use "about", "almost", "over".

### DIFF
--- a/app/models/pet.rb
+++ b/app/models/pet.rb
@@ -6,6 +6,7 @@
 #  age                :integer
 #  age_unit           :integer          default("months")
 #  application_paused :boolean          default(FALSE)
+#  birth_date         :datetime
 #  breed              :string
 #  description        :text
 #  name               :string
@@ -32,8 +33,9 @@ class Pet < ApplicationRecord
   has_many_attached :images
 
   validates :name, presence: true
-  validates :age, presence: true
-  validates :age_unit, presence: true
+  validates :birth_date, presence: true
+  #validates :age, presence: true
+  #validates :age_unit, presence: true
   validates :size, presence: true
   validates :breed, presence: true
   validates :sex, presence: true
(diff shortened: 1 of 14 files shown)


app/models/pet.rb
```diff
@@ -75,6 +77,28 @@ def append_images=(attachables)
...
+
+  def approximate_age_display
+    approximate_age => {years:, months:}
+
+    years_string = years.positive? ? "#{years} years" : nil
+    months_string = months.positive? ? "#{months} months" : nil
+
+    # using compact and join will correctly handle the case when either years or months is 0
+    [years_string, months_string].compact.join(', ')
+  end
```
albertchae, line 100: @edwinthinks suggested IRL to use https://api.rubyonrails.org/classes/ActionView/Helpers/DateHelper.html#method-i-distance_of_time_in_words, which doesn't produce the exact same output, but since this is all approximate anyway I think it's worth switching to it

edwinthinks (approved):

db/migrate/20230730134633_add_birthdate_to_pets.rb
```diff
@@ -0,0 +1,5 @@
+class AddBirthdateToPets < ActiveRecord::Migration[7.0]
+  def change
+    add_column :pets, :birth_date, :datetime
```
edwinthinks, line 3: Could we add a not null constraint here?

albertchae: done

PR #123, opened 2023-07-30
Add model tests

### LINKED ISSUE #105: Add basic model unit tests for associations & validations
Let's expand our test coverage as much as we think we can using shoulda following the https://github.com/rubyforgood/pet-rescue/pull/102

Don't focus too heavily on being 100% covered, just the basics like associations / validation

Thrillberg:
Fixes #105

I've added some model tests on validations and relationships. I skipped models that I think we're going to remove like the `*_account` models.

### Link to the issue
#105

### DIFF
--- a/app/models/match.rb
+++ b/app/models/match.rb
@@ -25,7 +25,7 @@ class Match < ApplicationRecord
   belongs_to :adopter_account
   belongs_to :organization
 
-  validate :belongs_to_same_organization_as_pet
+  validate :belongs_to_same_organization_as_pet, if: -> { pet.present? }
 
   after_create_commit :send_checklist_reminder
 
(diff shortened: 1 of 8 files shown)


app/models/match.rb
```diff
@@ -25,7 +25,7 @@ class Match < ApplicationRecord
   belongs_to :adopter_account
   belongs_to :organization
 
-  validate :belongs_to_same_organization_as_pet
+  validate :belongs_to_same_organization_as_pet, if: ->{ pet.present? }
```
Thrillberg, line 28: This is a bugfix! Thanks testing (and @edwinthinks)

edwinthinks (approved):

Issue #126, opened 2023-07-31
Use factorybot instead of fixtures


Thrillberg:

Thrillberg:
Closed by #149

PR #127, opened 2023-07-31
Begin replacing fixtures with FactoryBot

Thrillberg:
This PR introduces [FactoryBot](https://github.com/thoughtbot/factory_bot_rails), which I think we'll want to move towards. I'm keeping the fixtures in the codebase for now because existing tests rely on it but over time hopefully we can replace our usage of fixtures with factories instead.

`test/integration/adopter_application_test.rb` shows a more complicated/usual example of how that may look. I don't think we should have all of our factories in one file as I do here but for now it's kind of convenient.

If this looks good I think we might as well merge it now and just keep it as an ongoing process in the background that we'll gradually replace the tests and hopefully someday soon retire the fixtures!

### Link to the issue
#126

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -92,6 +92,9 @@ group :development, :test do
 
   # Analysis for security vulnerabilities
   gem "brakeman"
+
+  # Creating factory instantiations in tests
+  gem "factory_bot_rails"
 end
 
 group :development do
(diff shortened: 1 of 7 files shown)


egemen-dev:
Looks good to me 💻

Issue #98, opened 2023-07-29
After profile create, redirect to browse pets


Thrillberg:
Instead of landing on the profile show page, go to browse pets. The more cuteness we overload the user with, the better!

egemen-dev:
Merged

PR #129, opened 2023-07-31
Redirect user to browse pets after adopter profile has been created.

egemen-dev:

Redirects the user to browse pets after the adopter profile has been created.

### Link to the issue
https://github.com/rubyforgood/pet-rescue/issues/98

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -37,7 +37,7 @@ class RegistrationsController < Devise::RegistrationsController
 
   # redirect new adopter users to adopter_profile#new
   def after_sign_up_path_for(resource)
-    resource.adopter_account ? new_profile_path : root_path
+    resource.adopter_account ? adoptable_pets_path : root_path
   end
 
   def after_sign_in_path_for
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -35,9 +35,8 @@ class RegistrationsController < Devise::RegistrationsController
       :current_password)
   end
 
-  # redirect new adopter users to adoptable pets page
   def after_sign_up_path_for(resource)
-    resource.adopter_account ? adoptable_pets_path : root_path
+    adoptable_pets_path
   end
 
   def after_sign_in_path_for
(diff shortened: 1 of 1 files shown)


app/controllers/registrations_controller.rb
```diff
@@ -35,9 +35,9 @@ def account_update_params
       :current_password)
   end
 
-  # redirect new adopter users to adopter_profile#new
+  # redirect new adopter users to adoptable pets page
   def after_sign_up_path_for(resource)
-    resource.adopter_account ? new_profile_path : root_path
+    resource.adopter_account ? adoptable_pets_path : root_path
```
egemen-dev, line 40: This is a valid path. But I got some questions.
Are we supposed to create the `adopter_account` from the admin dashboard?
I am not sure where to create the `adopter_account` since `user has_one :adopter_account, dependent: :destroy` and there seems to be no after callback to create an `adopter_account` for the user.

app/controllers/registrations_controller.rb
```diff
@@ -35,9 +35,9 @@ def account_update_params
       :current_password)
   end
 
-  # redirect new adopter users to adopter_profile#new
+  # redirect new adopter users to adoptable pets page
```
edwinthinks, line 38:
```suggestion
```
I personally think this comment isn't needed

app/controllers/registrations_controller.rb
edwinthinks: Is the goal of this to make it so only adopters go into the browser page versus the staff? Honestly... I would just point to `adoptable_pets_path` without any logic to try to handle staff & adopter differently. Internally we've talked about coming back to address the staff post login experience.

What ya think?

edwinthinks (changes requested):
Left some questions about this, lemme know if you have questions.

app/controllers/registrations_controller.rb
egemen-dev: I agree. I will change `resource.adopter_account ? adoptable_pets_path : root_path` to `adoptable_pets_path`

edwinthinks: Excellent! Thanks for hearing me out :)

edwinthinks (approved):

PR #131, opened 2023-07-31
update to Ruby 3.1.4

robbkidd:

* Updates the version of Ruby to run against in the various places that it is declared.
* Updates CI to use a version of the ruby/setup-ruby Action that is aware of v3.1.4's existence.


.github/workflows/rubyonrails.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         uses: actions/checkout@v3
       # Add or replace dependency steps here
       - name: Install Ruby and gems
-        uses: ruby/setup-ruby@0a29871fe2b0200a17a4497bae54fe5df0d973aa # v1.115.3
+        uses: ruby/setup-ruby@v1.152.0
```
robbkidd, line 34: Update to a version of the Action that is [aware of Ruby v3.1.4's existence](https://github.com/ruby/setup-ruby/commit/904f3fef85a9c80a3750cbe7d5159268fd5caa9f).

Alternatively, this could go with the Action maintainer's [emphatic recommendation](https://github.com/ruby/setup-ruby#versioning) to ride the `v1` tag to always run the latest version of the Action:

```suggestion
        uses: ruby/setup-ruby@v1
```

Thrillberg (approved):
Let's doooo iiiiitttt!! Seems better to just keep up with `v1` in the GitHub action, right?

edwinthinks (approved):
Go go go!

PR #132, opened 2023-07-31
add more scripts for testing convenience

robbkidd:
### Changes

- `bin/test` does [what CI currently does](https://github.com/rubyforgood/pet-rescue/blob/be255ee86ec83c838cd2149d494dac518a9747b8/.github/workflows/rubyonrails.yml#L38-L48)
- `bin/audit` [runs security things](https://github.com/rubyforgood/pet-rescue/blob/be255ee86ec83c838cd2149d494dac518a9747b8/.github/workflows/rubyonrails.yml#L61-L72) (only brakeman for the moment)
- `bin/style` [runs the style checker](https://github.com/rubyforgood/pet-rescue/blob/be255ee86ec83c838cd2149d494dac518a9747b8/.github/workflows/rubyonrails.yml#L49-L60) (standardrb)


robbkidd:
I found using `bin/` scripts convenient so that:
1. I didn't have to set RAILS_ENV to "test" which I never seem to remember to do.
2. With a single command, I could run the same series of test steps as are run in CI .

These could instead be named Rake targets instead of scripts.

However the single name is defined, I recommend doing it so that the definition is made in one place. Use it in CI, use it in dev. The easier it is to run what CI runs, the earlier new bugs and regressions are caught. How often do you push updates to a branch thinking things are green only to have CI bark at you because you forgot to run the style checker?

PR #134, opened 2023-07-31
Transition more to factorybot

Thrillberg:
This PR converts a few more tests off fixtures and into using FactoryBot. It also fixes a source of flakiness that I had put into the factories so THAT'S GOOD. :D

### Link to the issue
#126


nsiwnf (approved):
lgtm!

Issue #94, opened 2023-07-29
Change country, county and state to be dropdowns


Thrillberg:
These can be selected from a dropdown rather than free text entry.

egemen-dev:
Merged

PR #136, opened 2023-08-01
Implement country state dropdown

egemen-dev:
Added country and state dropdowns for adopter_profile.

Because users rather click than typing in the keyboard :stuck_out_tongue_winking_eye:

Preview:
drp

Note: This can be extended to other countries. ~~As of now, I just added Canada and Unites States.~~

**Edit: Canada, United States and Mexico are included.**
### Link to the issue
https://github.com/rubyforgood/pet-rescue/issues/94


app/javascript/controllers/country_state_controller.js
```diff
@@ -0,0 +1,46 @@
...
+export default class extends Controller {
+  static targets = ["country", "state"];
+
+  connect() {
+    this.selectedCountry = this.countryTarget.dataset.initialValue;
+    this.selectedState = this.stateTarget.dataset.initialValue;
+
+    this.setCountryDropdown();
+    this.setStateDropdown();
+  }
```
edwinthinks, line 13: I'd be curious to see if we can try this approach using hotwire to avoid writing JS - https://www.youtube.com/watch?v=ReZS6wfh3lo

I'am a pretty big advocate for less JS to maintain.

edwinthinks (changes requested):
I think we ought to consider keeping things simple. The idea to dynamically fill in the state selection is not necessary imo. I would consider using https://github.com/thecodecrate/city-state with a select field.

app/javascript/controllers/country_state_controller.js
egemen-dev: Totally understandable. At first, I thought about using a gem and turbo requests to solve the problem, like in the video you shared. However, after careful consideration, I decided not to add the gem. Instead, I chose to handle the task directly on the client side.

The reason for this decision is that managing the logic on the client side can be beneficial. It reduces the burden on the server and avoids unnecessary GET requests for better overall performance and responsiveness.

nsiwnf: Such a good use of hotwire! I think this would be a great V2 @edwinthinks. Im kind of inclined to move forward with the PR as is since it works as expected and is not too much js to maintain

egemen-dev: Thank you for detailed explanation. I love these convos too! 

> Every bit of code added is weight and come with a cost.

I understand your perspective. That is right. 

> To summarize, I think we should use the easiest solution possible which I believe is a select form with options being countries

Since the PR is done and all working, I suggest keeping this as is. This ensures that the data collected is more accurate. Also users spend less time filling in boring information.

I am trying to look at the topic from user's perspective. If I were using this app and could only select the country, I might wonder why the state dropdown was excluded. 

However, I am totally open about what ever we decide to do. Let me know what I should do.

app/javascript/helpers/locationHelper.js
```diff
@@ -0,0 +1,73 @@
+export const locationHelper = {
+  countries: ["Canada", "United States"],
```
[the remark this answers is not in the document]
egemen-dev: I was not sure which countries to add. So, for the start, I just added 2 countries. How many countries should we include?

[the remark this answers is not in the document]
edwinthinks: You could use this gem -- https://github.com/thecodecrate/city-state#all-countries-of-the-world but it wouldn't work with JS out of the box. You'd have to pass them in.

app/javascript/controllers/country_state_controller.js
nsiwnf: @egemen-dev @edwinthinks Yeah, Im in favor of the hotwire solution but also ok with submitting this PR (with the Mexico changes) as long as we can commit to a V2 in the short term - happy to pair on that!

app/javascript/helpers/locationHelper.js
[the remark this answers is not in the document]
egemen-dev: About the gem - I passed the instance variable (country and states) to the erb file and added it to the DOM using data attribute. Then used that data (states) in JS controller. It worked, but I had trouble with passing the tests. I couldn't understand why. So I did not use it now.

I manually added the Mexico and its states to JS helper like US and Canada. This works and passes the tests. You can also check it out to confirm.

I think in the long term we might want to use Hotwire (Turbo) just like in the Go Rails video you mentioned earlier.

@edwinthinks @nsiwnf

app/javascript/controllers/country_state_controller.js
edwinthinks: @nsiwnf sounds like a excellent issue! 

I'am also good with merging this in with the understanding we will revisit. Let's keep the momentum going!

edwinthinks (approved):
As mentioned, let's merge this in today and then come back to this using hotwire approach as @nsiwnf mentioned. Let's keep the momentum going!

nsiwnf:
Created a new issue for V2 https://github.com/rubyforgood/pet-rescue/issues/141

nsiwnf (approved):

egemen-dev:
Hey guys, I just tried to implement the turbo stream solution for https://github.com/rubyforgood/pet-rescue/issues/141 out of curiosity and I think it's done. I haven't pushed any changes to this PR yet. If you guys say OK, then I will commit changes and there will be no need to revisit this feature to implement turbo streams again. Idk, maybe another person would like to implement turbo streams in another PR. Let me know what do you guys think? @edwinthinks @nsiwnf

egemen-dev:
> Awesome! Yeah please creat a PR for those changes - what a quick turnaround zap

I was not happy that we were going to revisit this in the future. Why not do it once and use it for a long time? :)

Since this PR is not merged, maybe push changes here? @nsiwnf

config/routes.rb
```diff
@@ -7,7 +7,11 @@
 
   resources :adoptable_pets, only: [:index, :show]
   resources :adopter_applications, only: [:index, :edit, :update], controller: "adoption_application_reviews"
-  resource :adopter_profile, except: :destroy, as: "profile"
+  resource :adopter_profile, except: :destroy, as: "profile" do
+    collection do
+      get "states"
+    end
```
edwinthinks, line 13: I think you can make a general endpoint outside of the `adopter_profile` to generalize it so it can be used in other spots. Check my other comment about making this more RESTful

app/controllers/adopter_profiles_controller.rb
```diff
@@ -46,6 +46,12 @@ def update
     end
   end
 
+  def states
+    @target = params[:target]
+    @name = params[:name]
+    @states = CS.states(params[:country]).invert
```
edwinthinks, line 52: I think we can make a seperate controller called
CountriesController in which you can make this route:

```
GET /countries/:country/states
```
This is a bit more restful and communicates a bit more clearly what resource you'd expect. In this case, "states"

app/views/adopter_profiles/states.turbo_stream.erb
```diff
@@ -0,0 +1,3 @@
+<%= turbo_stream.replace @target do %>
+  <%= select_tag @name, options_for_select(@states), 'data-country-state-target': 'state', class: 'form-control' %>
+<% end %>
```
edwinthinks, line 3: So good!

config/routes.rb
```diff
@@ -14,6 +14,12 @@
   resources :profile_reviews, only: [:show]
   resources :successes, only: [:index]
 
+  resources :countries, only: [] do
```
egemen-dev, line 17: Current route:
```
states_countries GET  /countries/states(.:format)  countries#states
```
I don't know, can this be improved in `routes.rb` file. The [] is kind of weird.

app/views/adopter_profiles/_form.html.erb
```diff
@@ -44,8 +44,9 @@
       <%= form.fields_for :location do |location| %>
           <%= location.label :country %>
           <%= location.select :country,
-                              [],
+                              CS.countries.invert,
                               { prompt: "Please select" },
+                              'data-path': states_profile_path,
                               'data-country-state-target': 'country',
                               'data-action': 'change->country-state#updateStates',
                               'data-initial-value': location.object.country,
```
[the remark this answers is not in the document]
egemen-dev: Tried that but all in vain

config/routes.rb
edwinthinks: Yeh it looks strange, but its normal :)!

I would do
```
resources :countries, only: [] do
   resources :states, only: [:index] 
end
```
So you can get the countries/:country_name/states route which is the most RESTful.

test/controllers/countries_controller_test.rb
```diff
@@ -0,0 +1,24 @@
+require "test_helper"
+
+class CountriesControllerTest < ActionDispatch::IntegrationTest
+  setup do
+    sign_in users(:adopter_with_profile)
+  end
+
+  test "should return turbo stream with the states in it" do
```
edwinthinks, line 8: Great test!

config/routes.rb
edwinthinks: ohh I just thought of why this may extend well. Even though its currently `[]` for countries, imagine we wanted to introduce more countries? You could use countries#index to see what countries are available. Even more wild you could maybe toss a query parameter in there to filter it down by geolocation as well.

All of this is like scope creep and not worth doing now imo, but just sharing this to show where I think this could go in the future.

egemen-dev: Please see https://github.com/rubyforgood/pet-rescue/pull/136/commits/8ff991aff6869b34dac438f979ea10cf28fd7c5a for updated route - some things changed as well.

New route:
```
country_states GET  /countries/:country_id/states(.:format) states#index
```

edwinthinks (approved):
INCREDIBLE! 

Thank you for working on this and your openness to my comments. This is really good! I'am excited to see where we can implement more great things like this in the app. 

Looking forward to working with you more.

PR #137, opened 2023-08-01
Apply "Dashkit" theme to organize UI/UX 

edwinthinks:

This PR is an experiment that is used to demonstrate a path to handling a sustainable UI. Here are some high-level details on what I believe is going to make the UI sustainable, consistent, and nice looking:

- Discourage creating new CSS classes, try your best utilize the utility classes in Bootstrap. Using the utility approach forces us to use a common language. 
- Use "shared" components either using partials, nice_partials, or ViewComponents which act as reusable elements that handle the styling for you without knowing the internal details. 
- Try to avoid using patterns outside of the theme provided! Study the theme (https://dashkit.goodthemes.co/index.html used in this one) very closely to ensure it is the right move to break pattern.

By following these steps, I believe we can have very nice looking UI that is consistent and beautiful without the headaches and fear of changing CSS.

Here is a quick link to a demo:
https://jam.dev/c/09c97bda-3448-44f6-9b54-6658b780a9ca

### Link to the issue


edwinthinks:
Added this video to express a bit more of the ideas presented in this PR about frontend approach - https://www.loom.com/share/6a2a7ecb041145609c62a4b746476b6d

PR #142, opened 2023-08-02
Initial stab at switching out logos

### LINKED ISSUE #139: Make header logo switch on subdomain
Currently, the header logo always shows the Baja Pet Rescue logo regardless of which subdomain the visitor is on. Let's make it change to a different logo with a different subdomain.

Thrillberg:
**I do not necessarily endorse my solution here** but I figured pushing up some not-so-great code might be a good conversation-starter! 😆 Anyway, the goal here is to get the logos to switch per the subdomain, which works with this PR. Observe!

Screenshot 2023-08-01 at 9 49 55 PM

Not sure why I called this `alta`. I guess it's just a play on `baja`. But it's probably more confusing than it's worth!

Screenshot 2023-08-01 at 9 50 31 PM

Hey look, Sean has his own pet shelter!

Ok so that's cool but I'm not happy about how I implemented it. I need ideas from you all to make this better! As you can see in the code I've got this `get_organization_settings` hash that will undoubtedly grow to an obnoxious degree. How should we store the settings of our organizations? Should we start mapping that out now? Is it too early? (never!)

It's a bit buried in the diff because I decided to move the header out to its own partial but the organization settings get invoked like this:
```ruby
<%= link_to image_tag(@organization_settings[:logo], class: 'logo me-2', alt: 'site logo'), root_path %>
```

Anyway, what do you think?

### Link to the issue
Fixes #139


app/controllers/application_controller.rb
```diff
@@ -1,20 +1,31 @@
...
+  def get_organization_settings
+    {
+      "alta" => {
+        logo: "logo.png"
+      },
+      "rubyforgood" => {
+        logo: "rubyforgood-logo.png"
+      }
+    }
+  end
```
edwinthinks, line 27: Hmm maybe we could instead use a attachment on Organization with ActiveStorage. So like `organization.logo`

app/views/layouts/application.html.erb
```diff
@@ -16,121 +16,7 @@
...
-                    data: { turbo: false },
-                    method: 'delete' %>
-              </li>
-            <% end %>
-          </ul>
-        </div>
-      </div>
-    </nav>
-
+    <%= render "layouts/header" %>
```
edwinthinks, line 19: I would either pass in organization or create a helper called `current_organization` to fetch the organization to configure the page views.

app/views/layouts/_header.html.erb
```diff
@@ -0,0 +1,113 @@
+<nav class="navbar navbar-expand-lg navbar-dark"
+      style='background-color: #00401F;' aria-label="Fourth navbar example">
+  <div class="container">
+    <div class="navbar-brand">
+      <%= link_to image_tag(@organization_settings[:logo], class: 'logo me-2', alt: 'site logo'),
```
edwinthinks, line 5:
```suggestion
      <%= link_to image_tag(Current.organziation.logo, class: 'logo me-2', alt: 'site logo'),
```
An idea of how to pass around the organization to make tweaks on the UI.

edwinthinks (commented):
Thanks for starting this convo starter! I added some more commentary on the approach in Slack.

app/controllers/application_controller.rb
edwinthinks: Technically I think we could use rich text columns for the various pages on the Organization to get that wysiwyg experience. It won't be super super customizable but enough to maybe be different enough?

app/views/layouts/_header.html.erb
nsiwnf: or `current_tenant.logo` ?

Issue #83, opened 2023-07-29
Vaccine Reminder Mailer V2: Based on Checklist Assignments


nsiwnf:
### What
Depends on completion of issues #76 and #82  

Send a reminder based on all checklist_assignment items.

Ex. 
```
Congrats on [pet]'s adoption!

Hope everything's going well! This is just a reminder on tasks related to [pet].

- Make sure vaccinations are up to date.
- assignment 2.
- assignment 3.

Have a great day!
```

### Why
(Further) reduce the burden of post-intake/post-adoption follow-up by sending out an automated reminder about pet vaccines adhering to the Organization's processes

PR #144, opened 2023-08-03
Checklist Assignment Mailer

garyhtou:

This PR hooks up the Match Mailer to actually show its list of Checklist Assignments (which can include vaccinations and other tasks).

<table>
<tr>
 <td>HTML
 <td>Text
<tr>
 <td>
 <td>
</table>

The data in the image was seed data generated by Faker, so it doesn't make much sense. But we can probably imagine what it'd look like with proper titles and descriptions.

Here's what it looks like when there's only one task:
<table>
<tr>
 <td>HTML
 <td>Text
<tr>
 <td>
 <td>
</table>

### Link to the issue
https://github.com/rubyforgood/pet-rescue/issues/83

### More details

- Mailer is only sent if the Match contains at least one required, incomplete assignment
- Assignments in the mailer are ordered by due date
- Only incomplete assignments are listed


app/models/checklist_assignment.rb
```diff
@@ -26,7 +27,19 @@ class ChecklistAssignment < ApplicationRecord
   scope :completed, -> { where.not(completed_at: nil) }
   scope :incomplete, -> { where(completed_at: nil) }
 
+  scope :required, -> { joins(:checklist_template_item).where(checklist_template_items: {required: true}) }
+  scope :optional, -> { joins(:checklist_template_item).where(checklist_template_items: {required: false}) }
+
+  before_create :set_due_date
```
edwinthinks, line 33: I would move this `before_validation` with `on: :create` so that you can add a validation on-top of the `due_date` (I noticed we do not allow null on this column)

app/models/checklist_assignment.rb
```diff
@@ -26,7 +27,19 @@ class ChecklistAssignment < ApplicationRecord
...
+  delegate :name, :description, :expected_duration_days, :required, :required?, :optional?, to: :checklist_template_item
+
   def completed?
     completed_at
   end
+
+  private
+
+  def set_due_date
+    self.due_date = Time.now + expected_duration_days.days
```
edwinthinks, line 43:
```suggestion
    self.due_date ||= Time.now + expected_duration_days.days
```
In case, someone manually adds them (in the future maybe) then we could skip this OR we could add a `if:` to only run this if the due_date is missing.

edwinthinks (changes requested):

PR #145, opened 2023-08-03
Validate `ChecklistTemplateItem#expected_duration_days` is >= 0

garyhtou:

Validates that the `expected_duration_days` fields on `ChecklistTemplateItem` is greater than or equal to zero.

### Link to the issue

Addresses this comment: https://github.com/rubyforgood/pet-rescue/pull/111#discussion_r1278410872


app/models/checklist_template_item.rb
```diff
@@ -21,6 +21,8 @@
 #
 class ChecklistTemplateItem < ApplicationRecord
   belongs_to :checklist_template
-  validates :name, :expected_duration_days, presence: true
   has_many :checklist_assignments, dependent: :destroy
+
+  validates :name, presence: true
+  validates :expected_duration_days, presence: true, numericality: {greater_than_or_equal_to: 0}
```
edwinthinks, line 27: Could you add a test please, there is a nice shoulda matcher we can use to make this happen fast :)

edwinthinks (changes requested):

app/models/checklist_template_item.rb
garyhtou: sorry it took me so long to get around to this!

nsiwnf (approved):

PR #148, opened 2023-08-03
Configure email addresses

### LINKED ISSUE #119: Remove hardcoded email addresses
And figure out a better solution.

Doing a global search for `gmail`, for example, is a good way to start.

nsiwnf:

Replacing the hard-coded emails with configurable ones
Will follow-up with other email content in #150 

### Link to the issue

Resolves #119

### DIFF
--- a/app/services/MultiTenantService.rb
+++ b/app/services/MultiTenantService.rb
@@ -9,4 +9,7 @@ class MultiTenantService
     Rails.configuration.email_from[:default][service_org]
   end
 
+  def contact_email
+    Rails.configuration.email_from[:contact][service_org]
+  end
 end
--- a/app/services/MultiTenantService.rb
+++ b/app/services/MultiTenantService.rb
@@ -1,0 +1,12 @@
+class MultiTenantService
+  attr_accessor :service_org
+
+  def initialize(service_org)
+    @service_org = service_org.to_sym
+  end
+
+  def default_email
+    Rails.configuration.email_from[:default][service_org]
+  end
+
+end
--- a/app/services/multi_tenant_service.rb
+++ b/app/services/multi_tenant_service.rb
@@ -1,5 +1,4 @@
 class MultiTenantService
-
   def initialize(service_org)
     @service_org = service_org.to_sym
   end
@@ -11,7 +10,7 @@ class MultiTenantService
   def contact_email
     Rails.configuration.email_from[:contact][service_org]
   end
-  
+
   private
 
   attr_accessor :service_org
--- a/app/services/multi_tenant_service.rb
+++ b/app/services/multi_tenant_service.rb
@@ -1,17 +1,17 @@
 class MultiTenantService
-  def initialize(service_org)
-    @service_org = service_org.to_sym
+  def initialize(service_org_name)
+    @service_org_name = service_org_name.to_sym
   end
 
   def default_email
-    Rails.configuration.email_from[:default][service_org]
+    Rails.configuration.email_from[:default][service_org_name]
   end
 
   def contact_email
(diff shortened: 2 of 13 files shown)


app/services/MultiTenantService.rb
```diff
@@ -0,0 +1,15 @@
+class MultiTenantService
```
edwinthinks, line 1: Can we change the name of the file to be snakecase to match Rail's convention. `multi_tenant_service.rb` resolves to this same class name

edwinthinks: Since this is related to organizations, or tenants maybe you can namespace this to start building a hierarchal structure like:
```
Organizations::EmailService
Tenants::EmailService
```

On another note, I think naming this to be specific to emails might make sense. Unless the intention is to extend this to address things beyond emails eventually?

app/services/MultiTenantService.rb
```diff
@@ -0,0 +1,15 @@
...
+  attr_accessor :service_org
```
edwinthinks, line 2: If this isn't going to be accessible outside of the class you can put this beneath private so the surface area of the class stays tiny. I found it good to keep the surface area small so things stay manageable.

```
private

attr_accessor :service_org
```

config/environments/development.rb
```diff
@@ -83,4 +84,11 @@
 
   # Devise config
   config.action_mailer.default_url_options = {host: "localhost", port: 3000}
+
+  alta_from_domain = "alta.localhost"
+  rubyforgood_from_domain = "rubyforgood.localhost"
+  config.email_from = {
+    default: {alta: "hello@#{alta_from_domain}", rubyforgood: "hello@#{rubyforgood_from_domain}"},
+    contact: {alta: "contact@#{alta_from_domain}", rubyforgood: "contact@#{rubyforgood_from_domain}"}
```
edwinthinks, line 92: We can probably store this on some sort of object or in the `Organization` table to avoid having to commit changes in order to onboard new organizations. I believe the more we have to commit code, the higher chances of breakage.

We aren't yet there yet so I think its fine to continue this way and can save that work for some other day if we see its needed. What ya think @nsiwnf ?

edwinthinks (changes requested):
Nice work! I am really liking our direction to allow for this multi tenant experience!

app/services/MultiTenantService.rb
nsiwnf: Yeah, I was thinking it might be nice to centralize it so I kept it generic for now

config/environments/development.rb
nsiwnf: Yeah thats a good idea. Maybe we can leave them in here until we talk a bit about how Organization is going to look?

app/services/multi_tenant_service.rb
```diff
@@ -0,0 +1,17 @@
+class MultiTenantService
+  def initialize(service_org)
```
edwinthinks, line 2:
```suggestion
  def initialize(service_org_name:)
```
This includes two suggestions, adding `_name` because it isn't clear what `service_org` is suppose to be. However, I think in this case we are thinking its going a string.

The other suggestion is to use keyword arguments. I prefer it because we stop having to rely on positional arugments which is hard to remember/know what they are. However, in this case we have only 1 parameter which isn't going to be a problem. The suggestion is up to you if you want to use keyword argument.

edwinthinks (approved):
One last suggestion, once you've made decisions of any changes you want to make then you can merge this in.

edwinthinks (approved):
One last suggestion, once you've made decisions of any changes you want to make then you can merge this in.

PR #149, opened 2023-08-03
Finish adding factory usage and remove a few fixtures

Thrillberg:

This PR _finally_ replaces all fixtures with factories! 😄 🎉 A few notes:
- I've added a note to the README about adding `APP_HOST=localhost` to the test command when trying to run system tests. Shout out to @edwinthinks for help debugging that!
- I configured Geocoder so that it wouldn't make real HTTP calls anymore 🙀 
- I've generally been very verbose on test setup, preferring not to use `before` blocks or extracting things to methods in `test_helper`. I know this is the opposite of DRY but I feel like tests should be wet and more manageable "in one screen" if that makes sense. Happy to compromise on maybe a damp solution if anyone feels like we should DRY this up more, though!
- Sorry about the size of this PR. Maybe it's too big, but I just wanted to get past the finish line on this fixture-extraction.

### Link to the issue
#126


test/test_helper.rb
```diff
@@ -45,4 +35,19 @@ def check_messages
       with.library :rails
     end
   end
+
+  # Don't make real HTTP calls with Geocoder
```
nsiwnf, line 39: 👏

nsiwnf (approved):
lgtm!

PR #152, opened 2023-08-04
Convert StatesControllerTest to use factories

Thrillberg:

This is a hotfix for my last PR which broke `main`. Turns out I needed to convert ooonnnne more test!

### Link to the issue


egemen-dev (approved):
lgtm! 👍

PR #154, opened 2023-08-05
Some config changes that will hopefully fix Heroku deployment

Thrillberg:
Set staging environment to use local disk storage. Azure is not set up with Heroku. Other changes don't make a meaningful difference but will hopefully clarify things.


nsiwnf (approved):
lgtm!

PR #157, opened 2023-08-07
Fix seed file for pet weight refactor

kasugaijin:
Quick fix to seed file to fix the reference to Pet Weight Units. Otherwise `bin/setup` throws an error `NoMethodError: undefined method weight_units for Pet:Class` as the method no longer exists.
### Link to the issue


nsiwnf (approved):
👍

PR #158, opened 2023-08-07
WIP internationalize model validation errors

kasugaijin:
create translation files for model validation errors
PR is a WIP, no need to review yet.
### Link to the issue

### DIFF
--- a/config/locales/pets.en.yml
+++ b/config/locales/pets.en.yml
@@ -1,0 +1,33 @@
+en:
+  errors:
+    messages:
+      # messages for active storage validations gem
+      content_type_invalid: "must be PNG or JPEG YOOOO"
+      file_size_out_of_range: "size must be between 10KB and 1MB"
+      limit_out_of_range: "5 maximum"
+  activerecord:
+    errors:
+      models:
+        pet:
+          attributes:
+            name:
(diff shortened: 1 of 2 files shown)


config/locales/pets.en.yml
```diff
@@ -0,0 +1,33 @@
+en:
+  errors:
+    messages:
```
kasugaijin, line 3: Internationalization for the active record validations gem is scoped like this. Now, we only use this gem right now on the Pet CRUD, so they are included in this file. However, I am inclined to move this to the main `en.yml` file as we are likely to want to enable uploads on other models in futures.

config/locales/pets.en.yml
```diff
@@ -0,0 +1,33 @@
...
+      content_type_invalid: "must be PNG or JPEG YOOOO"
+      file_size_out_of_range: "size must be between 10KB and 1MB"
+      limit_out_of_range: "5 maximum"
+  activerecord:
+    errors:
+      models:
+        pet:
+          attributes:
+            name:
+              blank: "can't be blank"
```
kasugaijin, line 14: if preferred I can refer all of these repetitive messages to a single messages key that is this value

Issue #33, opened 2023-07-28
Ad asterisks to the required fields


edwinthinks:
![Image](https://user-images.githubusercontent.com/11335191/256664144-83d3b3c4-cc62-4db1-9287-264a0da3b109.png)

cheemurakami:
Hi! I'd like to pick this up 😸

PR #163, opened 2023-09-02
Add asterisks to the required fields

cheemurakami:
Updated text field labels.

Screenshot 2023-09-02 at 3 03 59 PM

### Link to the issue

Fixes https://github.com/rubyforgood/pet-rescue/issues/33


app/views/organization_pets/_form.html.erb
```diff
@@ -6,7 +6,7 @@
                             value: current_user.staff_account.organization_id %>
 
       <div class='form-group'>
-        <%= form.label :name, class: 'mt-3' %>
+        <%= form.label :name, 'Name*', class: 'mt-3' %>
```
edwinthinks, line 9: Could you check to see if adding `required: true` adds the asterisk for us? I normally see the asterisk added using that approach.

edwinthinks (changes requested):
You should be able to use the `required:` field to add the asterisks. Let me know if this doesn't work.

app/views/organization_pets/_form.html.erb
cheemurakami: Hi @edwinthinks ! I tried to use `required: true`, but it wasn't applying the asterisk. I followed [this file](https://github.com/rubyforgood/pet-rescue/blob/main/app/views/contacts/new.html.erb#L32-L35) 🤔

edwinthinks: Hmm I suspect it is because this is using form_with and not `form_for` could you see if that makes a difference?

edwinthinks (approved):
LGTM! Sorry for the delay @cheemurakami, we are reworking some major pieces of the codebase.

Issue #174, opened 2023-09-09
Engineer can create Admin user for an organization


kasugaijin:
As an engineer I can create a new user with Admin role for an organization that has registered on the app.

Use rolify gem to establish admin role.

[ ] As a staff user with admin role I receive my username and password and can log in and I see admin on the top corner

kasugaijin:
closed: https://github.com/rubyforgood/pet-rescue/pull/178

PR #178, opened 2023-09-17
add Rolify gem and enable roles to be added to StaffAccount model for…

kasugaijin:
This PR uses the ROlify gem to allow an engineer to add a role to a StaffAccount instance associated with instances of Organization and User.

We no longer create staff accounts in the UI, so the flow is done in the console. When a new organization wants to sign up, the Engineer:
- creates Organization
- creates User
- creates StaffAccount
- adds role 'admin' to the StaffAccount, scoped to the Organization. 

### Link to the issue 
https://github.com/rubyforgood/pet-rescue/issues/174


app/models/role.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  belongs_to :resource,
+    polymorphic: true,
+    optional: true
+
+  validates :resource_type,
+    inclusion: {in: Rolify.resource_types},
+    allow_nil: true
+
+  scopify
```
edwinthinks, line 28: What is this additional `scopify` for?

test/models/role_test.rb
```diff
@@ -0,0 +1,28 @@
+require "test_helper"
+
+class RoleTest < ActiveSupport::TestCase
+  def setup
+    @organization = create(:organization)
+    @staff_account = create(:staff_account, organization: @organization)
+  end
+
```
edwinthinks, line 8: Since we've added https://github.com/thoughtbot/shoulda-matchers you can add some quick tests to make sure the validation & the belongs_to works.

test/models/role_test.rb
```diff
@@ -0,0 +1,28 @@
...
+    refute @staff_account.has_role?(:admin, @organization)
+  end
+
+  test "removes the admin role from staff_account" do
+    @staff_account.add_role(:admin, @organization)
+    @staff_account.remove_role(:admin, @organization)
+
+    refute @staff_account.has_role?(:admin, @organization)
+  end
```
edwinthinks, line 27: I don't think you need these tests because you are testing the gem "rollify" which we assume is working well. Although it adds some extra layer of protection against bugs, it is often redundant and should trust our gems to work properly.

edwinthinks (commented):
A few minor changes & thoughts and then we can get this merged in!

app/models/role.rb
kasugaijin: the entire class was created by the gem after running the rolify generator - I did not add scopify

test/models/role_test.rb
kasugaijin: Yeah fair enough. It was as much for my own apeasement that I had set it up correctly

test/models/role_test.rb
kasugaijin: For my benefit, how is this not testing Rails code? I.e. per the model unit tests I hae below, we don't want them as they test the gem functionality, doesn't this test Rails associations functionality? I am happy to add them, but then I'd argue a case for keeping the tests that ensure roles are added and removed.

app/models/role.rb
edwinthinks: Gotcha, i'll assume it has a purpose and we can keep it as such.

test/models/role_test.rb
edwinthinks: Yeh you are right they are close in similarity. I think these shoulda validations & associations are very low cost that it just checks that you included those lines of code in your model. For the low price of having it, I generally just include them in case a person wants to remove a validation unknowningly. I would think differently if they cost more lines of code to have them.

test/models/role_test.rb
edwinthinks: I think if we didn't have other test coverages that the chances of releasing code without the really critical rolify wouldn't be to common. If this was removed, then I suspect every (if not all) controller and system tests would break because we rely on it.

Accidental removals happen and I think that a sign of good test coverage is everything breaking if something like this happens. Let me know if that makes sense.

test/models/role_test.rb
kasugaijin: yes that is fine by me, I figured the same and added them.

edwinthinks (approved):
LGTM!

test/models/role_test.rb
kasugaijin: yes that does make sense! I have removed them.

kasugaijin:
> LGTM!

Just merged main and added the admin label to the UI using the new template. Just giving you a heads up in case you wanted to see. Otherwise I will merge in a bit.

Issue #177, opened 2023-09-15
Utilize path name slug to handle multi-tenancy over using subdomains to avoid having to manage the subdomain routes.


edwinthinks:
I suspect this will make things easier in the beginning and we can always move forward to using subdomains when the need arises.

edwinthinks:
Closed with #179

PR #179, opened 2023-09-17
Utilize the path instead of subdomain to manage multi-tenancy

edwinthinks:

Uses the `/path` instead of the subdomain to manage the multi-tenancy look up. This should be more sustainable in the long run using this approach.

https://jam.dev/c/492ce5db-b7f6-475e-a31e-b11ba92ea979

Notably a number of tests break using this. Maybe we can revisit which tests make sense as we are rebuilding a bunch of things.

### Link to the issue
https://github.com/rubyforgood/pet-rescue/issues/177


kasugaijin:
I will leave this to someone with more experience in tenant architecture than me to approve, but this seems to be a reasonable and slightly simpler approach.

edwinthinks:
> I will leave this to someone with more experience in tenant architecture than me to approve, but this seems to be a reasonable and slightly simpler approach.

Sure. In addition, I can work through and talk through this together in our Tuesday meeting. Most of the multi-tenancy comes from the acts_as_tenant gem, this really handles how we define it. Happy to talk through it.

PR #187, opened 2023-09-24
Admin dashboard route and page

kasugaijin:
This PR: 
- adds route to dashboard
- adds dashboard controller
- adds index view file with start of the navbar for dashboard UI

Note that
- once we are happy with the basic navbar options, will need to update classes to get functionality working properly
- some discussion (at least for my benefit) on how we want to organize view templates for this portion of the site as I am thinking we want a SPA feel here.

image


app/views/organizations/dashboard/index.html.erb
```diff
@@ -0,0 +1,104 @@
+ <!-- Sidebar -->
+ <nav class="navbar-vertical navbar">
+     <div class="vh-100" data-simplebar>
+         <!-- Brand logo -->
```
edwinthinks, line 4: I think this makes sense to be put in a partial on it's own and a layout for dashboard like pages.

app/views/organizations/dashboard/index.html.erb
```diff
@@ -0,0 +1,104 @@
...
+                    </a>
+                    <div id="navCourses" class="collapse @@if (context.page_group === 'courses') { show }"
+                        data-bs-parent="#sideNavbar">
+                        <ul class="nav flex-column">
+                            <li class="nav-item">
+                                <a class="nav-link @@if (context.page === 'allcourses') { active }"
+                                    href="@@webRoot/pages/dashboard/admin-course-overview.html">
+                                    Settings
+                                </a>
+                            </li>
```
edwinthinks, line 25: When we get to work on this page let's make sure to use https://github.com/comfy/active_link_to which makes things a bit easier to show active states on links.

app/views/organizations/dashboard/index.html.erb
kasugaijin: Ah yes now I know what we're doing.

app/views/organizations/dashboard/index.html.erb
kasugaijin: Yes the hrefs in there right now are dropped in from the template. All need replacing.

kasugaijin:
I updated to use a dashboard layout template that renders the navbar and whatever template is required depending on the selection from the navbar usind yield.

I think all we have to do is set up other templates, named and routed appropriately, and add below and it will work with the `render` 

```<% content_for :dashboard_content do %>
  <!-- Content for Dashboard link goes here -->
<% end %>```

kasugaijin:
FYI: I made this PR yesterday as a little launch pad for Jade to continue on with. As such, there's a few things that are incomplete with the intent they will be continue. I guess Jade could contribute to this branch if we don't want to merge it in. I am open.

PR #193, opened 2023-09-26
Update redirection logic after login to goto dashboard if they are verified staff

edwinthinks:

Adds logic to redirect verified staff to the dashboard so they are brought to management tools directly after login. Any non-verified staff are taking to the adoptable pets section.

### Demo

N/A

### How to Test
1. Goto the /alta/users/sign_in page
2. Login as staff@alta.com / 123456
3. You should be redirected to the dashboard page.

### Link to the issue

N/A


kasugaijin (approved):
LGTM

PR #197, opened 2023-09-26
Update Navbar Links

kasugaijin:
… because the text was hidden and this made it appear. Some more tinkering required to fix the position

I added new links to the navbar. 
In this process I had to create some new controllers and views. I decided to scope all relevant controller and views under Organization (including existing ones that were not, as it had to be done eventually). 

To do: 

- update broken tests to fix the url paths than have changed after scoping stuff under Organizations.

Note: the UI is a bit broken. Edwin is looking into fixing the scrollbar issues surrounding the links. 

image

### Link to the issue


edwinthinks:
Hey @kasugaijin I just updated with main with the fixed test. I'll review this now :)

app/controllers/organizations/foster_application_reviews_controller.rb
```diff
@@ -0,0 +1,4 @@
+class Organizations::FosterApplicationReviewsController < Organizations::BaseController
+  def index
+  end
+end
```
edwinthinks, line 4: Could we add the `layout "dashboard"` in these controllers?

edwinthinks (approved):
One minor comment. Once the tests are passing or you decide to skip them, feel free to merge in!

This is exciting!

app/views/organizations/foster_application_reviews/edit.html.erb
```diff
@@ -0,0 +1,61 @@
...
+  </div> <!--container-->
+</header>
+
+<!--main section-->
+<section class="pb-5" id="about_us">
+  <div class="container">
+    <div class="row">
+      <div class="col-md-6 mx-auto card p-5 ">
+        <div class='d-flex flex-row-reverse'>
+          <%= link_to "#{@application.adopter_account.user.first_name}'s Profile",
```
edwinthinks, line 20:
```suggestion
          <%= active_link_to "#{@application.adopter_account.user.first_name}'s Profile", 
```
Let's use `active_link_to` in place of `link_to` so it knows to add the active class when we are on that page the link directs towards.

kasugaijin: Ah yes you mentioned that before. I will change that.

app/controllers/organizations/foster_application_reviews_controller.rb
kasugaijin: certainly can

Issue #195, opened 2023-09-26
Update options in the avatar dropdown menu to be only Sign Out


edwinthinks:
Let's remove any dropdown options we don't yet support and make the Log Out button work
- Remove all options except for "Log Out"
- Update "Log Out" to actually log out the user

Watch the screen recording here: https://jam.dev/c/7d6e35af-a20c-4534-ba4b-0bf12e5e76bf

\-------------------------------------

**Website URL:**
[http://localhost:3000/alta/dashboard](http://localhost:3000/alta/dashboard)

**Device and browser info:**
Chrome 117.0.0.0 (1335x837) | Mac OS (arm) 10.15.7

**Date and time:**
September 26th 2023 | 11:57am UTC

**View developer information (console logs, network requests & timing):**
[jam.dev/c/7d6e35af-a20c-4534-ba4b-0bf12e5e76bf](https://jam.dev/c/7d6e35af-a20c-4534-ba4b-0bf12e5e76bf)

\-------------------------------------

*Captured with* *[jam.dev](https://jam.dev?utm_source=github&utm_medium=integration&utm_campaign=captured_with_link_2022_11)*

PR #198, opened 2023-09-26
Update: Removed unused buttons. Log out function fails upon button click

ErinClaudio:

- Removed extraneous dropdown options, retaining only the "Log Out" button.
- Updated the "Log Out" button to successfully log out the user.
- Adjusted system tests to account for dynamic organization slugs and improved test reliability.

### Link to the issue
[#195](https://github.com/rubyforgood/pet-rescue/issues/195)

### Concerns
I have reservations about the sad path test for the "non-authenticated user attempts to log out" scenario. While the test passes, I believe it can be enhanced. Would appreciate guidance or a review on this particular test.


kasugaijin:
Could you check to see if your change is causing the one test in the pipeline to fail?
Thanks!

edwinthinks:
Looks like the failed test has nothing to do with the changes. 
```
LoginTest#test_: when logging in as a staff member should direct to the organization's adoptable pets page. :
  Ferrum::ProcessTimeoutError: Browser did not produce websocket url within 5 seconds, try to increase `:process_timeout`. See https://github.com/rubycdp/ferrum#customization
```

This is usually caused because the CI machine is too slow. We can fix this by making a tweak in our settings. @ErinClaudio @kasugaijin we can safetly just ignore that. I'll open a issue to resolve it.

kasugaijin:
> Looks like the failed test has nothing to do with the changes.
> 
> ```
> LoginTest#test_: when logging in as a staff member should direct to the organization's adoptable pets page. :
>   Ferrum::ProcessTimeoutError: Browser did not produce websocket url within 5 seconds, try to increase `:process_timeout`. See https://github.com/rubycdp/ferrum#customization
> ```
> 
> This is usually caused because the CI machine is too slow. We can fix this by making a tweak in our settings. @ErinClaudio @kasugaijin we can safetly just ignore that. I'll open a issue to resolve it.

Ah yes. This is out flaky test. It failed a couple times ealier this evening too.

ErinClaudio:
@kasugaijin @edwinthinks Thank you, for the update!

app/views/layouts/dashboard.html.erb
```diff
@@ -197,56 +197,9 @@
...
-                    </ul>
-                    <div class="dropdown-divider"></div>
-                    <ul class="list-unstyled">
-                      <li>
-                        <a class="dropdown-item" href="../../index.html">
-                          <i class="fe fe-power me-2"></i> Sign Out
-                        </a>
-                      </li>
+                        <li>
+                          <%= button_to "<i class='fe fe-power me-2'></i> Sign Out".html_safe, destroy_user_session_path, method: :delete, class: 'dropdown-item', form: { data: { turbo: false } } %>
```
edwinthinks, line 201: To avoid having to use `.html_safe` here you could use the button_to and a pass it a block. 
```
<%= button_to url, ... do %>
   <i class='fe fe-power me-2'></i> Sign Out
<% end %>
```
You'll need to double check the syntax.

edwinthinks (changes requested):
Thanks @ErinClaudio for your PR. I left you a comment let me know if you have any Qs

app/views/layouts/dashboard.html.erb
ErinClaudio: Sorry, I did not mean to assign this for review. 
But thank I will make that change

ErinClaudio: This has been changed

edwinthinks (approved):
Thanks @ErinClaudio ! LGTM!

PR #201, opened 2023-09-27
Address issues with sidebar rendering in Dashboard

edwinthinks:

This PR fixes the issues we had with the sidebar installation. It turns out that when we copied the contents over from the inspector we were also copying over classes and changes that the simplebar library handles which caused things to not work. The way I fixed it was I went into the `src` folder in the downloaded geeks library and saw that the sidebar actually was much more bare then we had it.

### Link to the issue

N/A


kasugaijin (approved):
I'll give it a look!

Issue #190, opened 2023-09-26
Update the pets index page to utilize the new dashboard layout


edwinthinks:
Let's start utilizing our dashboard layout that we've been implementing by moving over the http://localhost:3000/alta/pets to match the new styling.

You can get to this page by going to Rescues > Alta > Login (use staff@alta.com) > Upper right corner > Our Pets

Acceptance Criteria
- [x] The layout uses the new geeksui layout we've labeled "dashboard"
- [x] The styling of the internal page stays the same. **Do not worry about updating the Pet cards yet**
- [x] Move the "Our Pets" to the header section
- [x] Move the "Create Pet" button to the upper right corner as a button

PR #204, opened 2023-09-27
Pets index styling

jadekstewart3:
Update "/org/pets" to utilize the geeksui styling.
Move the "Our Pets" to the header section
Move the "Create Pet " link to the upper right and hand corner and make it a button instead of a link

### Link to the issue
#190


app/views/organizations/pets/index.html.erb
```diff
@@ -14,12 +14,21 @@
...
-        <%= t('.our_pets') %>
-      </h1>
-    </div>
+    <%= content_for :header do  %>
+      <div class="text-center mb-5">
+        <h2 class="section-heading text-uppercase">
+          <%= t('.our_pets') %>
+        </h2>
+      </div>
+    <% end %>
```
edwinthinks, line 23: I think we should reduce the number of details in the header to be a bit smaller. This is because you'll have to repeat the div & h2 classes for every page. Instead could we do something like:
```
<%= provide(:header_title, t(".our_pets")) %>
```
and in the layout have something like:
```
      <div class="text-center mb-5">
        <h2 class="section-heading text-uppercase">
          <%= yield(:header_title) %>
        </h2>
      </div>
```

This is going to make the amount of information each page needs to write less. Let me know what you think :)

app/views/organizations/pets/index.html.erb
```diff
@@ -14,12 +14,21 @@
...
 
+    <!--create pet button-->
+      <%= content_for :button do  %>
+        <div class ="col-md-6 d-flex justify-content-center align-items-center">
+          <%= button_to t('.create_pet'), new_pet_path, class: "btn btn-primary" %><br>
+        </div>
```
edwinthinks, line 29:
```suggestion
          <%= button_to t('.create_pet'), new_pet_path, class: "btn btn-primary" %><br>
```
I think you can change it to this since the `col-md-6` is shrinking the button and I think it will look fine without the surrounding `div`

edwinthinks (changes requested):
Nice work @jadekstewart3 ! A few things to move around but it looks good! Let me know if you want to clarify on any of my comments.

app/views/organizations/pets/index.html.erb
```diff
@@ -14,12 +14,21 @@
   <div class="container mb-md-5 mt-md-5">
 
     <!--page heading-->
-    <div class="text-center mb-5">
-      <h1 class="section-heading text-uppercase underline">
-        <%= t('.our_pets') %>
-      </h1>
-    </div>
+    <%= content_for :header do  %>
```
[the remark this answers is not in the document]
jadekstewart3: @edwinthinks Thank you! I'm going to get on this right now!

edwinthinks (approved):
After you've fixed the merge conflict..... LGTM!

Thanks for tackling this!

edwinthinks:
LGTM!

PR #205, opened 2023-09-27
Upgrade flash messages

### LINKED ISSUE #186: As a user, I want to clearly see the flash messages I receive so I know if there is something I need to know
Let's upgrade our flash messages to utilize the styling of https://geeksui.codescandy.com/geeks/docs/toasts.html and combine it with https://www.stimulus-components.com/docs/stimulus-notification (so it fades out on its own).

## Requirements
- Preferably have the toast/flash message in the middle top 
- Make sure the flash message disappears after a few seconds to avoid blocking the screen too much. (Use the stimulus components)
- Include it in every layout except for `devise`

marlena-b:
Upgrade flash messages on every layout except the devise. It uses the bootstrap toast element and is triggered by the stimulus controller.

I didn't use the stimulus notifications because bootstrap toasts disappear automatically without any additional JS.

https://github.com/rubyforgood/pet-rescue/assets/96994176/356da918-2da2-45de-abc5-350d66c42cd1

### Link to the issue

Fixes #186

### DIFF
--- a/app/javascript/controllers/flash_message_controller.js
+++ b/app/javascript/controllers/flash_message_controller.js
@@ -1,0 +1,8 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  connect() {
+    var toast = new bootstrap.Toast(this.element, { delay: 3500 })
+    toast.show()
+  }
+}
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -3,6 +3,7 @@
   <%= render "layouts/shared/head" %>
 
   <body>
+    <%= render "layouts/shared/flash_messages" %>
     <% if Current.tenant.present? %>
       <%= render partial: "layouts/shared/navbar" %>
     <% else %>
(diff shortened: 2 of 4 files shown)


app/javascript/controllers/flash_message_controller.js
```diff
@@ -0,0 +1,8 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  connect() {
+    var toast = new bootstrap.Toast(this.element, { delay: 3500 })
```
marlena-b, line 5: The default delay of 5s seemed too long to me so I adjusted it to 3.5s. Let me know if it's ok

marlena-b:
I wasn't sure if I should write tests for those changes. Let me know if I should add them

app/javascript/controllers/flash_message_controller.js
edwinthinks: Sounds reasonable thank you!

edwinthinks (approved):
LGTM thanks for adding this contribution @marlena-b !

Issue #168, opened 2023-09-06
As a Organization Admin, I can invite other staff members to be users in my organization so that they can assist me in running my organization


edwinthinks:
We seek to provide the functionality for Organization Admins to be able to invite other staff members to have special access to their organization.

PR #224, opened 2023-10-02
Form for adding staff

nsiwnf:
# 🔗 Issue
#168 

# ✍️ Description
Page for adding new staff members
- Random password is generated and has to be reset using "forgot password"
- Terms of service agreement not required

# 📷 Screenshots/Demos
Screen Shot 2023-10-02 at 1 59 36 PM

### DIFF
--- a/app/controllers/organizations/staff_controller.rb
+++ b/app/controllers/organizations/staff_controller.rb
@@ -1,9 +1,31 @@
 class Organizations::StaffController < Organizations::BaseController
-  layout "dashboard"
-
   before_action :require_organization_admin
+  layout "dashboard"
 
   def index
     @staff_accounts = StaffAccount.all
+  end
+
+  def new
+    @staff = StaffAccount.new(user: User.new)
+  end
+
+  def create
+    @user = User.new(staff_params[:user].merge(password: SecureRandom.hex(8)))
+    @staff = StaffAccount.new(user: @user, verified: true)
+
+    if @user.save && @staff.save
+      @staff.add_role(staff_params[:role])
+      redirect_to staff_index_path, notice: "Staff saved successfully."
+    else
+      render :new, status: :unprocessable_entity
+    end
+  end
+
+  private
+
+  def staff_params
+    params.require(:staff_account)
+      .permit(:role, user: [:first_name, :last_name, :email])
   end
 end
--- a/app/controllers/organizations/staff_controller.rb
+++ b/app/controllers/organizations/staff_controller.rb
@@ -7,15 +7,15 @@ class Organizations::StaffController < Organizations::BaseController
   end
 
   def new
-    @staff = StaffAccount.new(user: User.new)
+    @user = User.new
+    @staff = StaffAccount.new(user: @user)
   end
 
   def create
-    @user = User.new(staff_params[:user].merge(password: SecureRandom.hex(8)))
-    @staff = StaffAccount.new(user: @user, verified: true)
+    @user = User.new(user_params.merge(password: SecureRandom.hex(8)))
 
-    if @user.save && @staff.save
-      @staff.add_role(staff_params[:role])
+    if @user.save
+      # @user.staff_account.add_role(staff_role[:role])
       redirect_to staff_index_path, notice: "Staff saved successfully."
     else
       render :new, status: :unprocessable_entity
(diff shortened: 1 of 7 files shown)


app/views/organizations/staff/_form.html.erb
```diff
@@ -0,0 +1,61 @@
+<!-- form -->
+<%= form_with model: staff, url: '/staff', method: :post do |form| %>
+  <%= form.hidden_field :organization_id, value: current_user.staff_account.organization_id %>
```
edwinthinks, line 3:
```suggestion
```
You shouldn't need this because `acts_as_tenant` automatically puts the current tenant on `.new`.
```
ActsAsTenant.with_tenant(Organization.find(id: 5)) do 
     StaffAccount.new # will automatically put organization_id=5, assuming acts_as_tenant is included in the model
end
```

app/controllers/organizations/staff_controller.rb
```diff
@@ -1,9 +1,31 @@
...
     @staff_accounts = StaffAccount.all
   end
+
+  def new
+    @staff = StaffAccount.new(user: User.new)
+  end
+
+  def create
+    @user = User.new(staff_params[:user].merge(password: SecureRandom.hex(8)))
+    @staff = StaffAccount.new(user: @user, verified: true)
```
edwinthinks, line 15: I think we can simplify this using nested form approach with accept_nested_attributes - https://www.ombulabs.com/blog/learning/rails/nested-forms.html

app/views/organizations/staff/_form.html.erb
```diff
@@ -0,0 +1,70 @@
+<!-- form -->
+<%= form_with model: staff, url: "/#{current_tenant.slug}/staff", method: :post do |form| %>
```
edwinthinks, line 2:
```suggestion
<%= form_with model: staff, url: staff_path, method: :post do |form| %>
```
I think this works as well to avoid having to hardcode the current_tenant.slug

app/views/organizations/staff/_form.html.erb
```diff
@@ -0,0 +1,70 @@
...
+  <div class="col-12 mt-3">
+    <%= form.submit 'Save profile', class: 'btn btn-primary' %>
+  </div>
+  </div>
+
+<% end %>
+
+
+
+
```
edwinthinks, line 70:
```suggestion
```

app/views/organizations/staff/new.html.erb
```diff
@@ -0,0 +1,27 @@
+<section class="container-fluid p-4">
+
+  <!--page heading-->
+  <%= provide(:header_title, "New staff") %>
+
+  <!--create pet button-->
```
edwinthinks, line 6:
```suggestion
```

edwinthinks (changes requested):
A few minor adjustments. Once those are addressed, I think we can merge this in :)

app/views/organizations/staff/_form.html.erb
nsiwnf: Hm, this gives me the error "No route matches {:action=>"show", :controller=>"organizations/staff"}, missing required keys: [:id]"

edwinthinks: Ahh I see now, it is `staff_index_path`, Didn't realize they do put `_index_` in there.

app/controllers/organizations/staff_controller.rb
```diff
@@ -1,9 +1,32 @@
...
     @staff_accounts = StaffAccount.all
   end
+
+  def new
+    @user = User.new
+    @staff = StaffAccount.new(user: @user)
+  end
+
+  def create
+    @user = User.new(user_params.merge(password: SecureRandom.hex(8)))
```
nsiwnf, line 15: @edwinthinks Not sure how to do `accept_nested_attributes` with `roles`. Should the role input be multi-select?

app/controllers/organizations/staff_controller.rb
```diff
@@ -1,9 +1,32 @@
...
+
+    if @user.save
+      # @user.staff_account.add_role(staff_role[:role])
```
edwinthinks, line 18:
```suggestion
      @user.staff_account.add_role(user_params[:staff_account_attributes][:role])
```
I think this is good enough for now

PR #225, opened 2023-10-02
Invert association between Location and AdopterProfile

### LINKED ISSUE #216: Refactor Location associations to facilitate multiple models with associated Location
Refactor Location associations to facilitate multiple models with association Location. Currently Location belongs_to AdopterProfile. We need the inverse of this association to enable more than one model to refer to a Location. 

Acceptance Criteria
- AdopterProfile belongs_to Location
- Location has many AdopterProfiles

marlena-b:
# 🔗 Issue
Fixes #216 

# ✍️ Description
This PR inverts the association between `Location` and `AdopterProfile` models.
Now:
* `AdopterProfile` belongs to `Location`
*  `Location` has one `AdopterProfile`. 

@kasugaijin suggested in the issue that `Location` should use `has_many :adopter_profiles` relation but from my understanding of the codebase the `has_one` relation makes more sense. Please let me know if I'm wrong, I will change it if needed.

### DIFF
--- a/db/migrate/20231002150909_change_association_between_location_and_adopter_profile.rb
+++ b/db/migrate/20231002150909_change_association_between_location_and_adopter_profile.rb
@@ -1,0 +1,35 @@
+class ChangeAssociationBetweenLocationAndAdopterProfile < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
+
+  def up
+    safety_assured do
+      add_reference :adopter_profiles, :location, foreign_key: true, index: {algorithm: :concurrently}
+
+      execute <<-SQL
+        UPDATE adopter_profiles
+        SET location_id = locations.id
+        FROM locations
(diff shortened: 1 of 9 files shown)


db/migrate/20231002150909_change_association_between_location_and_adopter_profile.rb
```diff
@@ -0,0 +1,35 @@
+class ChangeAssociationBetweenLocationAndAdopterProfile < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
+
+  def up
+    safety_assured do
```
marlena-b, line 5: I wrapped everything inside `safety_assured` because strong_migrations were complaining about splitting this code into multiple deployments. I think this was completely unnecessary for a system that does not have a production environment yet.

kasugaijin (approved):
looks great thank you!

Issue #222, opened 2023-10-01
Add "species" to pet model


nsiwnf:
DB migration to add "species" field to the pet model. Species can be things like "cat", "dog"

jadekstewart3:
I can do this!

jadekstewart3:
@nsiwnf are we including exotic pets like: "rabbit" "rodent" "lizard" "snake"?

nsiwnf:
> @nsiwnf are we including exotic pets like: "rabbit" "rodent" "lizard" "snake"?

Hmm, let's just do "cat" and "dog" for now

PR #233, opened 2023-10-03
Add species to pets table and set up enum in the pet model

jadekstewart3:
# 🔗 Issue
#222 

# ✍️ Description
Added species column to pets table, and setup species enum within the pets model.

# 📷 Screenshots/Demos
Screenshot 2023-10-03 at 2 16 30 PM

Screenshot 2023-10-03 at 2 16 43 PM

### DIFF
--- a/app/models/pet.rb
+++ b/app/models/pet.rb
@@ -31,6 +31,7 @@ class Pet < ApplicationRecord
   has_many :adopter_applications, dependent: :destroy
   has_one :match, dependent: :destroy
   has_many_attached :images
+  enum species: ["dog", "cat"]
 
   validates :name, presence: true
   validates :birth_date, presence: true
--- a/db/migrate/20231003182930_add_species2_pets.rb
+++ b/db/migrate/20231003182930_add_species2_pets.rb
@@ -1,5 +1,5 @@
 class AddSpecies2Pets < ActiveRecord::Migration[7.0]
   def change
-    add_column :pets, :species, :integer
+    add_column :pets, :species, :integer, null: false
   end
 end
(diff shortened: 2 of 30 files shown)


db/migrate/20231003182930_add_species2_pets.rb
```diff
@@ -0,0 +1,5 @@
+class AddSpecies2Pets < ActiveRecord::Migration[7.0]
+  def change
+    add_column :pets, :species, :integer
```
edwinthinks, line 3:
```suggestion
    add_column :pets, :species, :integer, null: false
```
Adding this `null: false` prevents at the database level from allowing rows without the species defined. Helps keep data more accurate in case of issues. This comes especially helpful if validations get skipped

app/models/pet.rb
```diff
@@ -31,6 +31,7 @@ class Pet < ApplicationRecord
   has_many :adopter_applications, dependent: :destroy
   has_one :match, dependent: :destroy
   has_many_attached :images
+  enum species: ["dog", "cat"]
```
edwinthinks, line 34: Let's add quick test for this using shoulda - https://matchers.shoulda.io/docs/v3.1.1/Shoulda/Matchers/ActiveRecord.html#define_enum_for-instance_method

edwinthinks (changes requested):

db/migrate/20231003182930_add_species2_pets.rb
edwinthinks: @jadekstewart3 in case you missed this comment :)

Ptrboro: @jadekstewart3 I think the `schema.rb` file was not updated with `null: false`. Easiest way to regenerate it is to run `rails db:rollback` and `rails db:migrate` :)

jadekstewart3: @Ptrboro Oh heck! Thank you!

edwinthinks (approved):
LGTM!

PR #237, opened 2023-10-04
Increase process timeout in tests

### LINKED ISSUE #203: Increase the process timeout so that slowness does not cause browser/system tests to fail randomly.
We've been getting flaky test failures caused by process timeouts. Let's make the necessary configurations to increase the timeout.

Error:
```
LoginTest#test_: when logging in as a staff member should direct to the organization's adoptable pets page. :
  Ferrum::ProcessTimeoutError: Browser did not produce websocket url within 5 seconds, try to increase `:process_timeout`. See https://github.com/rubycdp/ferrum#customization
```

Ptrboro:
# 🔗 Issue
Fixes #203 

# ✍️ Description
* Increased `process_timeout` for the driver to 20s (maybe 10s would work just fine but 20s is safer and doesn't cost us anything)
* I also bumped `city-state` gem to a newer version because it was throwing deprecation warnings all the time

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -148,7 +148,7 @@ end
 gem "chartkick", "~> 5.0"
 
 # Adds location data for cities and states around the world
-gem "city-state", "~> 0.1.0"
+gem "city-state", "~> 1.1.0"
 
 # Adds a simple way to fetch with Javascript
 gem "requestjs-rails", "~> 0.0.10"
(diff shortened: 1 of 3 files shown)


Ptrboro:
I tested this by rerunning the tests on CI a couple of times. All of the runs were successful

edwinthinks (approved):
LGTM!

Issue #175, opened 2023-09-09
Org Admin receives email with username and password upon account creation


kasugaijin:
As an organization admin, I receive an email with my username and password that allows me to log into my new account as the admin after I have registered for the app. 

Blocked by: https://github.com/rubyforgood/pet-rescue/issues/174

PR #240, opened 2023-10-04
Create new org and admin

kasugaijin:
# 🔗 Issue

[175](https://github.com/rubyforgood/pet-rescue/issues/175)

# ✍️ Description

This PR adds a service object that we can use via the console to:
- create a new organization
- create a new user
- create a new staff account
- add the admin role
- send email to admin user

There is also a new mailer and template for this flow.

# 📷 Screenshots/Demos


app/mailers/sign_up_mailer.rb
```diff
@@ -25,4 +25,12 @@ def admin_notification_new_staff(tenant_org)
       to: "wcrwater@gmail.com",
       subject: "New Staff Account")
   end
+
+  def create_new_org_and_admin(tenant_org)
+    @organization = params[:organization]
+    @user = params[:user]
+    mail(from: MultiTenantService.new(tenant_org).default_email,
```
kasugaijin, line 32: Not sure what to use as the `from` email. I suspect we'd want this to be from pet rescue rather than the org itself.

app/services/create_org_and_admin_service.rb
```diff
@@ -0,0 +1,102 @@
...
+  end
+
+  def send_email
+    return puts "**One or more of the prior steps failed. Email not sent.**" unless @success_counter == 4
+
+    OrganizationMailer.with(
+      user: @user,
+      organization: @organization,
+      password: @password
+    )
```
Ptrboro, line 84: Looks like the password is not needed here. The mailer uses `@user.password` instead.

```suggestion
    OrganizationMailer.with(
      user: @user,
      organization: @organization
    )
```

Ptrboro (approved):
LGTM! Leaving one suggestion

app/services/create_org_and_admin_service.rb
```diff
@@ -0,0 +1,102 @@
+# class to create a new organization, user, staff account with role admin, and send email via the console
+class CreateOrgAndAdminService
```
edwinthinks, line 2: Hey @kasugaijin, I've got a few ideas that I want to first say are suggestions and ideas for how to improve this. I'am a big believer in coming back to improve rather than put PRs in a holding pattern. With that said, here are some considerations and ideas to counter-act them.

Consider the situation that maybe 1 of the 4 actions failed. You'll need to re-run that 1 command until it works if you realize it didn't work. This can be problematic because the data for the organization will be in a "invalid" state with missing associations or records. One way to prevent this is issue is wrapping the entire process inside of a transaction block (https://api.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html). The transaction block lets you insert and update in seperate lines but rollbacks if there is an issue. I would use this transaction and use a signal method that takes a hash of information to be able to run.

Let me know if that makes sense

app/services/create_org_and_admin_service.rb
```diff
@@ -0,0 +1,102 @@
+# class to create a new organization, user, staff account with role admin, and send email via the console
```
edwinthinks, line 1: We may find ourselves creating more services like this. I think renaming this to be Organizations::CreateService would start allowing us to namepace these services to keep things organized.

app/services/create_org_and_admin_service.rb
```diff
@@ -0,0 +1,102 @@
...
+  private
+
+  def instructions
+    puts "Use the following methods and arguments in order to create a new Organization and Admin user:"
+    puts "  create_organization(name, slug)"
+    puts "  create_user(email, first_name, last_name)"
+    puts "  create_staff_account"
+    puts "  add_admin_role_to_staff_account"
+    puts "  send_email"
+  end
```
edwinthinks, line 97: Related to my comment above about wrapping everything in a single transaction. We could then remove this likely.

edwinthinks (changes requested):
Thanks for working on this @kasugaijin let me know if you have any Qs!

app/services/organizations/create_service.rb
```diff
@@ -0,0 +1,71 @@
...
+  def signal(args)
+    ActiveRecord::Base.transaction do
+      create_organization(args[:organization_name], args[:organization_slug])
+      create_user(args[:user_email], args[:user_first_name], args[:user_last_name])
+      create_staff_account
+      add_admin_role_to_staff_account
+      send_email
+    end
+  rescue => e
+    puts "An error occurred: #{e.message}"
```
edwinthinks, line 22:
```suggestion
    raise "An error occurred: #{e.message}"
```
So it is really loud when the error occurs.

edwinthinks (approved):
I think after the small suggestion we can get this merged in.

P.S - I'am curious about using https://github.com/collectiveidea/interactor to create a common service interface. I don't want to hold off the PR any longer so I think we can update it later or if you want to explore the use of the the gem.

edwinthinks (approved):
LGTM!

PR #251, opened 2023-10-10
Allow admins to invite other staff

### LINKED ISSUE #226: Send new staff an invitation to the org
As an admin, when I add a new staff to the org, I want them to get an invitation to complete their account details.

- Create invitation email for staff to create/update their account

Resources
https://github.com/scambra/devise_invitable

marlena-b:
# 🔗 Issue
Fixes #226 

# ✍️ Description
This Pull Request adds the functionality of inviting new staff members to the organization by administrators (staff admins) using the devise_invitable gem. The invitation acceptance flow is currently in a basic form but improvements will be addressed in a subsequent PR.

# 📷 Screenshots/Demos

https://github.com/rubyforgood/pet-rescue/assets/96994176/bf5a4266-1dcb-49e6-aec7-b88359a0a88b


app/controllers/organizations/invitations_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def create
+    @user = User.new(user_params.merge(password: SecureRandom.hex(8)).except(:staff_account_attributes))
+    @user.staff_account = StaffAccount.new(verified: true)
+
+    if @user.save
+      @user.staff_account.add_role(user_params[:staff_account_attributes][:roles])
+      @user.invite!(current_user)
+      redirect_to staff_index_path, notice: "Staff saved successfully."
```
kasugaijin, line 17: Can we chance to something like "Invite sent!"?

app/views/organizations/invitations/_form.html.erb
```diff
@@ -0,0 +1,39 @@
...
+      <div class="mb-3 col-md-6 col-12">
+        <div class="input-group me-3">
+          <%= staff_subform.select :roles, options_for_select([['Staff', :staff], ['Admin', :admin]]), class: 'form-control', required: true %>
+        </div>
+      </div>
+    <% end %>
+
+    <!-- button -->
+    <div class="col-12 mt-3">
+      <%= form.submit 'Save profile', class: 'btn btn-primary' %>
```
kasugaijin, line 35: Can we change button to 'Send Invite' ?

kasugaijin (approved):
Looks great! I tested it out locally and I like it. 

I know you said you had improvements to follow in another PR, so the following is my thoughts, some which you may already have considered:

- I am guessing the `invite!` call on line 16 of `Organizations::InvitationsController` is triggering the mailer. I am wondering if we are able to send any params to that? It would be good to send the organization so we can access its name and slug and recreate the Org's Url in the email e.g., https://www.petrescue.com/baja 
- The redirect after the invited user enters and saves their password to accept the invite, is this coming from registrations_controller   def after_sign_in_path_for ? This is another place where it would be good to scope it to the organization's dashboard rather than the root path, if possible.
- I think we could use the invitation_accepted_at column in the staff table to indicate whether or not the staff has accepted the invite (not directly related to this work) e.g., in the status column

db/migrate/20231009132343_devise_invitable_add_to_users.rb
```diff
@@ -0,0 +1,26 @@
+class DeviseInvitableAddToUsers < ActiveRecord::Migration[7.0]
```
kasugaijin, line 1: My initial thought without considering the gem would be to separate Invitations from Users and have an Invitations table/model. This would be more maintainable in the long run. 

However, looking at the gem, I see that adding to Users table is a 'normal' practice. I was thinking it doesn't make sense if only Org users use invites, as then we'd have a User class that for adopters has completely irrelevant fields. But, then I thought we could also leverage invitations between adopters as well at some point. 

This is not to say this needs to change. I think it works like this and is ok a. But, I think with more time and a bigger budget I'd go for separating those concerns :D

app/controllers/organizations/invitations_controller.rb
nsiwnf: I think both make sense here. The staff account does get saved and the invitation is sent

config/locales/devise_invitable.en.yml
```diff
@@ -0,0 +1,31 @@
...
+        header: "Set your password"
+        submit_button: "Set my password"
+    mailer:
+      invitation_instructions:
+        subject: "Invitation instructions"
+        hello: "Hello %{email}"
+        someone_invited_you: "Someone has invited you to %{url}, you can accept it through the link below."
+        accept: "Accept invitation"
+        accept_until: "This invitation will be due in %{due_date}."
+        ignore: "If you don't want to accept the invitation, please ignore this email. Your account won't be created until you access the link above and set your password."
```
nsiwnf, line 25: It seems like the account is still created? Or does devise_invitation do something to the User record that gets created?

nsiwnf:
These are great suggestions! Should we make them into tickets?

> Looks great! I tested it out locally and I like it.
> 
> I know you said you had improvements to follow in another PR, so the following is my thoughts, some which you may already have considered:
> 
>     * I am guessing the `invite!` call on line 16 of `Organizations::InvitationsController` is triggering the mailer. I am wondering if we are able to send any params to that? It would be good to send the organization so we can access its name and slug and recreate the Org's Url in the email e.g., https://www.petrescue.com/baja
> 
>     * The redirect after the invited user enters and saves their password to accept the invite, is this coming from registrations_controller   def after_sign_in_path_for ? This is another place where it would be good to scope it to the organization's dashboard rather than the root path, if possible.
> 
>     * I think we could use the invitation_accepted_at column in the staff table to indicate whether or not the staff has accepted the invite in the staff table (not directly related to this work) e.g., in the status column

nsiwnf (approved):
LGTM!

app/controllers/organizations/invitations_controller.rb
marlena-b: I think "Invite sent!" makes more sense from the user perspective. Fixed!

I also changed texts in a couple of other places from "Add staff" to "Invite staff".

app/views/organizations/invitations/_form.html.erb
marlena-b: Fixed, thanks!

db/migrate/20231009132343_devise_invitable_add_to_users.rb
marlena-b: I totally agree with your point and would love to make those changes too, but sadly, this gem doesn't let us do that (there is even an issue for this: https://github.com/scambra/devise_invitable/issues/888). 

Let's go with what we have for now and think about extracting it later on if needed.

config/locales/devise_invitable.en.yml
marlena-b: Hey @nsiwnf, this text was generated by Devise, and TBH I don't know why it says so. The user record will be created in the DB but with a random password so no one can access it. I will change the `Your account won't be created...` sentence.

marlena-b:
Thank you @kasugaijin and @nsiwnf for the review! 💪 

I will create separate issues for your suggestions and start working on them soon :)

Issue #213, opened 2023-10-01
An organization can edit, update their profile


kasugaijin:
Blocked by: https://github.com/rubyforgood/pet-rescue/issues/216

Enable each organization to edit and update their profile.
Create the necessary db table, model, route, controller and view directory to achieve this.

Acceptance criteria: 
- one can add a new organization profile in the console
- Table organization_profiles has columns: location_id (index), email, phone_number
- Router: only permits one profile per organization (see adopter profile route)
- Model: OrganizationProfile belongs to Organization, Location
- Controller: #edit and #update actions

yuricarvalhop:
Hey @kasugaijin, I could do this one

Some questions: 
Is this name the same that is currently on Organization?
Is there a planning to have multiple profiles per orga? otherwise this fields could be saved directly on the orga..

thanks!

kasugaijin:
Ok great. All yours. 
Good questions. I removed name the table columns required, as that will be on the Organization itself. An organization should have one profile. I think there's long-term benefit to splitting these into separate tables/models as the profile will likely grow a bit more complex. Also, for MVP we are creating Organizations and admin users via the console, and having to input all profile data at this stage would also be a lot more work. I see more pros than cons :)

yuricarvalhop:
@kasugaijin Thanks for the clarifications! 

Some more questions that popped up: 
Is the profile required?
Do we want to go the same approach from adopter profile? (That would mean add a `new` and `create` that can be accessed when the orga has no profile)

In case we dont want the two actions: 
should I add a migration to generate a profile for the existing orgas, or is just adding to the seeds enough?

kasugaijin:
@yuricarvalhop yes every organization should have a profile. I currently have [this PR](https://github.com/rubyforgood/pet-rescue/pull/240) so we can set up an organization via the console (creates organization, user, staff_account and add the admin role to the staff_account). Once your PR is merged, we can add the creation of necessary associations (location, profile) in this console script. 

As for the seeds, I would create the necessary associated objects eg. location and profile. 

Let me know if you want more details.

PR #252, opened 2023-10-10
Apply bootstrap_form to newly added organization_profile form

yuricarvalhop:
# 🔗 Issue
[213](https://github.com/rubyforgood/pet-rescue/issues/213)

# ✍️ Description
We added a new gem `bootstrap_form` which makes forms better readable, reducing the need for label and error specific fields. 

As this form was introduced after the gem has been introduced, this was still using the old pattern. This changes with this pr :)

# 📷 Screenshots/Demos
Before: 
image
Errors:
image

After (Already upgraded with new locale keys): 
image

Errors:
image

### DIFF
--- a/app/views/organizations/organization_profiles/_form.html.erb
+++ b/app/views/organizations/organization_profiles/_form.html.erb
@@ -1,10 +1,7 @@
-<%= form_with model: profile, :url => organization_profile_path do |form| %>
-
+<%= bootstrap_form_with model: profile, :url => organization_profile_path do |form| %>
   <% if profile.errors.count > 0 %>
     <div class="alert alert-danger mt-1" role="alert">
-      <p>
-        <%= t '.please_fix_the_errors' %>
-      </p>
+      <%= t '.please_fix_the_errors' %>
     </div>
   <% end %>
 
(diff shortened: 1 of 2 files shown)


yuricarvalhop:
@Ptrboro updating the form as we discussed

app/views/organizations/organization_profiles/_form.html.erb
```diff
@@ -50,7 +36,6 @@
             <div class="alert alert-danger mt-1" role="alert"><%= error.message %></div>
           <% end %>
```
kasugaijin, line 38: Can we keep this label? Province/State is clearer than Province state in my opinion.

app/views/organizations/organization_profiles/_form.html.erb
```diff
@@ -61,7 +46,6 @@
...
 
-          <%= location.label :city_town, 'City/Town' %>
```
kasugaijin, line 64: Same here - prefer we keep this label, too.

app/views/organizations/organization_profiles/_form.html.erb
yuricarvalhop: Hey @kasugaijin! I added the naming of the fields on the locale file, so now bootstrap_form should get the desired label. 

Its on the last commit the changes, could you please let me what you know? If you think the label should go back anyways, lemme know and Ill revert it.

Im gonna upload another image using the new translation

app/views/organizations/organization_profiles/_form.html.erb
yuricarvalhop: Commented above

app/views/organizations/organization_profiles/_form.html.erb
kasugaijin: @yuricarvalhop ah yes I see how that works now! The screenshots through me off. Yes it looks good.

kasugaijin (approved):
LGTM!

Ptrboro (approved):
🏅

Issue #244, opened 2023-10-07
Revise bootstrap_form field hints and double validation message error


reyes-dev:
Some leftover side effects of adding bootstrap_form to the project from the merge [#221](https://github.com/rubyforgood/pet-rescue/pull/221) that need addressing.

Tasks
- Move hints below labels in  [app/views/devise/registrations/edit.html.erb](https://github.com/rubyforgood/pet-rescue/pull/221/files/4c48a2fc57f5b505e6ccc14538db40cc817eaef3#diff-4d201ae5c67e8b121860359f6c67d044f5bf68c3d7a4687ca64089bfe5677a13)
- Also move the hints below the labels in [app/views/devise/registrations/new.html.erb](https://github.com/rubyforgood/pet-rescue/pull/221/files/4c48a2fc57f5b505e6ccc14538db40cc817eaef3#diff-744d99555c831425cf5883c199055bc537850fbdd075776c69dbad920497bb1f)
- Remove extra asterisk and realign remaining asterisk in [app/views/devise/registrations/new.html.erb](https://github.com/rubyforgood/pet-rescue/pull/221/files/4c48a2fc57f5b505e6ccc14538db40cc817eaef3#diff-744d99555c831425cf5883c199055bc537850fbdd075776c69dbad920497bb1f)
- Remove extra validation in [app/views/organizations/pets/_form.html.erb](https://github.com/rubyforgood/pet-rescue/pull/221/files/4c48a2fc57f5b505e6ccc14538db40cc817eaef3#diff-781e0525ee672f838f528dde170ce64634c5babfc318a22bd530907238b3424c)

reyes-dev:
I would like to handle this issue.

kasugaijin:
All yours! @reyes-dev

PR #260, opened 2023-10-12
Fix pets form field hints, asterisk, validations

reyes-dev:
# 🔗 Issue
[#244](https://github.com/rubyforgood/pet-rescue/issues/244)

# ✍️ Description
I moved the field hints below inputs for an improved appearance and removed the extra asterisk that remained in sign up page. I also removed the old validation errors in favor of the bootstrap_form validation errors, however it dawned on me that the other forms all still have hardcoded validation error messages. I'm not sure if that warrants a new issue, or how we wanted to handle that. 

For the record, if we want to continue to write our own custom validation error messages, bootstrap_form offers [options for handling validations and errors](https://github.com/bootstrap-ruby/bootstrap_form#validation-and-errors) like disabling automatic inline errors. I just assumed we wanted to let bootstrap_form handle this as well, so please let me know if I'm wrong.

# 📷 Screenshots/Demos
With old error validations
old_form_validations
Removed old error validations leaving only bootstrap_form errors
removed_old_form_validations

### DIFF
--- a/app/views/devise/registrations/edit.html.erb
+++ b/app/views/devise/registrations/edit.html.erb
@@ -34,9 +34,9 @@
             <% end %>
 
             <div class="form-group mb-3">
-              <i>(leave blank if you don't want to change it)</i><br />
               <%= f.password_field :password, autocomplete: "new-password", 
                                     class: 'form-control', label: 'New password' %>
+              <i>(leave blank if you don't want to change it)</i><br />
               <% if @minimum_password_length %>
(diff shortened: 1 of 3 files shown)


kasugaijin (commented):
Thank you!
Yes other forms will eventually need updating. However, I think it's best we wait until we have built those respective features out as things might change. I think we can let bootstrap form handle the errors for now.

kasugaijin (approved):

kasugaijin:
@all-contributors
please add @reyes-dev for code.

Issue #257, opened 2023-10-11
Staff can view checklist on pets show - tasks tab and interact with it


kasugaijin:
Blocked by: https://github.com/rubyforgood/pet-rescue/issues/256

As a staff member I can navigate to a pets show page, tasks tab, and see a checklist of items. I can add, edit, delete and mark complete each item. 

Acceptance criteria 
- visible on pets show, tasks tab
- allows CRUD of tasks
- task can be marked complete, with visual queue
- use reasonable dashboard template for this purpose (reach out to team on Slack if you have questions here)

Theme snippets can be gotten from https://geeksui.codescandy.com/geeks/docs/index.html and theme https://geeksui.codescandy.com/geeks/index.html

ErinClaudio:
@kasugaijin  Taking this one unless told otherwise.

ErinClaudio:
<img width="1118" alt="Screenshot 2023-10-11 at 2 13 21 PM" src="https://github.com/rubyforgood/pet-rescue/assets/20326770/89dffbb1-7aa5-4e51-9eb0-eadce1c6522d">
<img width="1424" alt="Screenshot 2023-10-11 at 2 13 40 PM" src="https://github.com/rubyforgood/pet-rescue/assets/20326770/2be10469-4989-46cb-87ef-b52a7bd9961a">


found this error when clicking on Task
I am working on resolving this now

ErinClaudio:
found it

ErinClaudio:
@kasugaijin 
I've encountered an issue on the pet's show page where clicking on different tabs (Overview, Task, Applications, Files) doesn't toggle visibility between sections; they all remain visible. I've considered two solutions: 1) Using JavaScript to manage tab clicks and toggle visibility, or 2) Utilizing Rails partial views to load the relevant section based on the selected tab dynamically. Do you prefer one of these approaches, or is there another method you'd recommend?

kasugaijin:
@ErinClaudio 
FYI I assigned you to the precursor ticket https://github.com/rubyforgood/pet-rescue/issues/256 yesterday...that's the background for the checklist work. My bad for not commenting and @ing you on there. I thought assigning would notify.

I am thinking this ticket https://github.com/rubyforgood/pet-rescue/issues/242 is the solution to the partials piece. 

How do we want to tackle this? My thinking is the work is to be done in this order: 
https://github.com/rubyforgood/pet-rescue/issues/242, then
[...]

PR #265, opened 2023-10-13
Staff can view checklist on pets show and interact

ErinClaudio:
# 🔗 Issue
[#257](https://github.com/rubyforgood/pet-rescue/issues/257)

# ✍️ Description

This PR introduces a new checklist feature for staff on the pet show page, allowing them to view and interact with tasks related to a pet's care. Key updates include:

- Addition of a Tasks Model
- Addition of a TasksController and within the Organizations module to handle task management.
- Expansion of the Pet model to incorporate task associations.
- Creation and update of helper methods and views for tasks.
- Establishment of a new database migration to define the tasks structure.
- Enhancement of routes to include task-related actions.
- Update of tests to cover new functionality and ensure integrity.

# 📷 Screenshots/Demos

https://github.com/rubyforgood/pet-rescue/assets/20326770/45ed860c-2ccb-4688-ae9c-d45cb9b207b2

# additional notes

- I'm sorry for the repeated delays.


kasugaijin:
Hi @ErinClaudio 

This approach is great. I think it is clean and makes sense. If there are any obvious reusable components, it would be great to invest in building them now (I think you started doing so with the overview list item). 

My only comment, which given this is a WIP may not be necessary, is could we make the partial take up the full width so it goes up to the dog image on the right like this?
image

app/controllers/organizations/pets_controller.rb
```diff
@@ -21,6 +21,7 @@ def edit
   end
 
   def show
+    @active_tab = ["overview", "tasks", "applications", "files"].include?(params[:active_tab]) ? params[:active_tab] : "overview"
```
aisayo, line 24: Does "overview" need to be included in the check against the active_tab params since the ternary defaults to it anyway?

ErinClaudio: @aisayo, good point. I will check

nsiwnf:
I dont think we actually need a separate checklist_item table actually. Let's see if we can keep track of tasks as "checklist_template_items" and the task list for the pet as "checklist_template". So maybe the pet model has a corresponding checklist_template
tasks_mock

app/models/pet.rb
```diff
@@ -30,6 +30,7 @@ class Pet < ApplicationRecord
   acts_as_tenant(:organization)
 
   has_many :adopter_applications, dependent: :destroy
+  has_many :tasks, dependent: :destroy
```
nsiwnf, line 33: I think we can reuse "checklist_template_items" here

ErinClaudio: @nsiwnf Thank you - I will get to work on a Task partial that more closely matches the above image. thank you for the clarification.

ErinClaudio:
@nsiwnf. closer, I will have more done on Monday. 
I have the crud features working, but it looks bad. And I want the edit feature to work like the create as modal. 
I also have some code clean up to work on.

app/controllers/organizations/pets_controller.rb
ErinClaudio: Thank you change made

ErinClaudio:
@kasugaijin @edwinthinks 

Editing a task incorrectly generates a nested task list, which requires a refresh to correct. Additionally, the delete button isn't immediately reflecting changes, again needing a refresh to update. This is due to my lack of ability in Turbo, but I'm unsure. 
I'm looking for some advice on how to eliminate the need for refreshing in these scenarios. Your expertise would be appreciated.
Screenshot 2023-11-02 at 1 55 25 PM

On a separate note, I've made minor tweaks to my code and partials that could benefit from your keen eye. A quick review from you would be invaluable.
Thank you

db/migrate/20231017180938_create_tasks.rb
```diff
@@ -0,0 +1,12 @@
...
+    create_table :tasks do |t|
+      t.string :name
+      t.text :description
+      t.boolean :completed, default: false
+      t.references :pet, null: false, foreign_key: true
+
+      t.timestamps
+    end
+  end
+end
```
nsiwnf, line 12: We have a couple checklist tables I think we were intending to reuse here instead of creating a new table. `checklist_template_items` and `checklist_template`.

ErinClaudio: copy that

edwinthinks:
@ErinClaudio here is a quick video going over what was going on https://www.loom.com/share/3a941a5184854839b34be0ad268020c3 -- the key takeaway is to be careful about the placement of the turbo frame tags. Let me know if that answers your questions.

edwinthinks (changes requested):

app/controllers/organizations/tasks_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+    render partial: "form", locals: {task: @task}
+  end
+
+  def create
+    @task = @pet.tasks.build(task_params)
+    if @task.save
+      redirect_to pet_path(@pet, active_tab: @active_tab)
+    else
+      render :new
+    end
```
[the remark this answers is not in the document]
ErinClaudio: Thank you.

app/views/organizations/tasks/_tasks.html.erb
```diff
@@ -0,0 +1,65 @@
...
+  <div class="row">
+    <div class="col-12 mb-4">
+      <div class="card">
+        <div class="card-header">
+          <!-- ... (rest of the card header) ... -->
+        </div>
+        <div class="card-body">
+          <h5>Tasks</h5>
+          <!-- This turbo-frame encapsulates the actual list of tasks -->
+          <turbo-frame id="tasks_list">
```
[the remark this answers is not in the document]
ErinClaudio: This has been removed

app/controllers/organizations/tasks_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+      respond_to do |format|
+        format.html { render :edit }
+      end
+    end
+  end
+
+  def destroy
+    @task = Task.find(params[:id])
+    @task.destroy
+    redirect_to pet_path(@pet), notice: "Task was successfully deleted."
```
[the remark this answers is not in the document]
ErinClaudio: Thank you this has been adjusted

app/assets/stylesheets/tabs.css
```diff
@@ -0,0 +1,10 @@
+.nav-link {
+  color: #888;
+  /* other styles for non-active tabs */
+}
+
+.nav-link.active {
+  color: #333;
+  font-weight: bold;
+  /* other styles for the active tab */
+}
```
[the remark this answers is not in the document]
ErinClaudio: Done this has been removed

ErinClaudio:
@kasugaijin @edwinthinks 

Considering the multi-tenant architecture of our application, I've introduced a new tasks table to establish a link with the pets table, which in turn connects to the organization table. As I'm still getting acquainted with the intricacies of multi-tenant applications, I'm concerned that the existing structure of checklist_template_items may not offer the same level of relational access. @nsiwnf, given your experience, could you confirm if my understanding is correct? This would greatly help before implementing this change and writing the corresponding tests.

Thank you

db/migrate/20231017180938_create_tasks.rb
```diff
@@ -0,0 +1,12 @@
+class CreateTasks < ActiveRecord::Migration[7.0]
+  def change
+    create_table :tasks do |t|
+      t.string :name
```
edwinthinks, line 4:
```suggestion
      t.string :name, null: false
```

edwinthinks (approved):
One minor change on the migration otherwise let's get this IN!

kasugaijin (approved):
Looks great, thanks Erin!

Issue #268, opened 2023-10-14
Staff can upload a profile pic


nsiwnf:
As staff, I can upload an avatar image
As staff, I can see my image in the log out/settings menu at the top right corner

As admin, I can see staff's images in the "Staff" index page 

Include a default image if staff has not uploaded an image

yuricarvalhop:
@nsiwnf can I work on this one?

yuricarvalhop:
@nsiwnf should Its worth it to move this logic into a concern, or you dont think we are gonna need this logic somewhere else?

nsiwnf:
Good question - we might want it for `adopters` as well so a concern would be a good approach!

PR #276, opened 2023-10-15
268 part 1: staff can upload profile pic

yuricarvalhop:
# 🔗 Issue
[268](https://github.com/rubyforgood/pet-rescue/issues/268)

# ✍️ Description
We want to allow staff members to be able to change their profile pictures. 

This is the part 1, with the concern, model changes and form changes. I decided to break this into two, because I want to know what people think about the naming and the current decision implementations, before I change all the mini avatar images. 

Notice that I put the relation under users table, not staff account. Please honest thoughts here. Thanks!

Part 2: Add view helper to change the places where we use this (navbar/staff index)

# 📷 Screenshots/Demos
With image:
image

Without image: 
image

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -33,7 +33,8 @@ class RegistrationsController < Devise::RegistrationsController
       :password,
       :password_confirmation,
       :signup_role,
-      :current_password)
+      :current_password,
+      :append_picture)
   end
 
   def after_sign_up_path_for(resource)
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -34,7 +34,7 @@ class RegistrationsController < Devise::RegistrationsController
       :password_confirmation,
       :signup_role,
       :current_password,
-      :append_picture)
+      :append_avatar)
   end
 
   def after_sign_up_path_for(resource)
--- a/app/models/concerns/avatarable.rb
+++ b/app/models/concerns/avatarable.rb
@@ -1,13 +1,13 @@
 module Avatarable
   extend ActiveSupport::Concern
-  
+
   included do
     has_one_attached :avatar
-  
+
     validates :avatar, content_type: {in: ["image/png", "image/jpeg"],
-                                               message: "must be PNG or JPEG"},
-                                size: {between: 10.kilobyte..1.megabytes,
-                                       message: "size must be between 10kb and 1Mb"}
+                                      message: "must be PNG or JPEG"},
+      size: {between: 10.kilobyte..1.megabytes,
+             message: "size must be between 10kb and 1Mb"}
   end
 
   def append_avatar=(attachable)
--- a/app/models/concerns/avatarable.rb
+++ b/app/models/concerns/avatarable.rb
@@ -1,16 +1,16 @@
-module Pictureable
+module Avatarable
   extend ActiveSupport::Concern
   
   included do
-    has_one_attached :picture
+    has_one_attached :avatar
   
-    validates :picture, content_type: {in: ["image/png", "image/jpeg"],
+    validates :avatar, content_type: {in: ["image/png", "image/jpeg"],
                                                message: "must be PNG or JPEG"},
                                 size: {between: 10.kilobyte..1.megabytes,
                                        message: "size must be between 10kb and 1Mb"}
   end
 
-  def append_picture=(attachable)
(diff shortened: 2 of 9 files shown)


test/models/user_test.rb
```diff
@@ -3,6 +3,34 @@
 require "test_helper"
 
 class UserTest < ActiveSupport::TestCase
+  context "avatarable" do
```
yuricarvalhop, line 6: Should I move this into a test helper concern?

app/models/user.rb
```diff
@@ -33,6 +33,8 @@
 #  index_users_on_reset_password_token  (reset_password_token) UNIQUE
 #
 class User < ApplicationRecord
+  include Avatarable
```
yuricarvalhop, line 36: Opinions on the table I added it, please

app/models/concerns/avatarable.rb
```diff
@@ -0,0 +1,16 @@
+module Avatarable
```
yuricarvalhop, line 1: Opinions on the naming also welcome

yuricarvalhop:
@nsiwnf we talked about it, so if you could gimme your thoughts, thanks!

app/models/concerns/avatarable.rb
nsiwnf: Yeah, I like the "avatar" naming

nsiwnf:
This looks good to me - admittedly I don't have too much experience adding a feature like this and just looked up some tutorials :)

Any thoughts @kasugaijin, @edwinthinks?

test/models/user_test.rb
kasugaijin: Yes I think so!

app/models/concerns/avatarable.rb
kasugaijin: Yes same!

app/controllers/registrations_controller.rb
```diff
@@ -33,7 +33,8 @@ def account_update_params
       :password,
       :password_confirmation,
       :signup_role,
-      :current_password)
+      :current_password,
+      :append_avatar)
```
kasugaijin, line 37: name it `avatar` instead? noun rather than an action in the params.

kasugaijin:
@yuricarvalhop @nsiwnf I think this implementation works fine. Some of my thoughts, more for discussion:
- having avatars related to the User model does make sense. Let's say we want to also allow adopters to have avatars, and a situation where a user is both staff and an adopter, then it makes sense they can use the same avatar for both use cases. 
- given we are relating to the user model, I am not sure it needs to go into a concern? I cannot think of another model where we would use an avatar, but I could be completely wrong here, and I am not opposed to it living in the concern at all. It keeps the options open.

yuricarvalhop:
Hi @kasugaijin @nsiwnf 
thanks for the comments, some thoughts: 

If we keep the avatar on the User, and cant imagine any further use for it on the future, then indeed it wouldn't need to be a concern. Letting it there would be only for code organization purposes, but I dont know if its worth it, as it adds complexity to organize such a small piece of logic (If you think this is a gain, or you think the situation might come on the future, we can let it as it is). 

so, how should I proceed? :)

kasugaijin:
@yuricarvalhop I agree, but I can't help but think we might end up using this in future to create avatars for an organization, for example. It might make sense to just compartmentalize this in the concern now so that it can be used in the future for other models. I think we can leave this PR as is :)

test/models/user_test.rb
yuricarvalhop: done!

yuricarvalhop:
@kasugaijin alright! I added a spec module, as commented above. Feel free to review it, Ill follow up with part 2 on the weekend :)

test/shared/avatarable_shared_tests.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
```
kasugaijin, line 2: is this RSpec? We are using Minitest :)

test/shared/avatarable_shared_tests.rb
```diff
@@ -0,0 +1,35 @@
...
+module AvatarableSharedTests
+  def self.included spec
```
kasugaijin, line 4: I thought this was RSpec for a minute until I realised it's using the shoulda gem!

kasugaijin (approved):
Looks good! Thank you.

PR #287, opened 2023-10-18
Fix flaky tests

Ptrboro:
# ✍️ Description
Once `default_url_options[:script_name]` was set (e.g. by calling `set_organization` [here](https://github.com/rubyforgood/pet-rescue/blob/main/test/integration/adoptable_pets_index_test.rb#L7)) it was then leaking to other tests. This change ensures that the script_name is cleared after each test so it should fix the majority of flaky tests.


kasugaijin (commented):
Good catch.

kasugaijin (approved):

Issue #285, opened 2023-10-18
Add org information to organization_profiles table


jadekstewart3:
Add Organization attribute to the organization profiles table.
Completion Criteria:
- [ ] about_us

PR #289, opened 2023-10-19
Add column 2 org profiles

jadekstewart3:
# 🔗 Issue
#285 

# ✍️ Description
Added an about_us column on the organization_profiles table. Updated factories and seeds to reflect the migration. Added validation for about us on organization_profile model, and added validation test.

# 📷 Screenshots/Demos

### DIFF
--- a/app/models/organization_profile.rb
+++ b/app/models/organization_profile.rb
@@ -29,7 +29,7 @@ class OrganizationProfile < ApplicationRecord
 
   before_save :normalize_phone
 
-  validates :phone_number, phone: {possible: true}
+  validates :phone_number, phone: {possible: true, allow_blank: true}
 
   private
 
--- a/app/models/organization_profile.rb
+++ b/app/models/organization_profile.rb
@@ -3,6 +3,7 @@
 # Table name: organization_profiles
 #
 #  id              :bigint           not null, primary key
+#  about_us        :string           not null
 #  email           :string
 #  phone_number    :string
 #  created_at      :datetime         not null
--- a/app/models/organization_profile.rb
+++ b/app/models/organization_profile.rb
@@ -3,7 +3,7 @@
 # Table name: organization_profiles
 #
 #  id              :bigint           not null, primary key
-#  about_us        :string           
+#  about_us        :string
 #  email           :string
 #  phone_number    :string
 #  created_at      :datetime         not null
--- a/app/models/organization_profile.rb
+++ b/app/models/organization_profile.rb
@@ -3,7 +3,7 @@
 # Table name: organization_profiles
 #
 #  id              :bigint           not null, primary key
-#  about_us        :string           not null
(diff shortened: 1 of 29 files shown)


app/models/organization_profile.rb
```diff
@@ -30,6 +31,7 @@ class OrganizationProfile < ApplicationRecord
   before_save :normalize_phone
 
   validates :phone_number, phone: {possible: true, allow_blank: true}
+  validates :about_us, presence: true
```
kasugaijin, line 34: can we for now not validate the presence of this? It will mean we have to provide more information when creating a new org, which is done by the console, and already has a fair bit of data required.

kasugaijin (approved):
Looks good to me! If we can address that one validation.

app/models/organization_profile.rb
edwinthinks: Agreed with @kasugaijin here.

My rule about adding validations is to ask wither or not the record is valid if it is missing this the attribute in question. A profile could still be legit if it is missing a `about_us` section in my mind. In other cases maybe there are attributes that are so important that if its missing we shouldn't save those records. A common one is a User without a email if we rely on the email heavily.

db/migrate/20231019155058_add_about_us2_organization_profiles.rb
```diff
@@ -0,0 +1,5 @@
+class AddAboutUs2OrganizationProfiles < ActiveRecord::Migration[7.0]
+  def change
+    add_column :organization_profiles, :about_us, :string, null: false
```
edwinthinks, line 3:
```suggestion
    add_column :organization_profiles, :about_us, :string
```
In case we want to remove this field's validation.

kasugaijin:
https://github.com/all-contributors
please add @yuricarvalhop for code.
please add @edwinthinks for review.

kasugaijin:
https://github.com/all-contributors
please add @jadekstewart3 for code.
please add @edwinthinks  for review.

kasugaijin:
@all_contributors
please add @jadekstewart3 for code.
please add @edwinthinks for review.

kasugaijin:
@all-contributors
please add @jadekstewart3 for code.
please add @edwinthinks for review.

db/migrate/20231003182930_add_species2_pets.rb
```diff
@@ -1,5 +1,5 @@
 class AddSpecies2Pets < ActiveRecord::Migration[7.0]
   def change
-    add_column :pets, :species, :integer, null: false
```
nsiwnf, line 3: This is an existing migration right? I though that this was already in `main`

jadekstewart3: this is true... how the heck did that get added agian?

nsiwnf (approved):

Issue #284, opened 2023-10-18
Organization Dashboard


jadekstewart3:
We need to set up a route that points to a show action in the Organizations Dashboard controller.

Completion Criteria:
- [x] Route that points to a index action in the organization dashboard controller
- [x] index action in the organization dashboard controller
- [x] Organization dashboard utilizes [this geeks ui layout](https://geeksui.codescandy.com/geeks/pages/profile-edit.html)
- [x] Displays organization name

kasugaijin:
Some clarification for the UI markup, I think we'd want to use this component within our existing admin dashboard template
<img width="969" alt="image" src="https://github.com/rubyforgood/pet-rescue/assets/95949082/2de79d37-d8ad-4272-984f-3140a7144dc5">

PR #290, opened 2023-10-19
Org dashboard

jadekstewart3:
# 🔗 Issue
#284  

# ✍️ Description
I tracked down where the organization dashboard was being routed. It was present, but being routed through the index action. I updated the index action to use instance variables scoped to the current user, and imported and formatted the html associated with the desired template. Added in organization name, and personalized welcome message to the user. 

I am unsure what exactly we want in the card for the dashboard I think some more discussion might need to take place, but I'm happy to make any changes suggested :)

# 📷 Screenshots/Demos

Screenshot 2023-10-19 at 1 42 19 PM

### DIFF
--- a/app/controllers/organizations/dashboard_controller.rb
+++ b/app/controllers/organizations/dashboard_controller.rb
@@ -3,6 +3,8 @@ class Organizations::DashboardController < Organizations::BaseController
   before_action :verified_staff
 
   def index
+    @user = current_user
+    @organization = @user.organization
     @hide_footer = true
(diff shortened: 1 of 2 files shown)


kasugaijin (approved):
Looks good to me!

kasugaijin:
@all-contributors
please add @edwinthinks for code.

kasugaijin:
Added Edwin in this PR as it didn't work when I tried in your migration PR :)

PR #296, opened 2023-10-22
Improve tests stability

Ptrboro:
# ✍️ Description
This PR addresses some inconsistencies in our test suite to enhance its stability and the overall developer experience. 

Here's a summary of the improvements made:
1. Changed the `setup` function in test_helper.rb to ensure it is called before the `setup` block in individual test files.
2. Resolved issues of Organizations generated with factories having duplicate slugs.
3. Clarified interconnected associations in factories to ensure consistency:
  ```
  match = create(:match)
  match.organization == match.pet.organization # => now true
```
4. Resolved issues of Users generated with factories having duplicate emails.
5. Streamlined the behavior of factories with ActsAsTenant:
```
user = create(:user)
user.organization == ActsAsTenant.test_tenant # => true
Organization.count # => 1 (was 2 before the fix)
```
6. Reconfigured ActsAsTenant for tests according to the [docs](https://github.com/ErwinM/acts_as_tenant#testing)

### DIFF
--- a/test/factories.rb
+++ b/test/factories.rb
@@ -1,11 +1,19 @@
 FactoryBot.define do
   factory :adopter_account do
-    user
+    transient do
+      organization {}
+    end
+
+    user do
+      if organization
+        association :user, organization: organization
+      else
+        association :user
+      end
+    end
 
     trait :with_adopter_profile do
-      after :create do |account|
-        create(:adopter_profile, adopter_account: account)
-      end
+      adopter_profile { association :adopter_profile, adopter_account: instance}
     end
   end
 
@@ -81,16 +89,14 @@ FactoryBot.define do
     zipcode { Faker::Address.zip_code }
 
     trait :with_adopter_profile do
-      after :create do |location|
-        create(:adopter_profile, location: location)
-      end
+      adopter_profile
     end
   end
 
   factory :organization do
     name { Faker::Company.name }
     sequence(:slug) { |n| Faker::Internet.domain_word + n.to_s }
-    profile { build(:organization_profile) }
+    profile { association :organization_profile, organization: instance }
   end
 
   factory :organization_profile do
@@ -98,12 +104,7 @@ FactoryBot.define do
     phone_number { Faker::PhoneNumber.phone_number }
     about_us { Faker::Lorem.paragraph(sentence_count: 4) }
     location
-
-    trait :with_organization do
-      after :build do |profile|
-        profile.organization = create(:organization, profile: profile)
-      end
-    end
+    organization { association :organization, profile: instance }
   end
 
   factory :pet do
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: app/controllers/organizations/invitations_controller.rb [hunk 1]
```diff
@@ -1,4 +1,8 @@
 class Organizations::InvitationsController < Devise::InvitationsController
+  include OrganizationScopable
+  # TODO: remove next line once we fix links in the invitation emails
+  skip_before_action :set_tenant, except: [:new, :create]
```
Ptrboro, line 4: After tweaking the ActsAsTenant setup in our tests, one of the tests start failing. There was an actual bug and `StaffAccount` kept linking itself to the organization with an id of 1. Including `OrganizationScopable` fixes that up, but we should probably drop the default value from the staff_account.organization_id column so we don't miss similar issues down the road.

test/factories.rb
```diff
@@ -163,46 +162,33 @@
...
-      end
-    end
-
-    trait :application_awaiting_review do
-      adopter_account
-
-      after :create do |user|
-        create :adopter_application, adopter_account: user.adopter_account, status: 0
-        user.adopter_account.adopter_profile = create(:adopter_profile)
-      end
```
Ptrboro, line 205: This trait was creating 4 different organizations under the hood. It was hard to fix so I decided it's better to remove it and build these associations in the test itself (test/integration/adoptable_pet_show_test.rb)

test/test_helper.rb
```diff
@@ -26,12 +26,12 @@ def set_organization(organization)
     Rails.application.routes.default_url_options[:script_name] = "/#{organization.slug}"
   end
 
-  def setup
-    ActsAsTenant.current_tenant = create(:organization, slug: "test")
+  setup do
```
Ptrboro, line 29: The issue was that `def setup` is called after the `setup` blocks in individual tests:

```
# test/test_helper.rb
def setup
  puts "I'm in test_helper.rb!"
end

# test/some_dummy_test.rb
setup do
  puts "I'm in dummy_test.rb!"
end

# > I'm in dummy_test.rb!
# > I'm in test_helper.rb!
```

By converting it to `setup do; ... end` we fix the order issue :) 

```
# test/test_helper.rb
setup do
  puts "I'm in test_helper.rb!"
end

# test/some_dummy_test.rb
setup do
  puts "I'm in dummy_test.rb!"
end

# > I'm in test_helper.rb!
# > I'm in dummy_test.rb!
```

This is important because we create an organization in this setup and assign it as current test tenant.

kasugaijin:
Very cool @Ptrboro thank you for looking into this. I need a bit more time to digest what's going on so unless someone beats me to it I review this week.

test/factories.rb
```diff
@@ -81,29 +89,22 @@
...
+      adopter_profile
     end
   end
 
   factory :organization do
     name { Faker::Company.name }
-    slug { Faker::Internet.domain_word }
-    profile { build(:organization_profile) }
+    sequence(:slug) { |n| Faker::Internet.domain_word + n.to_s }
```
kasugaijin, line 98: so we append a number here just to ensure it's unique? i.e., in the slim case Faker creates the same word twice.

kasugaijin (approved):
Looks great to me! For the most part you're making the tenant instance available and explicitly defining which org to associate to in the test data used by Faker. It makes sense to me and I cannot see anything I would change. This looks like a far more robust approach.

test/factories.rb
Ptrboro: Exactly :) One of the tests I was debugging was sometimes failing because of this (but I can't remember now which one).

PR #313, opened 2023-10-28
Prevent bootstrap base styling to conflict with geeks theme

edwinthinks:
# 🔗 Issue
N/A

# ✍️ Description
I noticed that some styling from the theme wasn't appearing correctly. Specifically padding on the table was incorrect:
Screenshot 2023-10-28 at 08 16 50

It should be more like this with more spacing to make it look less cluttered:
Screenshot 2023-10-28 at 08 17 28

It turns out that we were importing bootstrap when actually we don't need to!

# 📷 Screenshots/Demos
See above


kasugaijin (approved):
Ah good catch! LGTM.

PR #316, opened 2023-10-29
Geocode gem

kasugaijin:
# 🔗 Issue
No issue

# ✍️ Description
Currently geocode is run every time a Location is validated (this hits the Google API with a street address and gets lat and lon in return and saves to Location. The API was being hit while running seed.rb and causing this log output: `Geocoding API not responding fast enough (use Geocoder.configure(:timeout => ...) to set limit).`. So, to prevent the API call I added a proc to the callback to always run geocode in prod any time there is a validation, but to only run in development if there are no lat or lon. Added lat, lon to seeds so geocode does not run. Now seeds loads without the above log.

I am open to other ideas to handle this if there is something better. I considered config in the initializer, but since this is the only model using it (and likely the only one using it) it makes sense to keep the logic close to the model. I considered removing Geocode entirely as I don't know if we intend to use it, however, I think it makes more sense to keep it in case we want to use it in the future, and just configure it.

**Update:**
Given the use of Geocode and lat/lon are not required in the MVP it makes sense to remove this additional logic. This PR does that.

# 📷 Screenshots/Demos


edwinthinks:
@kasugaijin could you remind me what the lat/lon does for the application? If it isn't needed, I would just remove it tbh

kasugaijin:
> @kasugaijin could you remind me what the lat/lon does for the application? If it isn't needed, I would just remove it tbh

@edwinthinks it was for the Successes page that showed a pin on a Google map for every dog adoption https://www.bajapetrescue.com/successes

If this is something we don't think we will use I can get rid of it.

edwinthinks:
@kasugaijin I'd get rid of it if we can't think of any way it can provide value. We can always bring it back when we see value in it but I can't forsee how the lat/lng is going to help our stake holders or applicants

kasugaijin:
> @kasugaijin I'd get rid of it if we can't think of any way it can provide value. We can always bring it back when we see value in it but I can't forsee how the lat/lng is going to help our stake holders or applicants

PR updated to remove genocide :)

Ptrboro (approved):
LGTM! 💪 

We could also clean up the routes (line 31) and the Successes link in the menu :)

Screenshot 2023-11-01 at 15 53 02

kasugaijin:
> LGTM! 💪
> 
> We could also clean up the routes (line 31) and the Successes link in the menu :)
> 
> Screenshot 2023-11-01 at 15 53 02

Ah yes! Thank you

PR #317, opened 2023-10-29
Adjust design on pets#show page to match more closely to theme

edwinthinks:
# 🔗 Issue
N/A

# ✍️ Description
I noticed a fews things and fixed them:
- Nav tabs were being rendered within the page content whereas it should be near the header. Addressed using a new `@` variable
- Remove excess padding/spacing in the page content area below the header. Removed excess "section" element
- Removed the colored ring around the pet's image. I felt this was a bit jarring and decided to remove it.

# 📷 Screenshots/Demos
Before
Screenshot 2023-10-29 at 07 56 10

After
Screenshot 2023-10-29 at 07 52 42


edwinthinks:
I want us to consider using nice_partial or some other means to enforce design consistency. One way to achieve this using nice partials is to raise errors when certain variables are not provided. For example, we probably don't want pages without a header on them so we should enforce that. Here is what I'am thinking:
```
<%= render "layouts/dashboard/page", header_title: "Edit Penny" do |p| %>  # this is okay

<%= render "layouts/dashboard/page" do |p| %>  # this is NOT okay
```

kasugaijin:
@edwinthinks yes I like that idea of using checks to enforce consistency, especially in an open source project.

kasugaijin (approved):

Issue #308, opened 2023-10-26
As staff user my edit user settings form is rendered in the dashboard layout


kasugaijin:
The devise form for edit user settings should be rendered within the dashboard layout if the user has an associated StaffAccount.

MooseCowBear:
Is this the idea of what you're looking for?

![Screenshot_2023-10-29_at_7 57 48 AM](https://github.com/rubyforgood/pet-rescue/assets/82609260/66ee0083-f619-4dcc-89c1-91ab23c004e9)

kasugaijin:
Apologies I should have provided more detail here. The link to account settings should remain in the top right avatar, but the form to edit should render within the dashboard layout. Does that make sense? No links need to be added to the dashboard sidenav.

MooseCowBear:
Oh I see. That makes far more sense.

kasugaijin:
PR up: https://github.com/rubyforgood/pet-rescue/pull/319

PR #319, opened 2023-10-30
Render staff account settings inside dashboard

BakiVernes:
# 🔗 Issue
#308 

# ✍️ Description
Switched the layout of the devise account settings page depending on whether the user is staff or not.

# 📷 Screenshots/Demos
staff 
adopter


app/controllers/registrations_controller.rb
```diff
@@ -13,6 +14,14 @@ def new
 
   private
```
kasugaijin, line 16: Looks good! Such a minor nitpick here but I'd prefer a method named `set_layout` over `choose_layout` but that's all.

app/views/devise/registrations/edit.html.erb
```diff
@@ -1,40 +1,43 @@
...
-  <div class="container">
-    <div class="text-center">
-      <h1 class="section-heading text-uppercase underline">
-        Edit my user settings
-      </h1>
-    </div>
-  </div> <!--container-->
-</header>
+<% if current_user.staff_account %>
+  <%= provide(:header_title, "Edit my user settings")  %>
```
kasugaijin, line 2: I have not used these partials yet so not sure why the header title is in a conditional? Could we just set it as `Edit my user settings` regardless of user?

BakiVernes: Since dashboard uses a shared layout it utilises a [yield](https://github.com/rubyforgood/pet-rescue/blob/main/app/views/layouts/dashboard.html.erb#L230) for it's different possible titles. The account settings form should provide different header style depending if it is rendered inside dashboard or standalone. To better illustrate what I mean, if I omit the condition the following would happen:

Screenshot 2023-10-31 at 05 05 33


In the PR description screenshot you see that this condition adds the title to the appropriate spot :)

kasugaijin: Ahh great, thank you for the explanation @BakiVernes

kasugaijin:
@all-contributors
please add @BakiVernes  for code.

BakiVernes:
@Ptrboro That was a good find. I thought about your suggestion and it occurred to me that instead of checking if the user is signed in (which he should already be if he is trying to edit his account) a better approach would be to just have the `set_layout` method be triggered on the `:edit` like so: `layout :set_layout, only: :edit`. This eliminates the need for the signed_in? check. 

Also since I first missed this error [I added tests](https://github.com/rubyforgood/pet-rescue/pull/319/files#diff-357e3103120845ab89006e6a3362fdfed9923e3a48168ad8316ce83e5d26df84) to check if the account settings rendered inside dashboard or out depending on the account of the current user.

I think we are good to go on this PR 🤘

Ptrboro:
Hey @BakiVernes! Adding `only: :edit` is much better! 👏  We also need to do this on `:update`, otherwise it does not use proper layout if there are validation errors (see recording)


https://github.com/rubyforgood/pet-rescue/assets/16762860/9754f6ef-676a-49ee-bb1c-b0946fbe2f79

kasugaijin:
> Hey @BakiVernes! Adding `only: :edit` is much better! 👏 We also need to do this on `:update`, otherwise it does not use proper layout if there are validation errors (see recording)
> 
>  Screen.Recording.2023-11-01.at.15.28.39.mov

Could you please make an issue for that @Ptrboro ? Unless @BakiVernes want's to include that here.

BakiVernes:
@kasugaijin @Ptrboro Resolved!

kasugaijin:
> @kasugaijin @Ptrboro Resolved!

Just failing on formatting. Please run `rails standard:fix` and push that up.

Ptrboro (approved):
Well done @BakiVernes! 🏅 👏

PR #326, opened 2023-11-03
Add avatar component

BakiVernes:
# ✍️ Description

Added avatar component so it properly shows avatar wherever we need it. This component includes options for different sizes, avatar source, or displaying initials if no picture is available. It also has different background colors for different initials.

An example of how to use it:

`<%= render AvatarComponent.new(alt: current_user.first_name.first.upcase, src: current_user.avatar) %>`

This component shows a picture if the user has an avatar. If the user has no avatar it shows it's initial with a colored background.

I already added the component to the navbar/dashboard/dropdown.

# 📷 Screenshots/Demos
imageimage

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,6 +27,9 @@ gem "turbo-rails"
 # Hotwire's modest JavaScript framework [https://stimulus.hotwired.dev]
 gem "stimulus-rails"
 
+# A framework for creating reusable, testable & encapsulated view components
+gem "view_component"
+
 # Build JSON APIs with ease [https://github.com/rails/jbuilder]
 gem "jbuilder"
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -443,6 +443,10 @@ GEM
     tzinfo (2.0.6)
       concurrent-ruby (~> 1.0)
     unicode-display_width (2.4.2)
+    view_component (3.7.0)
+      activesupport (>= 5.2.0, < 8.0)
+      concurrent-ruby (~> 1.0)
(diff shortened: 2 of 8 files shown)


kasugaijin (approved):

kasugaijin:
looks good thank you. I will make an issue to add to the staff index page as well.

edwinthinks:
Thanks @BakiVernes  for your contribution! I like ViewComponents and I introduced at my work. I'am debating however if this is the right technical choice for this project. When deciding between ViewComponents and nice_partials, we opted to use nice_partials because it's similarity with regular rails partials. My reservation from using ViewComponent is forcing newer contributors to learn a new syntax for handling view logic and thus causing contributions to take longer.

I'am working with the internal team to decide if we should reverse that decision or continue with this. Apologies for the back & forth on this one but I feel it is important to re-exam here. 

What do you think @BakiVernes ?

kasugaijin:
@BakiVernes my fault. I initiated the discussion on this but made the decision to merge a little too soon. I think this is worth some consideration to streamline implementations and approaches within an open source app.

BakiVernes:
I agree @edwinthinks. I just saw something in the app that I wanted to touch up and did it without consider the broader picture. Since I am new to this project I don't know the in and outs of the architecture so bear with me if I make and oversight like this.

I'll gladly make another PR to remedy this and use nice_partials for this feature if you'd like.

kasugaijin:
Hey @BakiVernes thank you for taking the initiative! Are you open to doing the same but with nice_partials? That would be awesome.

PR #333, opened 2023-11-05
Establish a common approach for rendering pages and their requirements in the dashboard layout

edwinthinks:
# 🔗 Issue
N/A

# ✍️ Description
This PR introduces a new component "components/dashboard/page" which is used to enforce the styling of dashboard pages. Notice that it handles the placement of all the elements and also enforces that a `header_title` must be provided.

This isn't trying to replace all the views with the new approach but rather introduces it. We can in a seperate PR implement this pattern once it is reviewed & approved

# 📷 Screenshots/Demos
Screenshot 2023-11-05 at 15 55 07
Screenshot 2023-11-05 at 15 55 11


kasugaijin (approved):

Ptrboro (approved):
Love it!

Leaving one comment, feel free to ignore, we can fix it later :)

Issue #282, opened 2023-10-18
Add status to pets table


jadekstewart3:
Create a migration that will add a status to the pets table that allows the designation of the pets status "adoptable", "fosterable", or "adoptable and fosterable".

Completion Criteria:
- [ ] As an organization member, when I visit "/organization/pets/:id/edit" I see a field on that form to edit the status of the pet. After I change and update the status of a pet, when I visit "/organization/pets/:id", I can see the pets updated status displayed in the summary card.

reyes-dev:
I would like to be assigned this issue. A few quick questions, in case I get assigned:

1. I don't see an adoptable_pets table in the schema, do you mean the pets table?
2. Should the new column be named "status" or something like "adoptable_status"?

jadekstewart3:
@reyes-dev Hi! You're right. It should be on the pets table! Sorry about the confusion, I'll fix the ticket itself. I was originally thinking just status, but erring on the side of specificity makes sense also! 
@kasugaijin @edwinthinks @nsiwnf Do any of you have strong opinions on the title of this row?

kasugaijin:
I get where you're going with status but it is quite generic. I wonder if something like `availability` or `placement_type` or `adoptable_fosterable`. Ideally one could look at the column name and have a good chance of guessing what it's for.

jadekstewart3:
@reyes-dev Are you still working on this?

reyes-dev:
@jadekstewart3 I wasn't assigned so I haven't done any work on it.

jadekstewart3:
@reyes-dev oh! do you still want to work on it? I can assign you right now

PR #337, opened 2023-11-10
282 pet status

jadekstewart3:
# 🔗 Issue
#282 

# ✍️ Description
~WIP: **Placement type change after edit is not being updated**~  **Fixed**

Changes Made:
- Added a new column, placement_type, to the pets table for specifying the type of placement a pet is eligible for: "adoptable," "fosterable," or "adoptable and fosterable."
- Implemented an enum in the Pet model to handle the placement type variations.
- Introduced corresponding model tests to ensure the accuracy of the implemented changes.
- Updated the shared form at "/organization/pets/:id/edit" to include a field enabling staffers to modify and update the placement_type of the pet.
- Modified the summary card at "/organization/pets/:id/" to prominently display the placement type for each pet.

# 📷 Screenshots/Demos
Screenshot 2023-11-10 at 1 43 52 PM
Screenshot 2023-11-10 at 1 44 41 PM


app/models/pet.rb
```diff
@@ -33,6 +34,7 @@ class Pet < ApplicationRecord
   has_one :match, dependent: :destroy
   has_many_attached :images
   enum species: ["dog", "cat"]
+  enum placement_type: ["adoptable", "fosterable", "adoptable and fosterable"]
 
   validates :name, presence: true
```
kasugaijin, line 39: I think it would be a required field for the app to function properly, so we should validate presence?

db/migrate/20231110194816_add_status2_pets.rb
```diff
@@ -0,0 +1,5 @@
+class AddStatus2Pets < ActiveRecord::Migration[7.0]
+  def change
+    add_column :pets, :placement_type, :integer
```
kasugaijin, line 3: Per my comment above, this would be a good candidate for null: false?

app/views/organizations/pets/show.html.erb
```diff
@@ -67,6 +67,22 @@
...
+                               align-items-center">
+                  <div class="d-flex align-items-center">
+                    <i class="fe fe-book me-2"></i>
+                    <div class="ms-2">
+                      <h5 class="mb-0 text-body">Placement Type</h5>
+                    </div>
+                  </div>
+                  <div>
+                    <div>
+                      <p class="text-dark mb-0 fw-semibold"><%= @pet.placement_type %></p>
```
kasugaijin, line 81: Can we capitalize the first letter? May as well do the `Sex` field while you are at it :)

kasugaijin (approved):

Issue #307, opened 2023-10-26
As staff I can view adopter applications for a given pet via the /pet/:id Applications tab


kasugaijin:
Currently, all applications are viewable at `/adoption_application_reviews`
We want to also see the applications for a specific pet on that pet's show page under the Applications tab

Acceptance criteria
- a pet's adopter applications are shown on /pet/:id Applications tab
- follows implementation to render tab partials used for checklists ([PR](https://github.com/rubyforgood/pet-rescue/pull/265) still in progress)

MooseCowBear:
I'd like to take a stab at this.

MooseCowBear:
2 quick questions...

It looks like as of now there's no application partial, so my thought is to move the current application rendering that appears in adoption_application_reviews/index.html.erb into a partial that can be shared by both that view and the applications tab partial for the pet show view. Is it okay to go ahead and do that and then update the adoption_application_reviews/index.html.erb accordingly?

Second, it looks like the controller action for the tabs is going to be taken care of in pull #265 and that this line in the pets show view will render the partial for the selected tab:

![Screenshot_2023-11-07_at_7 05 49 AM](https://github.com/rubyforgood/pet-rescue/assets/82609260/55a8aefb-5d88-4244-b72a-cdcc9f825b59)

Does this mean I should add the tab partial for the pet's applications in the organizations/tasks directory?

MooseCowBear:
As of now, I've created the shared partial. It leaves the layout of adoption applications reviews index unchanged. For the applications tab on the show page, I could see the partial rendered in a card like: 

<img width="973" alt="Screenshot 2023-11-10 at 12 14 46 PM" src="https://github.com/rubyforgood/pet-rescue/assets/82609260/c4bc6178-9d16-4a21-b9ed-6f2b2f153c9b">

Or without the header: 

<img width="986" alt="Screenshot 2023-11-10 at 12 14 30 PM" src="https://github.com/rubyforgood/pet-rescue/assets/82609260/69baef7c-72d2-4543-8c60-f9497201e2ac">

Either way, that's what it would look like if dropped into the tasks/applications.html.erb file of pull #265. Since that pull hasn't been merged yet, I could put the card in its own partial file? Or hold off until that pull has been merged? Let me know.

kasugaijin:
Great! I think there's nothing to stop you carring on here. If yours is merged first, then @ErinClaudio can just accept the incoming on a conflict to overwrite his current file. 

[...]

PR #338, opened 2023-11-13
307 view adopter applications for pet

MooseCowBear:
# 🔗 Issue

[#307](https://github.com/rubyforgood/pet-rescue/issues/307)

# ✍️ Description

Creates a pet applications partial. In order to have the applications appear on the pet show page, I included the parts of 265 to create applications tab which didn't yet exist and the controller action to make them function. 

# 📷 Screenshots/Demos

Show page, applications tab: 

Screenshot 2023-11-13 at 5 28 38 PM

Applications review index page: 

Screenshot 2023-11-13 at 5 28 47 PM

### DIFF
--- a/app/controllers/organizations/pets_controller.rb
+++ b/app/controllers/organizations/pets_controller.rb
@@ -22,6 +22,7 @@ class Organizations::PetsController < Organizations::BaseController
   end
 
   def show
+    @active_tab = determine_active_tab # from 265
     @pause_reason = @pet.pause_reason
     return if pet_in_same_organization?(@pet.organization_id)
 
@@ -90,5 +91,9 @@ class Organizations::PetsController < Organizations::BaseController
     @nav_tabs = [
       {name: "Summary", path: pet_path(@pet)}
     ]
+  end
+
+  def determine_active_tab # from 265
+    ["tasks", "applications", "files"].include?(params[:active_tab]) ? params[:active_tab] : "overview"
   end
 end
--- a/app/models/adopter_application.rb
+++ b/app/models/adopter_application.rb
@@ -58,4 +58,8 @@ class AdopterApplication < ApplicationRecord
     adopter_application.status = :withdrawn
     adopter_application.save
   end
+
+  def applicant_name
+    "#{adopter_account.user.last_name}, #{adopter_account.user.first_name}"
+  end
 end
--- a/app/models/pet.rb
+++ b/app/models/pet.rb
@@ -92,7 +92,7 @@ class Pet < ApplicationRecord
 
   # all pets under an organization with applications and no adoptions
   def self.org_pets_with_apps(staff_org_id)
-    Pet.org_pets(staff_org_id).includes(:adopter_applications).where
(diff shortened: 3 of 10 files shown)


kasugaijin:
Awesome I will review this weekend. I realised that using icons alone in the table is probably not going to pass any accessibility assessment, so we might add text in the future...something a UX expert can advise on!

app/controllers/organizations/pets_controller.rb
```diff
@@ -93,4 +94,8 @@ def set_nav_tabs
       {name: "Summary", path: pet_path(@pet)}
     ]
   end
+
+  def determine_active_tab # from 265
```
kasugaijin, line 98: If this is duplicate code in the controller we could pull this into a concern and then multiple controllers can include it to use the logic.

app/models/pet.rb
```diff
@@ -95,7 +95,7 @@ def self.org_pets(staff_org_id)
 
   # all pets under an organization with applications and no adoptions
   def self.org_pets_with_apps(staff_org_id)
```
kasugaijin, line 97: For some reason, when I test this locally, the user model is nil for AdopterAccount with id = 3. So, Kamala and Brandon both load fine in the applications list, but Bad does not. I am not sure why, but want to check if this is happening to you also before I spend more time on it. I have tried re-running seeds multiple times just to be sure.

app/models/adopter_application.rb
```diff
@@ -58,4 +58,8 @@ def self.set_status_to_withdrawn(adopter_application)
     adopter_application.status = :withdrawn
     adopter_application.save
   end
+
+  def applicant_name
```
kasugaijin, line 62: Related to my comment below in `app/models/pet.rb` this method was giving me an error when trying to load applications because the user was retruning `nil` and therefore getting undefined method for nil class. If I added the safe navigator `&` I see that Brandon and Kamala load fine, but any applications from Bad were the issue, as the user was `nil`. I wonder if you have the same issue?

app/models/pet.rb
kasugaijin: PR merged. https://github.com/rubyforgood/pet-rescue/pull/340

app/controllers/organizations/pets_controller.rb
MooseCowBear: I didn't realize this method would be reused. I'll move it.

app/controllers/organizations/pets_controller.rb
```diff
@@ -1,4 +1,6 @@
 class Organizations::PetsController < Organizations::BaseController
+  include Tabbable
```
kasugaijin, line 2: Yes I realise now this is probably the only controller that will be using those tabs specifically. So, you could move this back into the controller. I thought you pulled it from another controller, but it's the same one.

MooseCowBear: I was a little confused, but then I thought there must be something I'm not seeing. In the future, I'll definitely ask :)

kasugaijin (approved):
Looks good to me!

Issue #328, opened 2023-11-03
Update the Organization Profile UI to use Geek Theme for profile


kasugaijin:
The template renedered for `/organization_profile/edit` should use the Geeks theme for the profile: https://geeksui.codescandy.com/geeks/pages/profile-edit.html

We should keep the avatar section (although not yet possible to upload on the OrganizationProfile model it can quickly be added with the  avatarable concern)

![image](https://github.com/rubyforgood/pet-rescue/assets/95949082/9950b532-1958-4aef-8507-6e1a9d84e415)

Acceptance criteria
- organization profile edit form uses the geeks layout provided above

queen-x-vee:
may I work on this pls

jadekstewart3:
@queen-x-vee Hi! I've almost finished this issue 😅 Let me know if you find another that you would like to take a crack at!

queen-x-vee:
seen one #332 @jadekstewart3 . can I work on this?

kasugaijin:
@queen-x-vee see my comment on that ticket!

PR #341, opened 2023-11-20
Org profile theme

jadekstewart3:
# 🔗 Issue
 #328 

# ✍️ Description
In this update, GeeksUI components have been integrated to leverage the theme's aesthetic and functionality. Notably, an Avatar section has been introduced to enhance user profiles. While the visual components are in place, the Avatar section is currently non-functional, as indicated in the view. 
I have also implemented happy and sad path testing for the organization_profile edit page, along with a new, and tested email validation.

# 📷 Screenshots/Demos
Updated Screenshot:
Screenshot 2023-11-28 at 6 00 21 PM


kasugaijin:
Looks good! A few thoughts
image

We could get rid of the EDIT Profile header and make it more like the template above with a small section liek the 
```
Profile Details
You have full control to manage your own account setting.
```
And include the horizontal rules as well to separate sections. 

Could we also lay the inputs out with two columns?

app/views/organizations/organization_profiles/_form.html.erb
```diff
@@ -4,57 +4,105 @@
...
+                                { prompt: "Please select" },
+                                'data-path': country_states_path(:country),
+                                'data-country-state-target': 'country',
+                                'data-action': 'change->country-state#updateStates',
+                                'data-initial-value': location.object.country,
+                                required: true,
+                                class: 'form-control' %>
+            <% profile.location.custom_messages(:country).each do |error| %>
+              <div class="alert alert-danger mt-1" role="alert"><%= error.message %></div>
+            <% end %>
```
edwinthinks, line 78: Not 100% sure but I think this isn't needed, doesn't bootstrap form handle showing the errors?

kasugaijin (approved):
Great job @jadekstewart3 
Now hit that green button!

Issue #342, opened 2023-11-22
Fix navbar dropdowns for mobile


jmilljr24:
1. The navbar avatar dropdown does not work on mobile due to it being a hover interaction. The breakpoint set in bootstrap is extra large (≥1200px).
2. When the navbar items collapse on mobile, links to adopt(sign up) and sign in are displayed in the dropdown when they should be hidden if the user is logged in.

![Screenshot from 2023-11-22 10-01-38](https://github.com/rubyforgood/pet-rescue/assets/16829344/cf294bd4-bf8f-46fb-b88d-282edae583ef)

jmilljr24:
I have a fix ready for this but wanted to asked before submitting the PR due to previous discussions about changing how the Avatar is implemented #326

PR #343, opened 2023-11-22
Fix mobile navbar dropdowns

jmilljr24:
#342  🔗 

# ✍️ Description

1. Add dropdown data target to Avatar which enables clicking instead of hover on mobile
2. Navbar collapse menu checks if user is logged in before displaying login/adopt menu items
3. Add tests for mobile screen size login/logout

# 📷 Screenshots/Demos

Screenshot from 2023-11-22 10-17-08

Screenshot from 2023-11-22 10-00-19

### DIFF
--- a/app/components/avatar_component.rb
+++ b/app/components/avatar_component.rb
@@ -32,11 +32,11 @@ class AvatarComponent < ViewComponent::Base
   end
 
   def avatar_with_image
-    image_tag @system_arguments[:src], class: @classes, alt: @alt
+    image_tag @system_arguments[:src], class: @classes, 'data-bs-toggle' => 'dropdown', alt: @alt
   end
 
   def avatar_with_initials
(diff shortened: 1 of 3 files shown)


jmilljr24:
If you end up accepting this, please take an extra moment to review. I am very new to all of this!!

kasugaijin:
Please run `rails standard:fix`

kasugaijin:
Hey @jmilljr24 thank you for taking the initiative! Much appreciated.
So you are right in noticing that we had discussion around the use of ViewComponent, and that instead we would prefer Nice Partials. So, I won't approve this for now as I am reverting this PR https://github.com/rubyforgood/pet-rescue/pull/326 which your PR refers to. It would be great if you want to do the same changes when the new PR is up to add avatars with nice_partials?

In the meantime, feel free to pick up another issue :)

jmilljr24:
Sounds good! I'll keep an eye out and see what I can do.

PR #312, opened 2023-10-27
Staff can upload and remove images for a pet

torresga:
# 🔗 Issue

https://github.com/rubyforgood/pet-rescue/issues/259

# ✍️ Description

This feature allows staff to upload and delete images for a pet.

I had a lot of trouble getting this to work for some reason. The biggest issue I had was when the “happy path” in `update_images` failed. When the image wasn’t attached, say because the validations failed (the image was too big, there were already 5 images uploaded), the blob of that image still existed in `@pet.images`. That’s why I called `purge` in the action. Let me know if there’s a better way to handle this. 

I'd also like to add a few more assertions to the controller test I added, for example, asserting that the response redirects to the pet's show page. I had some trouble getting that working as well, I suppose writing tests is not my strong suit. 😅 

# 📷 Screenshots/Demos

Screen Shot 2023-10-27 at 2 02 29 PM

PR #345, opened 2023-11-25
Staff can upload photos and records on a pet.

kasugaijin:
# 🔗 Issue
Building on top of this stale PR: https://github.com/rubyforgood/pet-rescue/pull/312

# ✍️ Description
PR adds ability to upload pet photos and records. 

Decisions:
- use a separate tab for image and record uploads as they are two distinct resources serving different purposes
- I did go down the route of creating a separate controller to manage file uploads so that we could stick to RESTful actions in the controllers, but I had trouble using render for pet-show to allow validation errors to show on the form, which I think is the best UX. I could get errors to show in a flash using redirect_to pet-show, but it was not consistent with other forms and I did not like it. So, I decided to keep the two non-REST actions on the Pets controller to make it easier to call render on a Pet template and allow form validation errors to show.

To build on this work (in separate issues/PRs): 
- display uploaded images in a (better?) way
- enable user to download or preview an uploaded record

# 📷 Screenshots/Demos
Peek 2023-11-25 15-30


Ptrboro (approved):
Great work @kasugaijin!! 👏

Just a thought - I think `files` is a clearer name than `records`. Records could be easily confused with database entries in discussions. Just my two cents but I'm not a native speaker, though. Your call :) Awesome job!

kasugaijin:
@Ptrboro Yes you are right! Files is better naming than Records. I will make the necessary changes. Thank you.

Issue #278, opened 2023-10-17
URL helpers in mailers do not prefix the path with organization slug


Ptrboro:
E.g. `accept_invitation_url` should return `{host}/alta/invitations/accept` instead of just `{host}/invitations/accept` in mailers. It works fine in controllers.

PR #346, opened 2023-11-29
Prefix links in mailers with organization slug

Ptrboro:
# 🔗 Issue
#278 

# ✍️ Description
All urls generated in mailers will now include the organization slug (if the organization is set). This will allow us to properly redirect users after signing up / accepting invitations, etc.

# 📷 Screenshots/Demos

https://github.com/rubyforgood/pet-rescue/assets/16762860/77004961-db92-42a7-ad75-7c09610d205e


app/mailers/application_mailer.rb
```diff
@@ -1,4 +1,15 @@
...
+  around_action :with_organization_path, if: -> { Current.organization }
+
+  private
+
+  def with_organization_path
+    default_url_options[:script_name] = "/#{Current.organization.slug}"
+    yield
+  ensure
+    default_url_options[:script_name] = nil
+  end
```
Ptrboro, line 14: I'm worried about the thread safety of this code but we don't have a better solution so I guess it's better than nothing :)

kasugaijin: Yes let's go with it until we need to do something else...if that comes.

kasugaijin (approved):

PR #348, opened 2023-11-29
Fix layout on wide screens

Ptrboro:
# ✍️ Description
The dashboard layout was broken on screens wider than 768px:
Screenshot 2023-11-29 at 14 55 37

Now it should be displayed correctly on all screens :)
There are still some issues with the sidebar sometimes, we can fix it separately.

# 📷 Screenshots/Demos

https://github.com/rubyforgood/pet-rescue/assets/16762860/a58bbd5d-b6b1-4fcf-94d4-167c5aed1916

### DIFF
--- a/app/views/layouts/dashboard.html.erb
+++ b/app/views/layouts/dashboard.html.erb
@@ -8,7 +8,7 @@
       <%= render "layouts/dashboard/sidebar" %>
 
       <!-- Page Content -->
-      <main id="page-content">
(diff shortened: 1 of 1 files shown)


kasugaijin (approved):
Ah let's try this class! We tried this earlier but I think the class you specified might be the fix. https://github.com/rubyforgood/pet-rescue/pull/331

PR #364, opened 2023-12-08
362 pet tasks validations

### LINKED ISSUE #362: Add validations to Pet tasks
Pet > show > Tasks I can create a task with no attribute data. 

Acceptance criteria:
- add model validation for name and body, presence true
- add null false constraint on Name column of Tasks table

jmilljr24:
# 🔗 Issue

Resolves #362 
# ✍️ Description

Add migrations for null constraint for name in tasks. In task model add validation to name and description.

Adds tests for model validations.

# 📷 Screenshots/Demos


jmilljr24:
The validations work as expected but I think a separate issue should be added for the front end modal to either prevent the user from clicking submit without the required information or have some sort of redirect. Right now if you hit submit, nothing is added to the db but there is a "content missing" in place of the model if you try to add a task after attempting to submit one without name/description.

I believe it's something to do with turbo-stream but I'm not sure exactly how to handle that. I'm sure there is something simple to prevent the form from being submitted. 

[https://jam.dev/c/4c29e57b-dd57-47c3-aa38-21d963d03c88](url)

edwinthinks:
@jmilljr24 yep that sounds like a good time to update the code so we are catching the invalid saving and respond back with the turbo frame saying there is an issue in the form. Would you like do it in the PR or merge this one and then do it in a seperate?

test/models/task_test.rb
```diff
@@ -33,4 +33,16 @@ class TaskTest < ActiveSupport::TestCase
...
+    task = build(:task, name: nil)
+    refute task.valid?, "task is valid without a name"
+    assert_not_nil task.errors[:name], "no validation error for name present"
+  end
+
+  test "invalid without description" do
+    task = build(:task, description: nil)
+    refute task.valid?, "task is valid without description"
+    assert_not_nil task.errors[:description], "no validation error for description present"
+  end
```
edwinthinks, line 47: You can use https://github.com/thoughtbot/shoulda-matchers which makes this a bit faster  and less having to write our own tests.

edwinthinks (changes requested):

jmilljr24:
> @jmilljr24 yep that sounds like a good time to update the code so we are catching the invalid saving and respond back with the turbo frame saying there is an issue in the form. Would you like do it in the PR or merge this one and then do it in a seperate?

I'd prefer to merge this and have it be opened as a separate issue. Maybe someone else can take a stab at that. I might be able to cobble something together eventually but it might be a while.

Ptrboro (changes requested):
I think we can simplify these migrations, we only need something like this:
```
class AddNotNullToNameInTasks < ActiveRecord::Migration[7.0]
  def change
    change_column_null :tasks, :name, false
  end
end
```

Did you run into any issues and that's why you added check constraints?

jmilljr24:
> I think we can simplify these migrations, we only need something like this:
> 
> ```
> class AddNotNullToNameInTasks < ActiveRecord::Migration[7.0]
>   def change
>     change_column_null :tasks, :name, false
>   end
> end
> ```
> 
> Did you run into any issues and that's why you added check constraints?

Yes, I started with just the change_column_null. I don't remember the exact verbiage but basically I got an error saying not to do do that because if you make a change or rollback it will cause issues. There error output gave the suggestion I used. I also researched it and the check constraints prevent there from being any issues in the future.

jmilljr24:
=====
rails aborted!
StandardError: An error has occurred, this and all later migrations canceled:

=== Dangerous operation detected #strong_migrations ===

Setting NOT NULL on an existing column blocks reads and writes while every row is checked.
Instead, add a check constraint and validate it in a separate migration.

class AddNotNullToNameInTasks < ActiveRecord::Migration[7.0]
  def change
    add_check_constraint :tasks, "name IS NOT NULL", name: "tasks_name_null", validate: false
  end
end

class ValidateAddNotNullToNameInTasks < ActiveRecord::Migration[7.0]
  def up
    validate_check_constraint :tasks, name: "tasks_name_null"
    change_column_null :tasks, :name, false
    remove_check_constraint :tasks, name: "tasks_name_null"
  end

  def down
    add_check_constraint :tasks, "name IS NOT NULL", name: "tasks_name_null", validate: false
    change_column_null :tasks, :name, true
  end
end

Ptrboro:
@jmilljr24 I see, this is caused by the [strong_migrations](https://github.com/ankane/strong_migrations) gem. It makes sure we don't create any migrations that could be unsafe in a production environment. But as we don't have a production yet we don't need to worry about this.

To skip the error you have to wrap your code in `safety_assured` block like so:

```
class AddNotNullToNameInTasks < ActiveRecord::Migration[7.0]
  def change
    safety_assured do
      change_column_null :tasks, :name, false
    end
  end
end
```

Btw @edwinthinks @kasugaijin what do you think about disabling this gem until we have production env? It doesn't give us any value now and just slows things down.

kasugaijin:
> @jmilljr24 I see, this is caused by the [strong_migrations](https://github.com/ankane/strong_migrations) gem. It makes sure we don't create any migrations that could be unsafe in a production environment. But as we don't have a production yet we don't need to worry about this.
> 
> To skip the error you have to wrap your code in `safety_assured` block like so:
> 
> ```
> class AddNotNullToNameInTasks < ActiveRecord::Migration[7.0]
>   def change
>     safety_assured do
>       change_column_null :tasks, :name, false
>     end
>   end
> end
> ```
> 
> Btw @edwinthinks @kasugaijin what do you think about disabling this gem until we have production env? It doesn't give us any value now and just slows things down.

100% I don't think it helps at all right now. Only hinders.

Ptrboro (approved):
LGTM! 👏

edwinthinks (approved):

Issue #352, opened 2023-11-29
Staff can download a file by clicking on the filename in the Pet > show > Files tab table


kasugaijin:
Acceptance criteria
- Staff can download a file by clicking on the filename in the Pet > show > Files tab table

DreamVenerable:
Can I give this try?

kasugaijin:
Sure @DreamVenerable all yours

PR #367, opened 2023-12-10
Improve image table design

diegolinhares:
# 🔗 Issue
Resolves #351 

# ✍️ Description
Improved the photos table by formatting the design of the table and adding links to download the image when clicking on the image or the filename.

# 📷 Screenshots/Demos
table

PR #370, opened 2023-12-13
352 staff download file

DreamVenerable:
# 🔗 Issue
#352 

# ✍️ Description
Added download button for staff to download pet files

# 📷 Screenshots/Demos
Screenshot 2023-12-13 at 4 37 56 AM


DreamVenerable:
Alright, added that. 

I saw the tests failing, is it normal or from my part?

kasugaijin:
Ah so it's not a test but Brakeman identified a security issue using user input (params[:id]) directly in a file path. It could be exploited. The params should be sanitised.
```
Confidence: Weak
Category: File Access
Check: SendFile
Message: Parameter value used in file name
Code: send_file(ActiveStorage::Attachment.find(params[:id]).blob.service.path_for(ActiveStorage::Attachment.find(params[:id]).blob[:key]), :filename => ActiveStorage::Attachment.find(params[:id]).blob[:filename], :disposition => "attachment")
File: app/controllers/attachments_controller.rb
Line: 1
```

This PR https://github.com/rubyforgood/pet-rescue/pull/367 recently merged enables image downloads using a helper method with no need for additional controller logic. Try that approach. I think its cleaner and less code is better if it works. If it works you can delete the controller stuff you added.

DreamVenerable:
Yup, what's way cleaner. I'll update it now

kasugaijin (approved):
LGTM

PR #374, opened 2023-12-17
Org can set default pet tasks

kasugaijin:
# 🔗 Issue
N/A

# ✍️ Description
This PR is the preliminary work to allow org staff to create default Pet tasks. This means, for any task set in Default Pet Tasks, when a new pet is created, these tasks will also be created for that pet.

A service object is used to create tasks for each newly created Pet.

To do in future issues:
- build out the CRUD of Default Pet Tasks via the UI using Turbo Stream
- Add recurring task functionality

- Possible future requirement - scope default tasks to pet type, e.g., dogs get a different set to cats.

Peek 2023-12-16 18-09

# 📷 Screenshots/Demos


app/models/pet_task.rb
```diff
@@ -0,0 +1,22 @@
+# A plain object that creates the default pet tasks for newly created Pet
+class PetTask
```
edwinthinks, line 2: This is where I would put this as a service instead of a model. This is the type of class that is meant to do one type action. So like a interactor or service object imo

app/models/default_pet_task.rb
```diff
@@ -0,0 +1,17 @@
+# == Schema Information
+#
+# Table name: default_pet_tasks
+#
+#  id              :bigint           not null, primary key
```
edwinthinks, line 5: Do we want to enforce order on these? If so, https://github.com/brendon/acts_as_list could be a nice addition

edwinthinks (commented):
Good start! Just left some thoughts on what we may want to add later.

app/models/pet_task.rb
kasugaijin: Yeh I am fine as a Service also. It's not what I would call domain logic.

app/models/default_pet_task.rb
kasugaijin: Oh yes good catch I forgot about that requirement.

app/services/organizations/default_pet_task_service.rb
```diff
@@ -0,0 +1,22 @@
+# Service object that creates the default pet tasks for a newly created Pet
+class Organizations::DefaultPetTaskService
+  def initialize(pet)
+    @pet = pet
+    @default_pet_tasks = DefaultPetTask.all
+  end
+
+  def create_tasks
```
gorangorun, line 8: You can replace this with a single/bulk query for a faster insert:

default_tasks = @default_pet_tasks.map do |task|
  {
    pet_id: @pet.id
    name: task.name,
    description: task.description,
    completed: false
   }
end

Task.create(default_tasks)

db/migrate/20231217000021_create_default_pet_tasks.rb
```diff
@@ -0,0 +1,11 @@
+class CreateDefaultPetTasks < ActiveRecord::Migration[7.0]
+  def change
+    create_table :default_pet_tasks do |t|
+      t.bigint :organization_id, null: false
```
gorangorun, line 4: Missing index. Either create a reference or add a index

t.references :organization

or

add_index :default_pet_tasks, :organization_id

kasugaijin: Yes good call. Index is going to beneficial here.

app/services/organizations/default_pet_task_service.rb
kasugaijin: Thank you for that. I totally forgot about passing an array of hashes to `create`. 
From what I have read it is not a single query, however, and I don't think is any more performant than what I had. It still treats each create as a separate insert statement. Though it does read a little cleaner so I will use it.

kasugaijin: There is the `insert_all` method that does bulk insert in a single query: https://edgeapi.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-insert_all-21
No objects are created and validations are skipped. Given the number of default tasks is probably unlikely to exceed, say, 20, I think using `create` will suffice. Though, open to other views on that.

app/models/default_pet_task.rb
kasugaijin: @edwinthinks so is the use of acts_as_list primarily for implement drag and drop ordering of tasks? Are we thinking drag and drop is MVP?

If we just want to enforce order, are you thinking on the default task we would have a `position` attribute that is used to set the same attribute on the Task for ordering tasks in the UI? That makes sense. But, what about for newly created tasks (not default)? We would have to set the `position` on the back end e.g., get the last Task's `position` and increment it by one. But then we'd have to handle simultaneous writes and race conditions.

I am thinking about this task list for MVP...in the simplest way possible. Feel free to shoot this down...

Why don't we just list uncompleted tasks by the created_at date? Then all default tasks will be above newly created tasks. Then list completed tasks beneath all of the uncompleted tasks. It's a simple way to keep the uncompleted stuff at the top. Recurring tasks will be added to this list, also. The way I see it, over time, the older default tasks should be completed, so will move to the bottom, and newly created, uncompleted tasks will remain at the top. This will work well for recurring tasks also.

For due dates, that's easily handled in MVP with a due date attribute, we display due date next to task name, and we just check if today is past this date, and if so, display a red overdue or the likes.

kasugaijin:
@all-contributors
please add @gorangorun for code. 
please add @phonghpham for code.
please add @DreamVenerable for code.
please add @diegolinhares for code.
please add @egemen-dev for code.
please add @DreamVenerable for code.
please add @MooseCowBear for code.

kasugaijin:
https://github.com/all-contributors
please add @DreamVenerable for code.
please add @diegolinhares for code.
please add @egemen-dev for code.
please add @DreamVenerable for code.
please add @MooseCowBear for code.

kasugaijin:
@all-contributors
please add @diegolinhares for code.
please add @egemen-dev for code.
please add @DreamVenerable for code.
please add @MooseCowBear for code.

kasugaijin:
@all-contributors
please add @egemen-dev for code.
please add @DreamVenerable for code.
please add @MooseCowBear for code.

kasugaijin:
@all-contributors
please add @DreamVenerable for code.
please add @MooseCowBear for code.

app/models/default_pet_task.rb
edwinthinks: > We would have to set the position on the back end e.g., get the last Task's position and increment it by one. But then we'd have to handle simultaneous writes and race conditions.

Good point. Though I don't think it is much of a case if we aren't dealing with high concurrency. There is a another gem for that which is similiar that handles that.

> I am thinking about this task list for MVP...in the simplest way possible. Feel free to shoot this down...

Thanks for sharing your thoughts! I agree let's keep the MVP simple. No need to add this if its going to add extra burden to releasing.

app/models/pet.rb
```diff
@@ -30,6 +30,8 @@
 class Pet < ApplicationRecord
   acts_as_tenant(:organization)
 
+  after_create { Organizations::DefaultPetTaskService.new(self).create_tasks }
+
```
Ptrboro, line 34: What do you think about explicitly calling this Service in the `PetsController#create` action rather than using a callback?

Callbacks can quickly become unwieldy and have some drawbacks. E.g. in tests, this callback will execute every time a Pet is created, even if the test doesn't require Tasks, leading to unnecessary database queries. This can considerably slow down the test suite as the number of tests grows.

app/services/organizations/default_pet_task_service.rb
```diff
@@ -0,0 +1,26 @@
...
+      }
+    end
+  end
+
+  def create_tasks
+    ActiveRecord::Base.transaction do
+      Task.create(default_tasks_array)
+    rescue ActiveRecord::RecordInvalid => e
+      Rails.logger.info "Error creating tasks: #{e.message}"
+    end
```
Ptrboro, line 24: I think the error handling here is a bit problematic. Let's untangle it :)

The first problem is that the rescue block is placed inside a transaction - if an exception is rescued inside a transaction block, the transaction will not automatically roll back. This means that if there is an error when creating e.g. third task then the other two will still be persisted in the DB. 

The other thing is that `Task.create` is missing a bang - now it doesn't raise an exception when validation fails. It simply saves only valid records and skips invalid ones. If we want all-or-nothing behavior, we need to use `Task.create!`.

Considering these points I think what we need here is something like:

```
def create_tasks
    Task.create!(default_tasks_array)
  end

rescue ActiveRecord::RecordInvalid => e
  Rails.logger.info "Error creating tasks: #{e.message}"
end
```

app/models/pet.rb
kasugaijin: @Ptrboro calling on the controller create action is how I started writing this. I then decided to move it to the model as I thought it is something we would want to run every time. and living on the model it would save duplication in future should we ever create a pet in a different way. However, I agree that callbacks can become a mess, and for now calling the service on the create action is fine and probably a better approach. I will make this change.

app/services/organizations/default_pet_task_service.rb
kasugaijin: I am very happy you caught this! I will admit I have not used transaction blocks much. I will make this change.

app/controllers/organizations/pets_controller.rb
```diff
@@ -29,11 +29,21 @@ def show
   end
 
   def create
```
kasugaijin, line 31: I have not used a transaction on a controller ation before, but it made sense to me to create the pet and default pet tasks in one hit, or none at all. Curious what your thoughts are on doing this.

test/models/default_pet_task_test.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  test "can update name" do
+    default_pet_task = create(:default_pet_task, description: "Old Name")
+
+    assert_equal "Old Name", default_pet_task.description
+
+    default_pet_task.update(description: "New Name")
+    assert_equal "New Name", default_pet_task.description
+  end
```
edwinthinks, line 34:
```suggestion
```
You probably don't need these tests since I think you are testing Rails and this isn't needed. We ought to trust that Rails can do these things :)

test/models/default_pet_task_test.rb
```diff
@@ -0,0 +1,35 @@
...
+  test "should have valid factory" do
+    assert build(:default_pet_task).valid?
+  end
+
+  test "should belong to an organization" do
+    default_pet_task = build(:default_pet_task)
+    assert_difference("default_pet_task.organization.default_pet_tasks.count", 1) do
+      default_pet_task.save
+    end
+  end
```
edwinthinks, line 16:
```suggestion
  should belong_to(:organization)
```
You can double check shoulda but there is a matcher for this :)

test/controllers/organizations/pets_controller_test.rb
```diff
@@ -57,4 +58,33 @@ class Organizations::PetsControllerTest < ActionDispatch::IntegrationTest
     assert_equal Mime[:html], response.media_type
     assert_response :redirect
   end
+
+  test "POST default pet tasks are created when pet is created" do
+    post pets_path, params: post_params
```
edwinthinks, line 63:
```suggestion
    assert_difference "Pet.count", 1 do
       post pets_path, params: post_params
    end
```
This is going to better ensure that we are certainly creating a Pet and then you can called `created_pet = Pet.last` and do your assertions.

test/controllers/organizations/pets_controller_test.rb
```diff
@@ -57,4 +58,33 @@ class Organizations::PetsControllerTest < ActionDispatch::IntegrationTest
...
+
+    assert_equal Pet.last.tasks.count, 1
+  end
+
+  private
+
+  def post_params
```
edwinthinks, line 70: This is used in one spot so I think you can put in the `test` block above

app/controllers/organizations/pets_controller.rb
edwinthinks: I think a transaction is a good idea to avoid those weird invalid states.

app/controllers/organizations/pets_controller.rb
```diff
@@ -29,11 +29,21 @@ def show
...
+    transaction_success = false
     @pet = Pet.new(pet_params)
 
-    if @pet.save
+    ActiveRecord::Base.transaction do
+      if @pet.save && Organizations::DefaultPetTaskService.new(@pet).create_tasks
```
edwinthinks, line 36: Instead of doing this `transaction_success` management you can do raise within the block and rescue that then call `ActiveRecord::Rollback`

```
   begin
      @pet.save! 
      Organizations::DefaultPetTaskService.new(@pet).create_tasks!
   rescue
      raise ActiveRecord::Rollback
   end
end
```
Any raised error would trigger the rollback.

app/controllers/organizations/pets_controller.rb
```diff
@@ -29,11 +29,21 @@ def show
...
+        transaction_success = true
+      else
+        raise ActiveRecord::Rollback
+      end
+    end
+
+    if transaction_success
```
edwinthinks, line 43: If you use the approach I suggest above you can do `@pet.persisted?` instead to know if the transaction worked.

edwinthinks (commented):
@kasugaijin I left some feedback on some minor improvement suggestions. I don't want to block this much longer but didn't want to lose the chance to share some of my thoughts.

test/models/default_pet_task_test.rb
kasugaijin: Yes makes sense.

test/models/default_pet_task_test.rb
kasugaijin: So turns out @edwinthinks 
```
context "associations" do
    should belong_to(:organization)
  end
```
works if I use `  belongs_to :organization` on the `DefaultPetTask` model.
But if I use `acts_as_tenant(:organization)` on the model that shoulda block fails. 

This is where I wonder which is more appropriate to use as association. The OrganizationProfile uses `belongs_to` whereas Pet uses `acts_as_tenant`. Is there a 'better' approach? I see for models that use `acts_as_tenant` they do not assert the belongs_to relationship with shoulda.

test/controllers/organizations/pets_controller_test.rb
kasugaijin: Done

test/controllers/organizations/pets_controller_test.rb
kasugaijin: I thought it might be useful for future tests. But moving params to a private method can be done in future if necessary.

app/services/organizations/default_pet_task_service.rb
Ptrboro: I'm glad I can help! I hope to have more time to contribute in January :)

test/models/default_pet_task_test.rb
edwinthinks: Anything that needs to be scoped to the organization tenant should use `acts_as_tenant`.

app/controllers/organizations/pets_controller.rb
```diff
@@ -31,9 +31,17 @@ def show
   def create
     @pet = Pet.new(pet_params)
 
-    if @pet.save
+    ActiveRecord::Base.transaction do
+      @pet.save!
+      Organizations::DefaultPetTaskService.new(@pet).create_tasks
```
edwinthinks, line 36: To make this work you'll need to make sure that `create_task` raises an error that is uncaught. I think the class we have it caught so doesn't bubble up here.

app/services/organizations/default_pet_task_service.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def create_tasks
+    ActiveRecord::Base.transaction do
+      Task.create!(default_tasks_array)
+    end
+  rescue ActiveRecord::RecordInvalid => e
+    Rails.logger.info "Error creating tasks: #{e.message}"
+    false
+  end
+  true
```
edwinthinks, line 27:
```suggestion
```
Not sure what this `true` is for.

app/services/organizations/default_pet_task_service.rb
```diff
@@ -0,0 +1,28 @@
...
+        description: task.description,
+        completed: false
+      }
+    end
+  end
+
+  def create_tasks
+    ActiveRecord::Base.transaction do
+      Task.create!(default_tasks_array)
+    end
```
edwinthinks, line 22: Mentioned above you'll want to let the raised error pass through to the top so it triggers the Rollback

edwinthinks (approved):
One minor bit, after that LGTM

app/controllers/organizations/pets_controller.rb
kasugaijin: Ahh yes right I see now.

app/services/organizations/default_pet_task_service.rb
kasugaijin: Yes that can go.

app/services/organizations/default_pet_task_service.rb
kasugaijin: got it so I have remove the rescue block and now have:

```
 def create_tasks
      Task.create!(default_tasks_array)
    end
  end
```

PR #384, opened 2023-12-27
Update task ux

Thrillberg:
# 🔗 Issue

# ✍️ Description

# 📷 Screenshots/Demos


app/views/organizations/tasks/_form.html.erb
```diff
@@ -1,9 +1,9 @@
-<%= form_for [@pet, @task], remote: true do |f| %>
```
edwinthinks, line 1: Somewhat of a unrelated thing I noticed is that maybe the `tasks` folder should be under `pets`. I think we got confused (or I did) when I didn't realize that we have a `pets/tabs` section instead.

app/views/organizations/tasks/edit.html.erb
```diff
@@ -1,26 +1,5 @@
...
-
-    <% end %>
-  </turbo-frame>
-</div>
\ No newline at end of file
+<%= turbo_frame_tag @task do %>
+  <%= bootstrap_form_with model: @task, url: pet_task_path(@task.pet) do |f| %>
+    <%= render partial: 'form', locals: {f: f} %>
+  <% end %>
+<% end %>
```
edwinthinks, line 5: Nice!

edwinthinks (approved):
Minor thoughts! And we can get this merged in imo!

P.S - I think the next step is to add the checkbox to mark tasks as done or not done. Could you create a issue for that?

app/views/organizations/tasks/_form.html.erb
Thrillberg: I've moved some templates in the latest commit but I'm not sure if it's an improvement? Things are still kind of split up in multiple places. What do you think?

Thrillberg:
Issue created here: https://github.com/rubyforgood/pet-rescue/issues/386

Thrillberg:
I've removed the failing test, which is not a great solution. But that test seems mostly meaningless to me, at this point, because on validation error the controller doesn't actually redirect or fail or anything bad--it just gives a 200OK. Is that the desired pattern, @edwinthinks ?

edwinthinks:
@Thrillberg I think that is a fine solution for now. I think adding more meaningful tests will come later.

Issue #388, opened 2023-12-28
Spike on Form builder architecture


Thrillberg:
![image](https://github.com/rubyforgood/pet-rescue/assets/10481391/8fa84f7a-d6aa-4f81-a31e-5fa4a78df327)

Notes from Mae:
>Another note, it knows which kind of Response input to display bc Question has both name and input_type

>My Response has fields for each kind of input type to simplify, so, fields are:
string_value
integer_value
file_value
boolean_value
array_value

>The Checklist/Tasklist builder was built the same way so as to store the List (Form) separate from the Assignments (Responses).

Ptrboro:
Hey, I wasn't part of the Spike discussion, so I might be missing some context, but I'm concerned that updating or deleting a question could mess up the historical response data. E.g. if a question is deleted, all associated responses would also be removed. I think keeping historical data is crucial, especially since it may contain important information like adopter contact details, and losing such data could be problematic.

But I think there's a straightforward solution. We can add question_text and input_type fields to the Response model and when creating a response, we should copy this information from the Question. This way, even if the question is later updated or deleted, we'll still have a record of the original question text at the time the response was submitted. It might also be necessary to modify or remove the foreign_key constraint between the Response and the Question models (so we can delete Question, without losing Responses).

I hope this is helpful :) We can also add this later if needed and keep the solution simple for now.

Thrillberg:
Thanks for raising the valid concern @Ptrboro ! I wonder if @edwinthinks 's idea of soft deleting would be sufficient? If we do that then we wouldn't have to add redundant data elsewhere. We could also use [the gem `papertrail`](https://github.com/paper-trail-gem/paper_trail) to keep track of versions of Questions (not sure if this addresses the need around Question deletion though). What do you all think?

For now, it's probably best to make Questions unable to be updated or deleted just for simplicity but I think we'd want to support updates/deletion in the future.

Ptrboro:
@Thrillberg @edwinthinks soft-deleting will fix the problem for deleted questions but not for updated ones 🤔 It will be common for people to change the text of a question instead of deleting and creating a new one. Non-tech people won't be aware that responses are linked to questions and changing the question text will change how the old submissions look like. Papertrail is worth researching but it's a big framework and may have its own set of challenges.

But I think we all agree to start with a simple and solid foundation and add this logic later :)

maebeale:
[...]

PR #389, opened 2023-12-28
Custom Form builder

Thrillberg:
# 🔗 Issue
#388 

# ✍️ Description
Lots of new tables and models.

### DIFF
--- a/app/controllers/organizations/forms_controller.rb
+++ b/app/controllers/organizations/forms_controller.rb
@@ -1,8 +1,8 @@
 class Organizations::FormsController < Organizations::BaseController
   before_action :verified_staff
-  
+
   layout "dashboard"
-    
+
   def edit
     @form = Current.organization.form || Form.create(
       organization: Current.organization,
@@ -10,4 +10,3 @@ class Organizations::FormsController < Organizations::BaseController
     )
   end
 end
-  
--- a/app/controllers/organizations/forms_controller.rb
+++ b/app/controllers/organizations/forms_controller.rb
@@ -1,0 +1,13 @@
+class Organizations::FormsController < Organizations::BaseController
+  before_action :verified_staff
+  
+  layout "dashboard"
+    
+  def edit
+    @form = Current.organization.form || Form.create(
+      organization: Current.organization,
+      name: "#{Current.organization.name}'s Application Form"
+    )
+  end
+end
+  
--- a/app/controllers/organizations/questions_controller.rb
+++ b/app/controllers/organizations/questions_controller.rb
@@ -4,6 +4,7 @@ class Organizations::QuestionsController < Organizations::BaseController
   def new
     @organization = current_user.organization
     @form = @organization.form || Form.create(organization: @organization, name: "#{@organization.name}'s Application Form")
+    @question = @form.questions.build
   end
 
   def create
@@ -15,16 +16,12 @@ class Organizations::QuestionsController < Organizations::BaseController
       respond_to do |format|
         format.turbo_stream { render turbo_stream: turbo_stream.replace("questions_list", partial: "organizations/forms/form", locals: {question: @question}) }
       end
-    else
-      # respond_to do |format|
-      #   format.turbo_stream { render turbo_stream: turbo_stream.replace(@task, partial: "organizations/pets/tasks/form", locals: {task: @task, url: pet_tasks_path(@task.pet)}) }
-      # end
     end
   end
 
   private
 
   def question_params
-    params.permit(:text, :input_type)
+    params.require(:question).permit(:text, :input_type)
   end
 end
--- a/app/controllers/organizations/questions_controller.rb
+++ b/app/controllers/organizations/questions_controller.rb
@@ -1,0 +1,30 @@
+class Organizations::QuestionsController < Organizations::BaseController
+  before_action :verified_staff
+    
+  def new
+    @organization = current_user.organization
+    @form = @organization.form || Form.create(organization: @organization, name: "#{@organization.name}'s Application Form")
+  end
+
+  def create
+    @organization = current_user.organization
+    @form = @organization.form
+    @question = Question.new(question_params)
+
+    if @question.save && FormQuestion.create(question: @question, form: @form)
+      respond_to do |format|
+        format.turbo_stream { render turbo_stream: turbo_stream.replace("questions_list", partial: "organizations/forms/form", locals: {question: @question}) }
+      end
+    else
+      # respond_to do |format|
+      #   format.turbo_stream { render turbo_stream: turbo_stream.replace(@task, partial: "organizations/pets/tasks/form", locals: {task: @task, url: pet_tasks_path(@task.pet)}) }
+      # end
+    end
+  end
+
+  private
+
+  def question_params
+    params.permit(:text, :input_type)
+  end
+end
--- a/app/controllers/organizations/questions_controller.rb
+++ b/app/controllers/organizations/questions_controller.rb
@@ -1,6 +1,6 @@
 class Organizations::QuestionsController < Organizations::BaseController
   before_action :verified_staff
-    
+
   def new
     @organization = current_user.organization
     @form = @organization.form || Form.create(organization: @organization, name: "#{@organization.name}'s Application Form")
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -47,7 +47,7 @@ class RegistrationsController < Devise::RegistrationsController
   end
 
   def after_sign_up_path_for(resource)
-    adoptable_pets_path
+    new_submission_path
   end
 
   def after_sign_in_path_for
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -19,9 +19,7 @@ class SubmissionsController < ApplicationController
 
     ActiveRecord::Base.transaction do
       responses.each do |response|
-        unless response.save
-          debugger
-        end
+        response.save
       end
     end
 
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -1,0 +1,36 @@
+class SubmissionsController < ApplicationController
+  before_action :authenticate_user!
+
+  def new
+    @form = Current.organization.form
+    @submission = Submission.new(form: @form)
+  end
+
+  def create
+    @submission = Submission.new(form: Current.organization.form, user: current_user)
+    responses = []
+    submission_params.each do |question_id, response|
+      responses << Response.new(
+        question_id: question_id,
+        string_value: response,
+        submission: @submission
+      )
+    end
+
+    ActiveRecord::Base.transaction do
+      responses.each do |response|
+        unless response.save
+          debugger
+        end
+      end
+    end
+
+    redirect_to adoptable_pets_path, notice: "Application form successfully submitted."
+  end
+
+  private
+
+  def submission_params
+    params.require(:submission)
+  end
+end
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -1,0 +1,14 @@
+# == Schema Information
+#
+# Table name: forms
+#
+#  id          :bigint           not null, primary key
+#  description :text
+#  name        :string
+#  created_at  :datetime         not null
+#  updated_at  :datetime         not null
+#
+class Form < ApplicationRecord
+  has_many :form_questions, dependent: :destroy
+  has_many :submissions, dependent: :destroy
+end
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -2,13 +2,21 @@
 #
 # Table name: forms
 #
-#  id          :bigint           not null, primary key
-#  description :text
-#  name        :string
-#  created_at  :datetime         not null
-#  updated_at  :datetime         not null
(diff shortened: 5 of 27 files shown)


kasugaijin (commented):
We have some existing overlap wtith the ChecklistX tables but I do prefer the Form naming over Checklists as it seems more appropriate for this use case, a custom form. So I'd say we keep these table names in your PR and get rid of the existing Checklist tables (in a separate issue/PR if you prefer). 

`checklist_assignments`
`checklist_template_items`
`checklist_templates`

Thrillberg:
Are Forms and Checklists intended to be the same thing? I thought they were two different things.

kasugaijin:
@Thrillberg my understanding was that the checklist tables were also going to be used for the Tasks functionality as well as the form builder. But we went with a tasks table instead to keep it simple. We can check in with the team in slack about that (I am not sure what was discussed at the conference re these checklist tables exactly).

db/migrate/20231228010428_create_responses.rb
```diff
@@ -0,0 +1,14 @@
+class CreateResponses < ActiveRecord::Migration[7.0]
+  def change
+    create_table :responses do |t|
+      t.references :submission, null: false, foreign_key: true
+      t.references :question, null: false, foreign_key: true
+      t.string :string_value
+      t.integer :integer_value
+      t.boolean :boolean_value
+      t.text :array_value, array: true
```
edwinthinks, line 9: Hmm I'am wondering if we want to store the value as a string and just serialize them? Maybe too much extra effort and it is best to just have these split up like this.

app/models/response.rb
```diff
@@ -0,0 +1,28 @@
+# == Schema Information
+#
+# Table name: responses
+#
+#  id            :bigint           not null, primary key
+#  array_value   :text             is an Array
```
nsiwnf, line 6: Whats an example of an array response?

nsiwnf: I could also see "date" being a response field - not sure if thats already captured here

db/migrate/20231228010428_create_responses.rb
nsiwnf: Yeah - I think that would look a little cleaner, we'd have to transform it anyways, right? So store a value "true" and value_type "boolean"?

Thrillberg: So is the verdict that this is ok as-is? Part of me also thinks folding this into the `string_value` is fine but I pretty much hewed as closely as I could to Mae's design.

app/models/response.rb
Thrillberg: An array would be multiple choices selected for one question, perhaps?

Date isn't captured per se but I agree. I'll try to add it in.

app/controllers/organizations/forms_controller.rb
```diff
@@ -0,0 +1,13 @@
+class Organizations::FormsController < Organizations::BaseController
+  before_action :verified_staff
+  
+  layout "dashboard"
+    
+  def edit
+    @form = Current.organization.form || Form.create(
```
Thrillberg, line 7: Conditionally creating the form seems suboptimal. Are we better off with a seed or something that we _know_ will run? We only want one form per organization.

app/controllers/organizations/questions_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    @question = Question.new(question_params)
+
+    if @question.save && FormQuestion.create(question: @question, form: @form)
+      respond_to do |format|
+        format.turbo_stream { render turbo_stream: turbo_stream.replace("questions_list", partial: "organizations/forms/form", locals: {question: @question}) }
+      end
+    else
+      # respond_to do |format|
+      #   format.turbo_stream { render turbo_stream: turbo_stream.replace(@task, partial: "organizations/pets/tasks/form", locals: {task: @task, url: pet_tasks_path(@task.pet)}) }
+      # end
```
Thrillberg, line 21: Oops I need to handle this part.

app/controllers/registrations_controller.rb
```diff
@@ -47,7 +47,7 @@ def account_update_params
   end
 
   def after_sign_up_path_for(resource)
-    adoptable_pets_path
+    new_submission_path
```
Thrillberg, line 50: This means that right after supplying email and password etc, the user will be shown the organization's form.

app/controllers/submissions_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+        question_id: question_id,
+        string_value: response,
+        submission: @submission
+      )
+    end
+
+    ActiveRecord::Base.transaction do
+      responses.each do |response|
+        unless response.save
+          debugger
```
Thrillberg, line 23: Need to remove this debugger.

app/models/form.rb
```diff
@@ -0,0 +1,22 @@
...
+#  updated_at      :datetime         not null
+#  organization_id :bigint
+#
+# Indexes
+#
+#  index_forms_on_organization_id  (organization_id)
+#
+class Form < ApplicationRecord
+  has_many :form_questions, dependent: :destroy
+  has_many :questions, through: :form_questions
```
Thrillberg, line 18: I think this can be removed. Not actually using this functionality (yet).

app/views/layouts/dashboard/_sidebar.html.erb
```diff
@@ -37,6 +37,11 @@
       </li>
 
       <% if current_user.staff_account && current_user.staff_account.has_role?(:admin, current_user.staff_account.organization) %>
+        <li class="nav-item">
+          <%= active_link_to edit_form_path, class: "nav-link" do %>
+            <i class="nav-icon fe fe-file-text me-2"></i> Application Form
```
Thrillberg, line 42: Is there a different icon we should use here?

app/views/submissions/_form.html.erb
```diff
@@ -0,0 +1,19 @@
...
+      <label for="response_<%= form_question.question.id %>">
+        <%= form_question.question.text %>
+      </label>
+      <input
+        type="text"
+        name="submission[<%= form_question.question.id %>]"
+        id="response_<%= form_question.question.id %>"
+        class="form-control"
+      >
+    </div>
```
Thrillberg, line 13: I struggled a lot with this form. I ended up abandoning Rails in favor of just interpolating with HTML to get the params in the shape I wanted. Does anyone know a more Rails-y way to do this?

config/routes.rb
```diff
@@ -7,13 +7,17 @@
...
   resources :adoptable_pets, only: [:index, :show]
   resource :adopter_profile, except: :destroy, as: "profile"
   resources :checklist_templates
   resources :donations, only: [:create]
 
   scope module: :organizations do
     resource :organization_profile, only: %i[edit update]
+    resource :form, only: [:edit, :update]
+    resources :questions, only: [:new, :create]
+    resources :form_questions, only: [:new, :create]
```
Thrillberg, line 20: Don't think we need this part.

Thrillberg (commented):
Ok I think I've done as much as I want to do for this PR. The rest is cleanup/tweaks. It works, minimally, but there are lots of things to improve about it. I have some notes-to-self in this review, and some questions for others.

app/views/layouts/dashboard/_sidebar.html.erb
nsiwnf: I think this one is fine - we could also use `file-plus`? I think we could also change the adopter/foster applications to `inbox`

https://feathericons.com/

db/migrate/20231228010428_create_responses.rb
nsiwnf: Per mae - it sounds like the static types are good for calculations/sorting, which sounds handy so I think this is ok as-is

Thrillberg:
Here are two videos:

As staff: https://www.loom.com/share/6a2940d9c2054811a154ea3a9825a3fb
As user: https://www.loom.com/share/f29563f88eb5479387a8a4e283a1b722

app/controllers/registrations_controller.rb
kasugaijin: Maybe something to consider later on, but if the user has not yet completed and saved the form, the after sign in path should take them to the form. Otherwise, if form is completed and saved, after sign in path could be the adopter or foster dashboard.

kasugaijin:
@all-contributors
please add @Thrillberg for code. 
please add @nsiwnf for code.

app/models/form.rb
```diff
@@ -0,0 +1,22 @@
...
+  has_many :submissions, dependent: :destroy
+
+  belongs_to :organization
```
kasugaijin, line 21: I think we want to use `acts_as_tenant` for all resources that belong to an organization so they are using the gem and scope appropriately. I believe then we don't have to use `current.organization...` and can just call Resource.first or whatever we need.

kasugaijin, line 21: Quick question - per Mae's sketch `Form` belongs to `FormQuestion`, which would belong to `Organization`. Am I missing something?

app/models/question.rb
```diff
@@ -0,0 +1,20 @@
...
+# Table name: questions
+#
+#  id         :bigint           not null, primary key
+#  input_type :integer          default("string")
+#  text       :string
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+class Question < ApplicationRecord
+  enum input_type: [:string, :boolean, :integer, :array]
```
kasugaijin, line 12: I think `Question` should be scoped to organization? Similar to my `Form` comment, this would be using `acts_as_tenant` to scope a `Question` to an organization.

app/views/layouts/dashboard/_sidebar.html.erb
kasugaijin: In case we don't settle on it we can use anything for now and make an issue later for icons once we've built the MVP out.

app/models/form.rb
Thrillberg: Ah ok. I was under the impression that `acts_as_tenant` was only for when we use multiple domains or sub-domains. But we're not, right? I guess if it adds convenience with scoping then it makes sense to keep in. I'll update this line.

app/models/question.rb
Thrillberg: Ok in doing this it occurred to me that it makes sense to let all of the new models in this PR be scoped under `organization` so I put a foreign key to `organization` on all of them. Does that sound reasonable to you?

app/models/form.rb
kasugaijin: yes I asked @edwinthinks about this and his recommendation was all belongs_to org associations should use acts_as_tenant

app/models/question.rb
kasugaijin: Do we need to scope them all to an org? I think it makes sense for Form and Question so we can easily access these. The rest I think we would be retrieving in the context of something else?

FormQuestion would be form.form_questions.where(form_id: x) which would be used to display an index of questions for a given form
Response would be user.submission.responses.where(form_id: x) - or I think we could say a User has responses through submissions to call user.responses.where(form_id: x), which would be used to display a user's responses to a given form

Maybe I am not thinking far enough out, but do we need to be able to retrieve all FormQuestion or all Responses or Submissions in the context of an Org? If so, then I think the belongs_to makes sense? Otherwise I don't think we need to scope them all. What do you think?

Thrillberg: I can go either way on this. In the past when I've done multi-tenant apps (which this _kinda_ is but kinda isn't) we were always very careful to scope everything to the tenant so that we don't accidentally leak data between tenants. But I think the stakes are lower here and the possibility of leakage seems lower too.

I suppose if we ever began a query with Response or FormQuestion then it'd pose a risk but that seems like enough of an edge case I'm happy to keep those unscoped. Should I go ahead and unscope those models off organization?

kasugaijin: Yes I do see your point, though. The scale we are at anticipating (not huge) and the nature of the development process (open source) I think that the chance of data leak is greater than the negative performance effects. I.e., at some stage a query with Response or FormQuestions might be allowed into production, causing a bit of mayhem. So, I agree with your approach to scope it. Thanks for the clarity on that!

app/controllers/organizations/forms_controller.rb
edwinthinks: Yeh if we are able to add a callback to auto create the Form on organization create would be good imo.

app/controllers/organizations/questions_controller.rb
```diff
@@ -0,0 +1,27 @@
+class Organizations::QuestionsController < Organizations::BaseController
+  before_action :verified_staff
+
+  def new
+    @organization = current_user.organization
+    @form = @organization.form || Form.create(organization: @organization, name: "#{@organization.name}'s Application Form")
```
edwinthinks, line 6:
```suggestion
    @form = @organization.form || Form.create(name: "#{@organization.name}'s Application Form")
```
This would be enough without specifying the organization since ActsAsTenant should auto set that.

app/views/submissions/_form.html.erb
edwinthinks: There is certainly a way -- but if this works for now. I think we can refactor.

kasugaijin:
Closed as we have merged the baseline implementation of this work in this PR https://github.com/rubyforgood/pet-rescue/pull/629

PR #401, opened 2023-12-30
332 Create Avatar Partial

### LINKED ISSUE #332: Add avatar partial so we can use it where we need an avatar
~~When a user visits `/staff` they see an avatar that is the user's uploaded image or name initials if no image has been uploaded.~~

~~See this PR for implementation: https://github.com/rubyforgood/pet-rescue/pull/326~~

This PR: https://github.com/rubyforgood/pet-rescue/pull/326 implemented a reusable view for avatars but it used view component (which we decided is not a library we want to incorporate at this time) so was reverted. We need the same outcome but using nice_partials instead. 

Acceptance criteria
- use nice partials to create avatar partial that can be used anywhere we need it on the site
- If no uploaded avatar image, use user's name, or organization, initials, otherwise uses uploaded image

phonghpham:
# 🔗 Issue
Resolves #332 

# ✍️ Description
Created a new view partial, _avatar.html.erb, using nice_partials. This partial handles both scenarios where a user has an attached avatar image and where a default set of initials is used instead.

# 📷 Screenshots/Demos

Alta_Pet_Rescue-avatar

### DIFF
--- a/app/views/layouts/dashboard.html.erb
+++ b/app/views/layouts/dashboard.html.erb
@@ -22,30 +22,12 @@
                 <!-- List -->
                 <li class="dropdown ms-2">
                   <a class="rounded-circle" href="#" role="button" id="dropdownUser" data-bs-toggle="dropdown" aria-expanded="false">
-                    <%= render "components/avatar" do |partial| %>
-                      <% partial.alt "#{current_user.first_name.first.upcase} Avatar" %>
-                      <% partial.initials "#{current_user.first_name.first}#{current_user.last_name.first}".upcase %>
-
-                      <% if current_user.avatar.attached? %>
-                        <% partial.avatar do %>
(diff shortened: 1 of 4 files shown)


app/views/layouts/dashboard.html.erb
kasugaijin, line 1: Each time we use the partial here we have to input 10 lines of code. I think this might be a good case for creating an avatar module in `app/helpers` and moving the logic there into a method e.g. `render_avatar_partial`
so all we need to call in the view file is one line, something like `<%= render_avatar_partial(current_user) %>`. I **think** the helper has access to `current_user` but if not you just need to pass that as well in the method call. 

Do you think that'll be better?

phonghpham: Certainly! Pushed up a refactor.

kasugaijin (approved):

PR #406, opened 2024-01-02
Config for staging db

nsiwnf:
# 🔗 Issue

DB not setup in staging

# ✍️ Description

Adding configuration for staging db

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -68,6 +68,11 @@ test:
   #   https://stackoverflow.com/questions/59089991/ruby-2-6-5-and-postgresql-pg-gem-segmentation-fault
   gssencmode: disable
 
+staging:
+  <<: *default
+  host: <%= ENV['STAGING_DATABASE_HOST'] %>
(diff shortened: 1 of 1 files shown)


config/database.yml
```diff
@@ -68,6 +68,11 @@ test:
   #   https://stackoverflow.com/questions/59089991/ruby-2-6-5-and-postgresql-pg-gem-segmentation-fault
   gssencmode: disable
 
+staging:
+  <<: *default
+  host: <%= ENV['STAGING_DATABASE_HOST'] %>
+  database: <%= ENV['STAGING_DATABASE_NAME'] %>
```
edwinthinks, line 74: If this is in Heroku I think it would be `DATABASE_URL` as the ENV?

nsiwnf: Ah, ok I can update that

nsiwnf: Oh, it looks like it's an either/or thing? DATABASE_URL is already set in heroku configs, so Im not sure why it's not working

PR #407, opened 2024-01-03
adding faker for seeding staging env

nsiwnf:
# 🔗 Issue

Faker error on staging seed

# ✍️ Description

adding faker for seeding staging env

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -85,10 +85,13 @@ gem "active_link_to"
 gem "nice_partials"
 
 group :development, :test, :staging do
+  gem "faker"
+end
+
+group :development, :test do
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -84,7 +84,7 @@ gem "rails-i18n"
 gem "active_link_to"
 gem "nice_partials"
 
-group :development, :test do
+group :development, :test, :staging do
```
edwinthinks, line 87: Hmm I might rather put the gems needed in it's own group so that we aren't including `pry` and `debug` in the deployed environment. It isn't that big of a deal though to include it.

edwinthinks (approved):
Added a suggestion which I'am not strong to holding

Issue #402, opened 2024-01-01
As Staff I can add a 'due date' on a pet task


kasugaijin:
**Update:** PR https://github.com/rubyforgood/pet-rescue/pull/409 for this task is almost complete, someone just needs to take it past the finish line :) Please ensure designs adhere to Tasks V2 here [FIGMA](https://www.figma.com/file/0jVgYASUJy0KiX3BVc3dFM/Tasks?type=design&node-id=0-1&mode=design&t=wqtzgNOthljYf2fy-0) (just focusing on the due date design, you can leave the attachments and conditional colouring etc...for another issue).

Staff need to be able to define a due date for a pet task. Not all tasks require a due date.

Acceptance criteria
- Tasks db table has new column called `due_date` type `datetime`
- new/edit task form has a date picker where due date can be selected
- index table has Due Date column showing due date value
- overdue tasks should be easy to identify as overdue if the task is not completed (I suggest using the 'expired' marker from [this page of the Geeks theme](https://geeksui.codescandy.com/geeks/pages/student-subscriptions.html) and placing it next to the date value in the Index table saying 'Overdue'). Maybe a Task model scope to check if a task is overdue or not is a suitable approach.
- tests

Note, I suggest a PR for this issue be made after this PR to modify how we write migrations is approved and merged: https://github.com/rubyforgood/pet-rescue/pull/400

edwinthinks:
I can tackle this one!

kasugaijin:
Hey @edwinthinks I am unassigning you so we can get someone else to finish this off. Thanks for doing most of it :)

kasugaijin:
@nsiwnf this issue is still alive! Pr just needs finishing.

PR #409, opened 2024-01-03
[402] Due date on tasks

edwinthinks:
# 🔗 Issue
#402 

# ✍️ Description

Demo - https://www.loom.com/share/c7a2ec4e5216417e891d4d737d90959c

# 📷 Screenshots/Demos
Doing something fun this time and recording my process while implementing this - 

Part (1) - https://www.loom.com/share/0c7b71212c184367afdf9d44a0113d4a

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -83,6 +83,7 @@ gem "rails-i18n"
 # Provides helper methods to easily add 'active' tag on links
 gem "active_link_to"
 gem "nice_partials"
+gem "local_time"
 
 group :development, :test, :staging do
   gem "faker"
--- a/Gemfile
+++ b/Gemfile
@@ -83,7 +83,6 @@ gem "rails-i18n"
 # Provides helper methods to easily add 'active' tag on links
 gem "active_link_to"
 gem "nice_partials"
-gem "local_time"
 
 group :development, :test, :staging do
   gem "faker"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -242,7 +242,6 @@ GEM
     listen (3.8.0)
       rb-fsevent (~> 0.10, >= 0.10.3)
       rb-inotify (~> 0.9, >= 0.9.10)
-    local_time (2.1.0)
     loofah (2.21.3)
       crass (~> 1.0.2)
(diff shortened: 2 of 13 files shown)


kasugaijin:
@edwinthinks standard failed fyi

kasugaijin:
Looks good! I will watch the video when I get some time this weekend :)
Are you going to add the 'overdue' warning (https://github.com/rubyforgood/pet-rescue/issues/402) on the index and task/show? Or do you want that to be in a separate issue?

edwinthinks:
Thanks @kasugaijin I think I'll include it in here in addition to include local_time to display the due dates

Gemfile
```diff
@@ -83,6 +83,7 @@ gem "rails-i18n"
 # Provides helper methods to easily add 'active' tag on links
 gem "active_link_to"
 gem "nice_partials"
+gem "local_time"
```
edwinthinks, line 86: I didn't actually need this library but I'am thinking we may need this anyways eventually!

edwinthinks:
@kasugaijin ready for re-review, I didn't include local_time like I mentioned before because i don't believe we actually need it yet.

kasugaijin:
The walk through video is great @edwinthinks.
I think this would be good to link to in the ReadMe for new people on how to pick up an issue and make code changes. Thoughts?

kasugaijin:
I agree that description on a task should not be required. I can see many cases where just a name alone will suffice. I have made an issue to remove this contraints on `Task` and `DefaultPetTask` and the `null: false` constraint on default_pet_tasks tabke.

kasugaijin:
Hey @edwinthinks pipeline is not passing. I re-ran the pipeline a few times and it's the same system test. 

Also, when I checked out the branch to play with it the turbo stream is doing something weird with the inline for for me, on create and edit. Not sure why this would work on your machine and not mine.

pet tasks

nsiwnf:
Reverted the local-time change and updated design to match figma

Screen Shot 2024-01-23 at 6 37 53 PM

nsiwnf (approved):
lgtm

Issue #395, opened 2023-12-29
Staff can CRUD Default Pet Tasks


kasugaijin:
As a staff I can CRUD default pet tasks.

Acceptance criteria
- uses the same implementation as the Pet > Tasks tab to CRUD a new task (UI and Turbo Stream)
- I can create, update, edit a default pet task

vmyts539:
May I?

kasugaijin:
All yours @vmyts539

PR #412, opened 2024-01-04
[#395] Staff can CRUD Default Pet Tasks

vmyts539:
# 🔗 Issue
[Issue 395](https://github.com/rubyforgood/pet-rescue/issues/395)

# ✍️ Description
Add CRUD actions for Default Pet Tasks.

# 📷 Screenshots/Demos

https://github.com/rubyforgood/pet-rescue/assets/33605808/3e591bd9-a975-4bb0-b0b1-2bfce93f480a

### DIFF
--- a/app/controllers/organizations/default_pet_tasks_controller.rb
+++ b/app/controllers/organizations/default_pet_tasks_controller.rb
@@ -1,7 +1,7 @@
 class Organizations::DefaultPetTasksController < Organizations::BaseController
   layout "dashboard"
 
-  before_action :verify_organization_for_current_user, only: [:new, :create, :edit, :update, :destroy]
+  before_action :verify_organization_for_current_user
 
   def index
     @default_pet_tasks = DefaultPetTask.all
--- a/app/controllers/organizations/default_pet_tasks_controller.rb
+++ b/app/controllers/organizations/default_pet_tasks_controller.rb
@@ -1,7 +1,59 @@
 class Organizations::DefaultPetTasksController < Organizations::BaseController
   layout "dashboard"
 
+  before_action :set_organization, only: [:new, :create, :edit, :update, :destroy]
+
   def index
     @default_pet_tasks = DefaultPetTask.all
+  end
+
+  def new
+    @task = @organization.default_pet_tasks.build
+  end
+
+  def create
+    @task = @organization.default_pet_tasks.build(task_params)
+
+    if @task.save
+      redirect_to default_pet_tasks_path, notice: "Default pet task saved successfully."
+    else
+      flash.now[:alert] = "Error creating default pet task."
+      render :new, status: :unprocessable_entity
+    end
+  end
+
+  def edit
+    @task = @organization.default_pet_tasks.find_by(id: params[:id])
+    raise ActiveRecord::RecordNotFound if @task.nil?
+  end
+
+  def update
+    @task = @organization.default_pet_tasks.find(params[:id])
+
+    if @task.update(task_params)
+      redirect_to default_pet_tasks_path, notice: "Default pet task updated successfully."
+    else
+      render :edit, status: :unprocessable_entity
+    end
+  end
+
+  def destroy
+    @task = @organization.default_pet_tasks.find(params[:id])
+    @task.destroy
+
+    redirect_to default_pet_tasks_path, notice: "Default pet task was successfully deleted."
+  rescue ActiveRecord::RecordNotFound
+    redirect_to default_pet_tasks_path, alert: "Failed to delete default pet task."
+  end
+
+  private
+
+  def set_organization
+    @organization = current_user.organization
+    raise ActiveRecord::RecordNotFound if @organization.nil?
+  end
+
+  def task_params
+    params.require(:default_pet_task).permit(:name, :description)
   end
 end
--- a/app/controllers/organizations/default_pet_tasks_controller.rb
+++ b/app/controllers/organizations/default_pet_tasks_controller.rb
@@ -1,18 +1,18 @@
 class Organizations::DefaultPetTasksController < Organizations::BaseController
   layout "dashboard"
 
-  before_action :set_organization, only: [:new, :create, :edit, :update, :destroy]
+  before_action :verify_organization_for_current_user, only: [:new, :create, :edit, :update, :destroy]
 
   def index
     @default_pet_tasks = DefaultPetTask.all
   end
 
   def new
-    @task = @organization.default_pet_tasks.build
+    @task = DefaultPetTask.new
   end
 
   def create
-    @task = @organization.default_pet_tasks.build(task_params)
+    @task = DefaultPetTask.new(task_params)
 
     if @task.save
       redirect_to default_pet_tasks_path, notice: "Default pet task saved successfully."
@@ -23,12 +23,12 @@ class Organizations::DefaultPetTasksController < Organizations::BaseController
   end
 
   def edit
-    @task = @organization.default_pet_tasks.find_by(id: params[:id])
+    @task = DefaultPetTask.find_by(id: params[:id])
     raise ActiveRecord::RecordNotFound if @task.nil?
   end
 
   def update
-    @task = @organization.default_pet_tasks.find(params[:id])
+    @task = DefaultPetTask.find(params[:id])
 
     if @task.update(task_params)
       redirect_to default_pet_tasks_path, notice: "Default pet task updated successfully."
@@ -38,7 +38,7 @@ class Organizations::DefaultPetTasksController < Organizations::BaseController
   end
 
   def destroy
-    @task = @organization.default_pet_tasks.find(params[:id])
+    @task = DefaultPetTask.find(params[:id])
     @task.destroy
 
     redirect_to default_pet_tasks_path, notice: "Default pet task was successfully deleted."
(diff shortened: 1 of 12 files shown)


kasugaijin:
Very cool! Instead of the elipses, could we have two buttons for Edit and Delete like we have in the Pet Tasks?
image

vmyts539:
> Very cool! Instead of the elipses, could we have two buttons for Edit and Delete like we have in the Pet Tasks? image

Done
image

app/controllers/organizations/default_pet_tasks_controller.rb
```diff
@@ -1,7 +1,59 @@
 class Organizations::DefaultPetTasksController < Organizations::BaseController
```
kasugaijin, line 1: All looks good. Raising errors in controller actions is different from implementation in other controllers in this app but I am not opposed to it.

One thing, can we use `DefaultPetTask.new` instead of `@organization.default_pet_tasks.build` because `DefaultPetTask` has a belongs_to relatinship to Organization. So any instances will be scoped to the organization automatically thanks to the acts_as_tenant gem. No need to build them using active record associations directly.

vmyts539: > All looks good. Raising errors in controller actions is different from implementation in other controllers in this app but I am not opposed to it.
> 
> One thing, can we use `DefaultPetTask.new` instead of `@organization.default_pet_tasks.build` because `DefaultPetTask` has a belongs_to relatinship to Organization. So any instances will be scoped to the organization automatically thanks to the acts_as_tenant gem. No need to build them using active record associations directly.

I see that this `act_as_tenant` checks the URL to set the tenant, but the routes of default pet tasks are not nested and the request can still be performed without organization as a param
image

Then we have a security issue, I just tried to create a default pet task for another organization by manually changing the organization in the form path
image
if we don't restrict organization param in the controller then this will happen
image

I checked `OrganizationScopable`, and somehow `current_user` got lost on the way here, but the organization is being set to the one that I manually entered and a default pet task is created with the wrong organization.
image

kasugaijin: Ooh I see. Thank you for the information! Can we just scope the routes under an organization? They should be.

vmyts539: > Ooh I see. Thank you for the information! Can we just scope the routes under an organization? They should be.

I guess we can, but not as a part of this task, since it is out of scope for this particular issue as it requires changes throughout the application. For this task, I suggest leaving the issue I've discovered as it is and open another issue that to fix this vulnerability and modify/refactor routes/act_as_tenant functionality.

vmyts539: > Ooh I see. Thank you for the information! Can we just scope the routes under an organization? They should be.

And I will apply changes you've requested a bit later

vmyts539: > Ooh I see. Thank you for the information! I am on my phone so can't look in depth but I see that default let tasks are within the org block in the routes so not sure why it's not being scope to an org. Would we have a similar issue for other resources like Pet? Screenshot_20240106-100736

it's not scoped because it uses `module:`
image
but even if it wasn't that won't add `organization_id` as it will just be `/organizations/default_pet_tasks/` instead of `/atla/default_pet_task`
to achieve what is intended we need to use the example I've provided above 
image

kasugaijin: Yes you're right. I didn't realise this so thanks for explaining it to me. It now explains why in the org/pet controller we call `pet_in_same_organization?(org_id)` (method lives in app controller) on actions that CRUD a record. Perhaps we can implement similar here. It's not perfect, but will be consistent and something we can address at a later time as you stated, not in this task.

kasugaijin: @vmyts539 What do you think about my above comment? No rush! Just making sure you saw it as I did not @ you.

kasugaijin: Great, thank you @vmyts539 . DId you want to split that into two issues?

vmyts539: > Great, thank you @vmyts539 . DId you want to split that into two issues?

Yeah, I'll finish with this one, then I'll investigate the possible changes to the application logic, and based on that, maybe we will create a task to rework stuff

kasugaijin:
I think this is a cleaner approach than how we handle the Organizations::PetController, so I think it makes sense to also implement this in other organization scoped resources to have consistency.

vmyts539:
> I think this is a cleaner approach than how we handle the Organizations::PetController, so I think it makes sense to also implement this in other organization scoped resources to have consistency.
> 
> FYI please run `rails standard:fix`
> 
> We just had a PR https://github.com/rubyforgood/pet-rescue/pull/419/files to remove the `verified_staff` attribute so merging main should fix those broken tests.

I've conducted investigation, should we proceed here or somewhere else?

kasugaijin:
> > I think this is a cleaner approach than how we handle the Organizations::PetController, so I think it makes sense to also implement this in other organization scoped resources to have consistency.
> > FYI please run `rails standard:fix`
> > We just had a PR https://github.com/rubyforgood/pet-rescue/pull/419/files to remove the `verified_staff` attribute so merging main should fix those broken tests.
> 
> I've conducted the investigation, should we proceed here or somewhere else? Also, sorry for taking so much time, kinda had a lot going on

@vmyts539 no problem at all. We appreciate your contributions. Could you please merge main branch into this PR? It should fix the failed tests.

Also, I say let's keep this PR as is so we can get it in. And then make a new issue to address an alternative solution to ensure the user belongs to the org for CRUDing Org scoped resources.

Thanks!

app/controllers/organizations/default_pet_tasks_controller.rb
Ptrboro: Nice find @vmyts539! 👏 
I also think our current setup with `acts_as_tenant` is sometimes problematic. The organization slug is not handled by routes but instead by a custom Middleware. This can be surprising and doesn't always play nice with rails (e.g. we it caused problems with url generation in emails).

Alto it is not perfect I think what you did with `verify_organization_for_current_user` is sufficient for now, we just need to ensure it is called before every action in the admin panel but that's a separate issue 👍

app/controllers/organizations/default_pet_tasks_controller.rb
```diff
@@ -1,7 +1,54 @@
...
   layout "dashboard"
 
+  before_action :verify_organization_for_current_user, only: [:new, :create, :edit, :update, :destroy]
```
Ptrboro, line 4: Any reason why you skipped the index? 🤔 I think we should also check it, we don't want to let users peek into tasks that were added in another org 😄 
```suggestion
  before_action :verify_organization_for_current_user
```

app/controllers/organizations/default_pet_tasks_controller.rb
```diff
@@ -1,7 +1,54 @@
...
+      redirect_to default_pet_tasks_path, notice: "Default pet task saved successfully."
+    else
+      flash.now[:alert] = "Error creating default pet task."
+      render :new, status: :unprocessable_entity
+    end
+  end
+
+  def edit
+    @task = DefaultPetTask.find_by(id: params[:id])
+    raise ActiveRecord::RecordNotFound if @task.nil?
```
Ptrboro, line 27: `.find` will already raise `ActiveRecord::RecordNotFound` exception if the record is missing so we can shorten this to:
```suggestion
    @task = DefaultPetTask.find(params[:id])
```

app/controllers/organizations/default_pet_tasks_controller.rb
```diff
@@ -1,7 +1,54 @@
...
+    end
+  end
+
+  def destroy
+    @task = DefaultPetTask.find(params[:id])
+    @task.destroy
+
+    redirect_to default_pet_tasks_path, notice: "Default pet task was successfully deleted."
+  rescue ActiveRecord::RecordNotFound
+    redirect_to default_pet_tasks_path, alert: "Failed to delete default pet task."
```
Ptrboro, line 46: Typically in these scenarios, applications simply display a 404 page. This usually occurs when a user attempts to delete a record that has already been deleted by someone else moments earlier. Presenting a warning in such cases might be confusing, as users will observe that the pet has in fact been removed from their list.

I think we can simply get rid of this `rescue` and let Rails respond with 404, wdyt? :)

vmyts539: [Peek 2024-01-19 13-42.webm](https://github.com/rubyforgood/pet-rescue/assets/33605808/a25fece1-2022-4d88-b8e2-4490cd68e6f2)

Due to Turbo it will rerender the current page regardless, just without the flash message
image

vmyts539:
@all-contributors
please add @Ptrboro for code.

app/controllers/organizations/default_pet_tasks_controller.rb
Ptrboro: Ok, let's leave it as is then 👍

Ptrboro (approved):
Well done @vmyts539 ! 👏

kasugaijin:
Are you on the ruby for good slack channel? This post was made yesterday re Authorization, and I thought you would be interested based on the discussion above: https://rubyforgood.slack.com/archives/C05JB50LEN5/p1705714171013559

Issue #410, opened 2024-01-03
As admin I can toggle on/off to disable staff members on staff index table


kasugaijin:
Similar to how we toggle on/off Pause Applications in Pets index table, we should do the same for the Staff index table so an admin can toggle on and off to disable/enable a user

Acceptance Criteria
- Staff Index table column name is 'Disabled'
- Use toggle button with turbo stream to update this attribute dynamically (see Pause Applications for how to)

![image](https://github.com/rubyforgood/pet-rescue/assets/95949082/7a548e73-8202-45ef-b3b6-6f2e7a758cb1)

MooseCowBear:
I can work on this

PR #419, opened 2024-01-07
410 admin can toggle to disable staff

MooseCowBear:
# 🔗 Issue

#410 

# ✍️ Description

Adds a deactivate column with a toggle to staff index page with turbo stream response and removes the status column. Removes the verified column from the staff accounts table, replaces references to it with a check for nil in the deactivated_at column, and removes mentions of needing to be verified from email templates. Also updates factories and tests to reflect the change. 

I didn't rename the verified_staff method in the application controller, but it might be more accurate to say active staff now. If you want, I can change that too. 

# 📷 Screenshots/Demos

Screenshot 2024-01-07 at 4 02 50 PM
Screenshot 2024-01-07 at 4 03 06 PM


kasugaijin:
Yes please do change the method name to something more applicable like you mentioned...something like `staff_is_active?`
I think a predicate `?` method makes more sense here. So it should return true or false rather than handle the redirect itself.

kasugaijin:
Merge conflict FYI.

nsiwnf (changes requested):
LGTM - just need to resolve the merge conflicts

MooseCowBear:
Whoops. Okay should be good now.

nsiwnf (approved):

PR #423, opened 2024-01-08
Remove safety_assured from migration

Ptrboro:
# ✍️ Description

This migration was merged after the removal of the strong_migrations gem. However, it still contained a `safety_assured` block, which led to its failure.


kasugaijin (approved):
Thanks @Ptrboro 
I advised this be added to another PR but given it's causing more widespread issues this makes more sense.

Issue #426, opened 2024-01-08
User only sees one error under input when form submit validation fails for edit registration


kasugaijin:
Currently if you go to User > edit registration and make an error in the form e.g., password and submit, it renders an error at the top of the form AND under the input field. We only want an error under the input field (using `bootstrap_form_for`. This should also apply for an attachment e.g., if you try to attach a pdf file, an error should display under the attachment input.

atbalaji:
I would like to work on this issue @kasugaijin

atbalaji:
Hi @kasugaijin , few questions to clarify,

-  Are we removing the top error part across the app to have a uniform ui?

- The top error part removal has been handled by removing https://github.com/rubyforgood/pet-rescue/blob/7261dc0b8a06632507845e78bfc929cd003a89cd/app/views/devise/registrations/edit.html.erb#L14 .
 
- I encountered something else, on providing a random pdf file as attachment for picture, I am encountering the following error 
**ActionController::UrlGenerationError at /users
 No route matches {:action=>"purge", :controller=>"attachments", :id=>nil}, missing required keys: [:id]** here.  Should we create a separate issue for that?

Shall I make a PR with only the removal change?

kasugaijin:
Hi @atbalaji 
Yes please remove the top error only. As part of this issue, we should display the error below the input field for the attachment. As an example, enter a wrong password and submit the form and you will see the password field is highlighted in red with an error.

As for the pdf issue, I just merged a fix for that so please merge main into your branch and that should be resolved.

atbalaji:
Ok @kasugaijin . I have a doubt here . This fixes the issue . But I am having a doubt of should I leave **validates :append_avatar** as it is or should i change this validation to **validate** and add custom methods and check all conditions in it since :append_avatar is a method. Before raising PR wanna know your thoughts


![Screenshot 2024-01-11 010952](https://github.com/rubyforgood/pet-rescue/assets/18101543/0d54529e-3950-483c-a109-8f64a6b8fa14)

kasugaijin:
@atbalaji I think it is OK to just use `validates` and set the conditions therein. It is also how we handle other attachments e.g. on Pet https://github.com/rubyforgood/pet-rescue/blob/main/app/models/pet.rb

atbalaji:
Fine. I will raise a PR after testing and adding possible tests in case needed.

atbalaji:
[...]

PR #435, opened 2024-01-11
Handle Error display on user edit registrations and related to avatarable

atbalaji:
🔗 Issue
#426 

# ✍️ Description
- Removed errors appearing on top of user > edit registrations
- Handled error display on both user > edit registrations as well as Organizational Profiles edit where earlier proper error message was not displayed for Attach Picture
- Removed append_avatar method from avatarable and changed all occurrences of :append_avatar to :avatar
- Handled test cases as well

### DIFF
--- a/app/controllers/organizations/organization_profiles_controller.rb
+++ b/app/controllers/organizations/organization_profiles_controller.rb
@@ -21,7 +21,7 @@ class Organizations::OrganizationProfilesController < Organizations::BaseControl
       :phone_number,
       :email,
       :about_us,
-      :append_avatar,
+      :avatar,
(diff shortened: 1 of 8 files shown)


kasugaijin:
@atbalaji please run `rails standard:fix` to pass the standard check.

atbalaji:
> @atbalaji please run `rails standard:fix` to pass the standard check.

Done @kasugaijin . Earlier when I ran rails  standard:fix i was of the impression i would get suggestions in terminal instead of direct changes. Has reverted my change which introduced standard:fix and added a new commit

kasugaijin (approved):
Looks good.

Issue #418, opened 2024-01-07
Investigate possible flaky tests


kasugaijin:
I have run the remote main branch pipeline multiple times and it passes. However, a few PRs are getting failed tests, seemingly flaky as sometimes they pass and do not. Further, if I run main branch tests locally `rails test:all`, sometimes it passes, other times a test will fail...typically it is:
`Organizations::HomeControllerTest#test_: GET #index should render index template`
`UsersTest#test_user_can_log_out`
`LoginTest#test_: when logging in as a staff member should direct to the organization's pets index view.`

Can someone please try running main branch tests locally and see what you get? And if inclined, see why this might be happening.

MooseCowBear:
Some preliminary results from running the tests on my machine, repeating each of the following runs 50 times each:

```test:all``` Returns the greatest number of failures. I also get a errors saying no file is found for some of the controller tests some of the time. 27 out of 50 contained fails/errors about files not being found, plus 4 out of 50 produced Ferrum:StatusError (not sure what this means or why it's happening and google is not helping)

```test``` Returns 1 failure and 7 errors out of 50. 

```test:system``` 2 failures out of 50.

```test test/system/login_test.rb``` No failures. One Ferrum:StatusError toward the end of the run.

I don't know that I'm equipped to diagnose the cause of flaky tests, but it feels like maybe the greater number of failures/errors when running the full test suite could be a clue?

**Comment by kasugaijin** [comment]
Awesome thank you @MooseCowBear for taking a look and verifying!

**Comment by jmilljr24** [comment]
Just to add so more info if it helps, my PR failed three times before finally working. Once was the HomeControllerTest and twice with PetsControllerTest. No issues locally. I can't really offer any other insight.

```
 Error:
  Organizations::PetsControllerTest#test_: POST #attach_images should attaches an image and redirects to pet photos tab with success flash. :
  Errno::ENOENT: No such file or directory @ rb_sysopen - /home/runner/work/pet-rescue/pet-rescue/tmp/storage/sh/q5/shq5ctzwe0orbwscyu5rjtbcenmh
```
```
  Error:
  Organizations::HomeControllerTest#test_: GET #index should render index template. :
  ActionView::Template::Error: undefined method `city_town' for nil:NilClass
  
[...]

PR #442, opened 2024-01-18
418 flaky tests partial fix

MooseCowBear:
# 🔗 Issue

Partial fix for #418

# ✍️ Description

Changes all path assertions in systems tests to use `has_current_path?`, and adds a block to test_helper to make sure tests only clean up their own attachments. 

It does not fix the intermittent undefined city_town method issue. 

# 📷 Screenshots/Demos


kasugaijin (approved):
thank you!

Issue #350, opened 2023-11-29
As staff I can create a pet in draft so it is not visible to foster/adopters 


kasugaijin:
EDIT: This [PR](https://github.com/rubyforgood/pet-rescue/pull/358) is mostly done! Thank you to @heyapricot for getting it thus far. It just needs to cross the finish line...I think it's just some tests left to go. Anyone want to pick it up? 

Right now, when staff create a pet, it is automatically visible on the pets available for adoption page that fosters and adopters see. We want to change this so staff can create a pet in draft mode so the pet exists in the database, but is not visible to potential adopters, but is visible to staff via the dashboard.

Acceptance criteria
- new column on pets table, called `is_published`, data type `boolean`, `default: false`
- validation on model, required true
- pet new/edit form has a toggle between `draft` and `published` next to, and similar to, the toggle for `applications paused`
- the pet's index table clearly shows when a pet is `published` or `draft` and staff can also toggle the pet between draft and publish via the pets index table, similar to the toggle for `applications paused`
- pet show page summary table displays the value for is_published as Published: 
- update tests for creating/editing a pet and add tests validating that draft pets do not show on the page of adoptable dogs that adopters and fosters see.

kasugaijin:
@heyapricot how's this one going? Hit us up in Slack if you have any questions :)

mononoken:
I would be interested in picking up what is left.

kasugaijin:
Sound good. Thank you @mononoken

PR #443, opened 2024-01-19
[350] As staff I can create a pet in draft so it is not visible to foster/adopters

mononoken:
# 🔗 Issue
https://github.com/rubyforgood/pet-rescue/issues/350

# ✍️ Description
This feature adds a `published` attribute to Pets so that staff can create draft pets that do not display to unauthorized users. I updated the previous work to match the current main branch and added a few things.

## Notes
- I did not update tests for creating/editing pets as I could not find any. I was tempted to add them in this branch, but I thought that these tests are critical and autonomous enough that they probably warrant a separate issue and pull. I did update the index page test so that it checks for published pets displaying. If you do want me to create these creating/editing tests in this PR, let me know. If it will be a new issue, I would be happy to work on it.
- I noticed that all the tests in [test/integration/adoptable_pet_show_test.rb](https://github.com/rubyforgood/pet-rescue/blob/main/test/integration/adoptable_pet_show_test.rb) are skipped with the comment "while new ui is implemented". I took this as a sign that these should not be worked on currently. However, I did add a placeholder test with the other skipped tests.
- I found an error in [test/integration/adoptable_pets_index_test.rb](https://github.com/rubyforgood/pet-rescue/blob/main/test/integration/adoptable_pets_index_test.rb). The tests currently assert using an undefined `@pet_count` variable. The test currently pass despite this error, but in this PR I have it defined so that it works as expected. If this PR does not get pulled, this needs to be fixed separately.
- The original issue calls for this attribute to be called `is_published`. However, this felt very not-Ruby to me and this `is_` format is not used anywhere else in the app, so I renamed it to just `published`.

# 📷 Screenshots/Demos
pet-drafts-index
Left side shows the index interface for staff and the right side demonstrates how the draft pets do not display for unauthenticated users.

pat-drafts-form
Adds published toggle to form as well.

### DIFF
--- a/db/migrate/20231201204221_add_is_published_to_pets.rb
+++ b/db/migrate/20231201204221_add_is_published_to_pets.rb
@@ -1,5 +1,0 @@
-class AddIsPublishedToPets < ActiveRecord::Migration[7.0]
-  def change
-    add_column :pets, :is_published, :boolean, default: true, null: false
-  end
-end
--- a/db/migrate/20231201204221_add_is_published_to_pets.rb
+++ b/db/migrate/20231201204221_add_is_published_to_pets.rb
@@ -1,0 +1,5 @@
+class AddIsPublishedToPets < ActiveRecord::Migration[7.0]
+  def change
+    add_column :pets, :is_published, :boolean, default: true, null: false
+  end
+end
--- a/app/controllers/adoptable_pets_controller.rb
+++ b/app/controllers/adoptable_pets_controller.rb
@@ -8,6 +8,10 @@ class AdoptablePetsController < Organizations::BaseController
 
   def show
     @pet = Pet.find(params[:id])
+    unless @pet.published
+      redirect_to adoptable_pets_path, alert: "You can only view published pets."
+    end
+
     if AdopterApplication.adoption_exists?(current_user&.adopter_account&.id, @pet.id)
       @adoption_application = AdopterApplication.where(pet_id: @pet.id,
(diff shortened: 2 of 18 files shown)


db/migrate/20231201204221_add_is_published_to_pets.rb
```diff
@@ -0,0 +1,5 @@
+class AddIsPublishedToPets < ActiveRecord::Migration[7.0]
```
kasugaijin, line 1: I appreciate seeing the history on the migrations, but could you condense this into one migration?
I agree with naming the column `published` with default of `false` and null false.

you should be able to delete all three, make a new one, drop your local db and rerun migrations.

kasugaijin:
Thank you for taking this on! It looks good. Just one comment re migrations. 

Yes quite a few tests are skipped as they are closely coupled with the DOM and it makes sense to finish such tests when we are happy with the front end. We still want to add new tests for feature functionality right now, but none to really worry about in this PR I think.

mononoken:
> Thank you for taking this on! It looks good. Just one comment re migrations.
> 
> Yes quite a few tests are skipped as they are closely coupled with the DOM and it makes sense to finish such tests when we are happy with the front end. We still want to add new tests for feature functionality right now, but none to really worry about in this PR I think.

Thank you for reviewing! I have the migrations consolidated into one now.

kasugaijin (approved):
Looks good!

Issue #454, opened 2024-01-31
Pet Task Epic: Add colour shading to background of each task depending on its 'status'


kasugaijin:
Completed pet tasks should show green, uncompleted and not overdue should remain white, and uncompleted and overdue should be shaded red, per the image.
Refer to the theme for styling: https://geeksui.codescandy.com/
See Tasks V2 for design: https://www.figma.com/file/0jVgYASUJy0KiX3BVc3dFM/Tasks?type=design&node-id=0%3A1&mode=design&t=vyTdeotEtckQCTFB-1

![image](https://github.com/rubyforgood/pet-rescue/assets/95949082/0f5b166b-1fe5-4b91-a64b-e8fc052c8f73)

tmorton:
👋 I'd like to do this tomorrow.

kasugaijin:
@sparshalc I see you have put up a PR for this. Please in future only put up PRs for issues that you assigned to. Someone else was going to take this one. Not a problem..just a rule for open source.

kasugaijin:
@sparshalc thank you!

PR #460, opened 2024-01-31
Add colour shading to background of each task depending on its 'status' 

sparshalc:
This pr addressed #454  . I have updated the relevant code and add color shading to background of each task depending on its status.
image

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -28,16 +28,15 @@ class Task < ApplicationRecord
   default_scope { order(created_at: :asc) }
 
   def determine_shading_class
-    case
-    when completed?
+    if completed?
       "bg-success-subtle"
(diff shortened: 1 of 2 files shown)


kasugaijin:
@sparshalc please add a screenshot to help reviewers :)

kasugaijin:
This is from my review.

This is a good start. But I think we can restructure this for improved longevity.

How about we create two scopes on the Task model. One called overdue? (due_date < Time.current) and one called not_overdue? (due_date > Time.current). These will be helpful then througout the application.

Now you can just conditionally set the style in the view file based on the result of task.overdue?. No need for a helper method.

Does this make sense?

kasugaijin (approved):
Looks good!

Issue #457, opened 2024-01-31
Pet Task Epic: As staff I can create a recurring Task on a pet


kasugaijin:
We need to build a simple recurring task feature. Rather than worry about automated task creation with background jobs, we will keep it very simple. If a task is recurring, the second instance of the task will be created when the first instance is completed. If the task has due dates, the due date of the new task will be the due date of the first task + the `recurrence_in_days` value. 

Acceptance Criteria
- migration to add `recurrence_in_days` attribute to Tasks table, integer 
- validation on Task model user enters integer
- add field to Task form
- when a recurring task is marked complete, a new task is created that has an identical Name and Description. If there is a due date, the new due date is calcuated.*

*We will need to handle an edge case where a resucrring task with a due date remains overdue beyond the second hypothetical due date. For example, a Task has a recurrence_in_days of 5, and it is due on the 5th, but it remains overdue past the 11th, if we calculate the new due date to be the due date of the first task + the `recurrence_in_days` value, the new task will be created with a due date of the 10th, and alread by overdue. In this case, it should be created with a due date of the 15th.

This is a relatively large chunk of work. If you want to take this on but would like to break it up, or have questions, hit us up in Slack!

MooseCowBear:
I'd like to work on this

kasugaijin:
@MooseCowBear all yours!

PR #464, opened 2024-02-06
457 create recurring task on pet

MooseCowBear:
# 🔗 Issue
#457 

# ✍️ Description

Adds a recurring task to pet by adding columns `recurring` (boolean) and `next_due_date_in_days` (integer) to tasks, and updates the pet task form to include fields for the new columns.

Since it doesn't make sense to have a `next_due_date_in_days` value without the task also being recurring and having a due date, I added some JS to disable the field unless those other two values have been filled in and added a validation on the model to also check for that condition. 

I moved the form submission on checkbox click to a Stimulus controller and added the alert/prevention of form submission in the case someone tries to mark a completed recurring task incomplete. The logic for creating a new recurring task when a recurring task has been marked complete occurs in a service object. 

Let me know if you'd prefer something different. Happy to make changes.

# 📷 Screenshots/Demos

Screenshot 2024-02-06 at 6 41 59 AM

Screenshot 2024-02-06 at 6 42 32 AM

### DIFF
--- a/app/controllers/organizations/tasks_controller.rb
+++ b/app/controllers/organizations/tasks_controller.rb
@@ -24,6 +24,7 @@ class Organizations::TasksController < Organizations::BaseController
   end
 
   def update
+    reset_next_due_if_missing
     if @task.update(task_params)
       if @task.recurring && @task.completed_previously_changed?(from: false, to: true)
         Organizations::TaskService.new(@task).create_next
@@ -71,5 +72,11 @@ class Organizations::TasksController < Organizations::BaseController
 
   def task_params
     params.require(:task).permit(:name, :description, :completed, :due_date, :recurring, :next_due_date_in_days)
+  end
+
+  def reset_next_due_if_missing
+    unless task_params.has_key?(:next_due_date_in_days)
+      @task.next_due_date_in_days = nil
+    end
   end
 end
--- a/app/controllers/organizations/tasks_controller.rb
+++ b/app/controllers/organizations/tasks_controller.rb
@@ -25,6 +25,7 @@ class Organizations::TasksController < Organizations::BaseController
 
   def update
     if @task.update(task_params)
+      Organizations::TaskService.new(@task).create_next
       respond_to do |format|
         format.turbo_stream { render turbo_stream: turbo_stream.replace("tasks_list", partial: "organizations/pets/tasks/tasks", locals: {task: @task}) }
       end
--- a/app/controllers/organizations/tasks_controller.rb
+++ b/app/controllers/organizations/tasks_controller.rb
@@ -24,7 +24,8 @@ class Organizations::TasksController < Organizations::BaseController
   end
 
   def update
-    reset_next_due_if_missing
+    @task.next_due_date_in_days = nil unless task_params.dig(:next_due_date_in_days)
+
     if @task.update(task_params)
       if @task.recurring && @task.completed_previously_changed?(from: false, to: true)
         Organizations::TaskService.new(@task).create_next
@@ -72,11 +73,5 @@ class Organizations::TasksController < Organizations::BaseController
 
   def task_params
     params.require(:task).permit(:name, :description, :completed, :due_date, :recurring, :next_due_date_in_days)
-  end
-
-  def reset_next_due_if_missing
-    unless task_params.has_key?(:next_due_date_in_days)
-      @task.next_due_date_in_days = nil
-    end
   end
 end
--- a/app/controllers/organizations/tasks_controller.rb
+++ b/app/controllers/organizations/tasks_controller.rb
@@ -25,7 +25,10 @@ class Organizations::TasksController < Organizations::BaseController
 
   def update
     if @task.update(task_params)
-      Organizations::TaskService.new(@task).create_next
+      if @task.recurring && @task.completed_previously_changed?(from: false, to: true)
+        Organizations::TaskService.new(@task).create_next
+      end
+
       respond_to do |format|
         format.turbo_stream { render turbo_stream: turbo_stream.replace("tasks_list", partial: "organizations/pets/tasks/tasks", locals: {task: @task}) }
       end
--- a/app/controllers/organizations/tasks_controller.rb
+++ b/app/controllers/organizations/tasks_controller.rb
@@ -66,6 +66,6 @@ class Organizations::TasksController < Organizations::BaseController
   end
 
   def task_params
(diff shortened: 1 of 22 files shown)


kasugaijin:
Awesome @MooseCowBear I will review this week! Looking forward to it.

app/controllers/organizations/tasks_controller.rb
```diff
@@ -25,6 +25,7 @@ def edit
 
   def update
     if @task.update(task_params)
+      Organizations::TaskService.new(@task).create_next
```
kasugaijin, line 28: Ok cool I see what you are doing here. I wonder, for sake of readability on the controller, if we want to check if a new task needs to be created in the Controller, then instantiate the service? Currently, this will instantiate the service regardless, e.g., when a typo is fixed in the name or description, and this is not a necessary step. So, we could add a conditional that calls the service object if `@task.recurring && @task.completed_previously_changed?(from: false, to: true)` on update. Does that make sense?

app/services/organizations/task_service.rb
```diff
@@ -0,0 +1,38 @@
...
+      )
+    end
+  end
+
+  private
+
+  def next_due_date
+    return nil unless @task.due_date
+
+    if overdue?
```
kasugaijin, line 24: I think this is a nice simple approach for MVP for setting the next due date.

app/services/organizations/task_service.rb
```diff
@@ -0,0 +1,38 @@
...
+    else
+      @task.due_date + @task.next_due_date_in_days.days
+    end
+  end
+
+  def need_next?
+    @task.recurring && @task.completed_previously_changed?(from: false, to: true)
+  end
+
+  def overdue?
```
kasugaijin, line 35: We now have a method on Task `def overdue?` to do this. It uses `Time.current` which I think should work fine.

One change it could have, is to check if the task is complete or not...do we want to consider a completed task overdue? Probably not.

kasugaijin:
Very cool @MooseCowBear! Thank you for taking this one on. Hopefully it was fun to work on?
I left some comments. When I am on my personal computer, I will checkout this branch and play around with it in the UI, too. But, I think the implementation looks good so far.

app/controllers/organizations/tasks_controller.rb
MooseCowBear: I was so on the fence about whether to err on the side of keeping the controller code minimal or not, but I think you're right that it's better to keep the logic of it up front.

app/services/organizations/task_service.rb
MooseCowBear: I think for a more general `overdue?` method (one that needs to work outside of the task service object) it should be `Time.current` and not `updated_at`. I was thinking when the service object checks if a task is overdue it only does it if the update was a change from incomplete to complete, meaning the `updated_at` value was the time at which the task became complete. But I don't think `updated_at` would be the right comparison in the general case. Like you said, there's no guarantee in other cases that the `updated_at` value would reflect the time at which the task became completed since there could be other updates that had been made. 

I agree that probably a task that is complete shouldn't be considered overdue. But if `overdue?` becomes something like: 

```
def overdue? 
  due_date < Time.current if due_date && !completed
end 
```

and the service object uses it, won't tasks always be not overdue when the service object gets instantiated? (since those tasks will be completed). 

One option, could be to split off the time aspect into its own method like: 

```
def due_date_passed?
 due_date < Time.current if due_date
end
```

and then overdue? could be:

```
def overdue?
  due_date_passed? && !completed
end
```

Then the service object would use `due_date_passed?` and the view that is checking for overdue-ness could remove the conjunction. 

I don't know if there will be other things that want to check for overdue-ness but if so it might be worth making it more accurate? Intuitive? -- I don't know the word I'm looking for here...

kasugaijin: Ah yes good points, you are right! Then I think your original approach makes sense. We can have a method on the Task for general use, and have a method on the Service object that does what it needs to do.

app/controllers/organizations/tasks_controller.rb
kasugaijin: Yes I think a controller is going to be read more often, so it's better if it tells the story, rather than have to look in the Service object to figure out what is going on.

kasugaijin:
@MooseCowBear I played around with it and I think it's a great first pass at this functionality. Good job.
There was one case that we might want to handle - create a task with a due date, recurring is true, and a value in next due in days. Then edit the task, and turn off recurring and try to save - it will not save. A simple fix could be some stimulus JS to clear the next due in days field if the recurring checkbox is unchecked. Thoughts?

MooseCowBear:
I think that will be a nice solution. I'll make the update.

kasugaijin:
> I think that will be a nice solution. I'll make the update.
> 
> Edit: As I'm looking at this, one other change that might make sense is to move the current inline JS that's enabling/disabling the next_due_date_in_days form field to the Stimulus controller as well. That way all the form related JS stays together. What do you think?

Yes I agree. It makes sense to keep like logic together.

MooseCowBear:
I added the Stimulus controller to remove the next_due_date_in_days value if the recurring checkbox is unchecked, but I ran into a slight complication which was that the task still wouldn't update because a blank value submitted from the form got omitted from the params being passed to the backend. 

I got around this by adding a method to the controller that will manually change the value to nil if the next_due_date_in_days key is missing from the task params, which works, but I'm not sure if there's a better way to handle it. Let me know what you think.

kasugaijin:
> I added the Stimulus controller to remove the next_due_date_in_days value if the recurring checkbox is unchecked, but I ran into a slight complication which was that the task still wouldn't update because a blank value submitted from the form got omitted from the params being passed to the backend.
> 
> I got around this by adding a method to the controller that will manually change the value to nil if the next_due_date_in_days key is missing from the task params, which works, but I'm not sure if there's a better way to handle it. Let me know what you think.

Ah yes I have had to do something like this before. If I recall, rails doesn't submit the param when you set it to `""` with JS, right? 
I think handling this in the controller is OK. We could simplify it by removing the method and just placing the logic as a one-liner, as the first line in the #update action e.g.
`@task.next_due_date_in_days = nil unless task_params.has_key?(:next_due_date_in_days)` 
Or, 
`@task.next_due_date_in_days = nil unless task_params.dig(:next_due_date_in_days)`

kasugaijin:
@MooseCowBear great work!

kasugaijin (approved):

PR #466, opened 2024-02-13
Authorization framework with Action Policy

mononoken:
# 🔗 Issue
This is the beginning of #450 , but it does not close it yet. My goal is to apply the framework to every controller action by the end of it. I will continue to work on the rest of the controllers in the meantime.

# ✍️ Description
This PR is the beginning of the authorization framework with Action Policy. I have not applied the framework to the whole app, but I have applied it to 5 controllers which require both shared and unique configuration.

I wanted to show this PR so that reviewers can get eyes on the progress and confirm it's going in the right direction. If any patterns seem like they could be improved or should be abandoned, please let me know.

This PR will most likely be an intro to Action Policy to anyone who reviews this. I think you will find most of Action Policy intuitive and their docs are awesome. I have a quick summary of some of its features below if it will help you but please feel free to skip it unless you are confused. 

## Summary of Action Policy in pet-rescue
Anything I do not cover can most likely be found in the [Action Policy docs](https://actionpolicy.evilmartians.io/#/).

Take a quick look at the [current `Organizations::PetsController` on main](https://github.com/rubyforgood/pet-rescue/blob/main/app/controllers/organizations/pets_controller.rb). Now check out the new one in the PR. Hopefully, you see the improvement!

All of the authorization logic has been removed from the controller and instead exists in the `Organizations::PetPolicy`. This Policy is used via `authorize!` calls.

For actions involving existing records such as `update`, `destroy`, and `show`. We can call authorize on the record that is fetched in the `set_` method. This will pass the record as a default context to the policy:
```ruby
  def set_pet
    @pet = Pet.find(params[:id])

    authorize! @pet
  end
```
Action Policy finds the correct policy implicitly using the name of the model object passed to `authorize!`. If the controller is namespaced, it will look for the policy in the same namespace too. The lookup behavior is further explained in the docs.

However, some actions don't use an existing record, such as `new`. In this case, you can pass the class name to help Action Policy find the correct policy. In cases where there is no record, you may require more context than the defaults. The defaults are the current user and the record passed. In `new` here, I pass the `Current.organization` as an organization context so that in the Policy, we can make sure the current user's organization matches the current one.

Action Policy also support scoping. This is very useful for `index` actions. In this `index`, I call the scope using `authorized_scope` which returns `relation.where(organization: user.organization)` (relation being the `@q.result` `ActiveRecord::Relation` it was passed).

In `Organizations::PetPolicy`, you will see that you can alias rules so that rules that share common logic don't have to be rewritten. 

If you are not familiar with Pundit, Action Policy looks up the rule for the action by looking in the policy for a rule that matches the method name plus `?`. Action Policy also can have a default rule specified and out-of-the-box comes with `manage?` as a default rule, except toward `index?`, `new?`, and `create?`. So, in `Organizations::PetPolicy`, we just need to define what `manage?` checks which in this case is a call to `permission?` with `:manage_pets`. `permission?` is **not** a part of Action Policy. I will explain that below.

For logic that is shared between many rules, you can define pre-checks. In this case, I have a pre-check for both `verify_organization!` and `verify_active_staff!`. These are defined in `ApplicationPolicy` since they are shared.

For `verify_organization!`, I made some adjustments so that we can use `manage?` for both actions that have an existing record and those that don't. Recall that `update` uses `authorize! @pet` but `new` does not have an existing `@pet` to pass, so it is given the organization context directly. I have:
```ruby
  def organization
    @organization || record.organization
  end

  def verify_organization!
    deny! unless user.organization_id == organization.id
  end
```
The organization getter will get the `@organization` context if it exists. If it does not, it assumes a record was passed and gets the organization from that. That way, `organization` is available for `verify_organization!` with the same logic for both cases.

I have tested authorization using the recommendations from Action Policy. palkan recommends that you unit test each Policy and then you simply need to test that the controller actions are using the proper authorization. No need to make multiple request tests for different user contexts. I really like this method and it works cleanly and is fast.

One note on tests though, is that I currently create the context objects in the policies unit tests. I want to refactor these into using `build_stubbed` instead but the one thing that I think may presently hinder that is `rolify`. However, I *think* fixing that will be easy.

## `permission?` and authorizable.rb
In our policies, we could have rules like this instead:
```ruby
def manage?
  user.has_role?(:staff) || user.has_role?(:admin)
end
```
We could DRY it for other tests like this:
```ruby
def manage?
  staff? || admin?
end

private

def staff?
  user.has_role?(:staff)
end

def admin?
  user.has_role?(:admin)
end
```
What I do not like about that is that it only becomes clear what each role is permitted to do in the policy. These permissions can exist either in our heads or in the code. The `Authorizable` concern is an attempt to get this in the code. It also has central definitions for each role, so you can look in one space to find what is permitted for each role.

I am not completely happy with `Authorizable` as I have it written. For one, it is not open/closed. To make it so, I think we could easily create a `permissions` table and model, that can be associated with users. It would be useful for customizing individual permissions, if that is something clients may want. We could use roles simply for setting default permissions to users in this scenario.


kasugaijin:
Awesome @mononoken I will review this week! Looking forward to seeing what you have done.

app/controllers/organizations/pets_controller.rb
```diff
@@ -1,34 +1,33 @@
 class Organizations::PetsController < Organizations::BaseController
```
kasugaijin, line 1: Way better :+1:

db/migrate/20240201102207_rolify_create_roles_with_users.rb
```diff
@@ -0,0 +1,18 @@
+class RolifyCreateRolesWithUsers < ActiveRecord::Migration[7.0]
```
kasugaijin, line 1: I think assigning roles to the user will simplify things. Good idea.

app/policies/application_policy.rb
```diff
@@ -0,0 +1,45 @@
+# Base class for application policies
```
kasugaijin, line 1: :+1:

kasugaijin:
@mononoken awesome work so far. Thank you for taking this on. I think Action Policy is similar enough to Pundit, and easy enough to follow, that it gets my vote for use here. To your points regarding permission? and authorizable.rb, I agree that it's better to explicitly name each policy based on what it is actually doing, and map those to roles like you have done. I am not opposed to how you have handled it in authorizable.rb. It very easy to modify, and also easy to read and understand what each role has authorization to do. Having it in a table would detract from this...but I am open to discussion on that!

mononoken:
> @mononoken awesome work so far. Thank you for taking this on. I think Action Policy is similar enough to Pundit, and easy enough to follow, that it gets my vote for use here. To your points regarding permission? and authorizable.rb, I agree that it's better to explicitly name each policy based on what it is actually doing, and map those to roles like you have done. I am not opposed to how you have handled it in authorizable.rb. It very easy to modify, and also easy to read and understand what each role has authorization to do. Having it in a table would detract from this...but I am open to discussion on that!

Thank you for your time reviewing, @kasugaijin ! 

> Having it in a table would detract from this...but I am open to discussion on that!

I think I agree I am not sure I am the biggest fan of the table idea. In thinking and researching other options that would make the permissions code more open to extension, I found [this comment by palkan (author of Action Policy)](https://github.com/palkan/action_policy/discussions/246). He makes Permissions an attribute of users instead and stores them in JSONB field (or array). I think that's a better idea.

That said, I think authorizable.rb as it is currently is working fine and could be left as is until the need for permissions to become more extensible arises. Like, if client admins wanted the option to customize their individual staffs' permissions.

kasugaijin:
@mononoken 
> That said, I think authorizable.rb as it is currently is working fine and could be left as is until the need for permissions to become more extensible arises. Like, if client admins wanted the option to customize their individual staffs' permissions.

This is my sentinment, also. For MVP, what you have is a practical enough solution that is easy to understand, tells the story of what roles have what policies, and easy enough to update. If there comes such a time where we need more fine control of policies at the user level, we can explore (I think a JSON attribute would work fine).

mononoken:
Ah, just realized. @kasugaijin, I am not sure if you wanted to push the reviewed content that I had on this PR or wait until I have Action Policy implemented in the rest of the controllers. I just got back from a trip and started working again in the branch, but if you wanted to do the former I could fix any potential merge conflicts and start another branch. Let me know whatever you prefer.

kasugaijin:
> Ah, just realized. @kasugaijin, I am not sure if you wanted to push the reviewed content that I had on this PR or wait until I have Action Policy implemented in the rest of the controllers. I just got back from a trip and started working again in the branch, but if you wanted to do the former I could fix any potential merge conflicts and start another branch. Let me know whatever you prefer.

Sorry...somehow I missed this question @mononoken. I think we can wait as it looks like you are making good progress, and would make sense to review all at once :)

app/policies/adopter_application_policy.rb
```diff
@@ -0,0 +1,33 @@
...
+  authorize :pet, optional: true
+
+  pre_check :verify_adopter_profile!, only: %i[index? create?]
+  pre_check :verify_pet_adoptable!, only: %i[create?]
+
+  def update?
+    applicant? && permission?(:withdraw_adopter_applications)
+  end
+
+  def index?
```
kasugaijin, line 15: should this also check for applicant? to make sure a user only sees their applications.

test/policies/adopter_application_policy_test.rb
```diff
@@ -0,0 +1,189 @@
...
+    context "when user is adopter with profile" do
+      setup do
+        @user = create(:user, :adopter_with_profile)
+      end
+
+      should "return true" do
+        assert_equal @action.call, true
+      end
+    end
+  end
```
kasugaijin, line 99: Adding to above comment - adopter user should only see the applications they made.

test/policies/adopter_application_policy_test.rb
```diff
@@ -0,0 +1,189 @@
...
+      setup do
+        @user = nil
+      end
+
+      should "return false" do
+        assert_equal @action.call, false
+      end
+    end
+
+    context "when user is adopter without adopter account" do
```
kasugaijin, line 121: I woud rename this when user not associated with adopter account - minor detail really, but it's the association that makes them an adopter.

app/models/concerns/authorizable.rb
```diff
@@ -23,14 +23,15 @@ def permissions
     ADOPTER_PERMISSIONS + %i[
       manage_pets
       manage_tasks
+      view_organization_dashboard
```
kasugaijin, line 38: So I think we will end up with a separate dashboard for Org Staff, and another for Adopters and Fosterers. So, we can remove adopter ability to view the org dashboard, which is for staff.

test/policies/organizations/dashboard_policy_test.rb
```diff
@@ -0,0 +1,72 @@
...
+      setup do
+        @user = nil
+      end
+
+      should "return false" do
+        assert_equal @action.call, false
+      end
+    end
+
+    context "when user is adopter" do
```
kasugaijin, line 32: this test name conflicts with the factory argument ie user is not an adopter

app/controllers/adopter_applications_controller.rb
```diff
@@ -1,38 +1,42 @@
...
+  verify_authorized
+
+  before_action :authenticate_user!
 
   def index
-    @applications = AdopterApplication.where(adopter_account_id:
-                                             current_user.adopter_account.id)
+    authorize!
+
+    @applications = authorized_scope(AdopterApplication.all)
```
kasugaijin, line 9: so this `authorized_scope` is?

```
relation_scope do |relation|
    relation.where(organization: user.organization)
  end
```

kasugaijin: So this will allow an adopter to see all adopter applications scoped under the org. This should also restrict applications to show only those made by the adopter. They should not see any one elses.

app/controllers/organizations/pets_controller.rb
```diff
@@ -1,34 +1,35 @@
...
+  verify_authorized
```
kasugaijin, line 2: So, for my benefit (have not looked at the docs so excuse my laziness) this method will run the prechecks in the org pet policy based on naming convention? i.e. Organizations::PetPolicy

app/models/concerns/authorizable.rb
```diff
@@ -0,0 +1,56 @@
...
+    withdraw_adopter_applications
+    create_adopter_profiles
+    manage_adopter_profiles
+  ].freeze
+
+  STAFF_PERMISSIONS = (
+    ADOPTER_PERMISSIONS.excluding(
+      %i[
+        create_adopter_profiles
+        manage_adopter_profiles
```
kasugaijin, line 32: staff don't need to create adopter applications for MVP either

kasugaijin:
Wow amazing @mononoken this is a lot of really good work. Thank you.
I do not have any issues with the 'how'. It all makes sense and will make managing user auth way more rigorous. There are a couple comments on the 'what' i.e., specific user authorizations, but that is easy stuff to fix. I am happy with the implementation. @nsiwnf over to you if you have any additional comments!

app/policies/adopter_application_policy.rb
mononoken: The checking for the applicant ownership for `index?` is actually done in the `relation_scope` instead of in `index?`. So far, I have used scoping in policies only for index actions for things like this. The scope is missing in this commit though. I added the scope in the next commit once I realized it was missing in this first iteration. c0e41c5

test/policies/adopter_application_policy_test.rb
mononoken: This is related to the other comment reply. I did add a test for `relation_scope` to test the scope in the next commit c0e41c5.

Sorry for the confusion! I should have probably amended this commit once I made these changes before making the next commit.

test/policies/adopter_application_policy_test.rb
mononoken: Good catch!

I do have a question for best practice in this project. Would it be better for me to:
1. Rebase this commit with the fix
2. Rebase a new commit after this commit to fix
3. Add a new commit at the end of the branch to fix

app/controllers/organizations/pets_controller.rb
mononoken: This method is actually used to check that every action in the controller has a policy check called within it. For example, if `#new` did not have any authorization logic in it, an error would raise when calling that action. The final form should have this call only in `ApplicationController`. That way, all controller actions should require some form of authorization. I am currently adding it individually to controllers as I work on them so that I can still navigate to unauthorized pages when testing stuff in development.

There is also `skip_verify_authorized` that can be used to skip this requirement for specific actions, like seen here:
https://github.com/mononoken/pet-rescue/blob/e1d6102e4b7877242fc1cc532132bbcbc57114a7/app/policies/adoptable_pet_policy.rb#L2C3-L2C57

app/controllers/adopter_applications_controller.rb
mononoken: You are correct. `relation_scope` is the 'default scope' that is used for `authorized_scope` for Rails with Action Policy. (More detail: The "relation" in `relation_scope` actually refers to `ActiveRecord::Relation`. When `authorized_scope` is called with an `ActiveRecord::Relation`, it will match with this `relation_scope` in the policy.)

Sorry again for the confusion. This is also related to the next commit as this [comment](https://github.com/rubyforgood/pet-rescue/pull/466#discussion_r1508533397) mentions. In that commit, I am overwriting the `relation_scope` defined in `ApplicationPolicy` by redefining it in `AdopterApplicationPolicy`.

At some commit, I did end up adding this comment to help clarify this:
```ruby
  # Default authorized_scope; override for individual policies if necessary.
```

In regards to the default in `ApplicationPolicy`:
```ruby
  relation_scope do |relation|
    relation.where(organization: user.organization)
  end
```
I thought this would be a good default, since I don't think there is any action where we would want a user to be able to scope outside of their own org.

mononoken:
> Wow amazing @mononoken this is a lot of really good work. Thank you. I do not have any issues with the 'how'. It all makes sense and will make managing user auth way more rigorous. There are a couple comments on the 'what' i.e., specific user authorizations, but that is easy stuff to fix. I am happy with the implementation. @nsiwnf over to you if you have any additional comments!

Thanks a bunch for taking the time to review @kasugaijin ! I know this was a lot to look over. Once I have an answer to [this comment](https://github.com/rubyforgood/pet-rescue/pull/466#discussion_r1508540306), I will get your comments fixed.

test/policies/adopter_application_policy_test.rb
kasugaijin: I think it is OK to add a new commit at the end to keep it easier for you. My approach here in review could have been better, too. I started going through commits just because I wanted to see things in chunks. But, then I just looked at the whole picture via Files Changed in the end. So, I realise some of my comments on things were really no longer relevant due to later changed. Apologies!

app/policies/adopter_application_policy.rb
kasugaijin: Ahh I see ok thank you for clarifying that.

app/controllers/adopter_applications_controller.rb
kasugaijin: Yes I agree that is a good default.

app/controllers/organizations/pets_controller.rb
kasugaijin: Ah very cool. Ok thank you for the explanation!

test/policies/adopter_application_policy_test.rb
mononoken: ca924d545f7d063ee5adeeda43cb272826fdb30f resolves

mononoken:
@kasugaijin I made the changes mentioned in your comments.

kasugaijin:
> @kasugaijin I made the changes mentioned in your comments. Let me know if I missed any.

Looks good. Thank you!

nsiwnf:
This is looking great @mononoken!

mononoken:
Uff da! Going on over a month, but I think I am done implementing Action Policy to every controller!

However, there are two tests that were just added a few hours ago that are failing. I have [comments on those here](https://github.com/rubyforgood/pet-rescue/commit/393530f8a6da6a847b8c95892c910de806c28d07#r139607412).

For @kasugaijin and @nsiwnf , I think not too much has changed since your last reviews as most of the remaining stuff was able to reuse logic and patterns. I think the biggest differences since the last review are:
1. I have `verify_authorized` in `ApplicationController` which means if any controller action is not authorized and called, it will raise an error. Actions that do not require authorization have added `skip_verify_authorized` to make it explicit that they are skipping authorization.
2. I added a rescue to `ApplicationPolicy` that will redirect users when performing an unauthorized action. I want to further customize this behavior to give more actionable feedback to users, but I think this is fine for now. 

In regards to 2, ActionPolicy seems to have good support for streamlining providing failure reasons with i18n support. I may work on implementing that next in a separate issue.

app/controllers/application_controller.rb
```diff
@@ -13,12 +15,6 @@ def switch_locale(&action)
 
   private
 
-  def adopter_with_profile
-    return if current_user.adopter_account&.adopter_foster_profile
-
-    redirect_to root_path, alert: "Unauthorized action."
-  end
-
   def check_if_adopter
```
nsiwnf, line 18: I think this can be deleted too - looks like it's not referenced anymore

app/controllers/matches_controller.rb
```diff
@@ -1,10 +1,13 @@
 class MatchesController < ApplicationController
-  before_action :active_staff, :same_organization?
+  before_action :set_pet, only: %i[create]
+  before_action :set_match, only: %i[destroy]
 
   before_action :set_pet, only: %i[create]
   before_action :set_match, only: %i[destroy]
 
   def create
+    authorize! context: {organization: @pet.organization}
```
nsiwnf, line 9: just curious - could `authorize_scope` be used here?

app/controllers/matches_controller.rb
```diff
@@ -1,10 +1,13 @@
 class MatchesController < ApplicationController
-  before_action :active_staff, :same_organization?
+  before_action :set_pet, only: %i[create]
+  before_action :set_match, only: %i[destroy]
```
nsiwnf, line 3: Are these lines duplicated on lines 5-6?

app/controllers/organizations/adoption_application_reviews_controller.rb
```diff
@@ -1,9 +1,17 @@
...
-  before_action :active_staff
+  before_action :set_adopter_application, only: %i[edit update]
+
   layout "dashboard"
 
   def index
-    @q = Pet.org_pets_with_apps(current_user.staff_account.organization_id).ransack(params[:q])
+    authorize! AdopterApplication,
+      context: {organization: current_user.organization},
+      with: Organizations::AdopterApplicationReviewPolicy
```
nsiwnf, line 9: Do we need the `with` here? It looks like it would be handled correctly -> https://github.com/palkan/action_policy/blob/master/docs/namespaces.md

app/controllers/organizations/invitations_controller.rb
```diff
@@ -1,20 +1,25 @@
 class Organizations::InvitationsController < Devise::InvitationsController
   include OrganizationScopable
 
-  before_action :require_organization_admin, only: [:new, :create]
   layout "dashboard", only: [:new, :create]
 
   def new
+    authorize! StaffAccount, context: {organization: Current.organization},
+      with: Organizations::InvitationPolicy
+
```
nsiwnf, line 9: Little confused about this part - so in the policy it's checking if this current_user has `:invite_staff` permission. How does it know which set of roles to check for? How is `StaffAccount` being used here?

app/controllers/organizations/pets_controller.rb
```diff
@@ -45,7 +44,7 @@ def create
   end
 
   def update
-    if pet_in_same_organization?(@pet.organization_id) && @pet.update(pet_params)
+    if @pet.update(pet_params)
```
nsiwnf, line 47: Would we need an `authorize!` here as well?

nsiwnf: oh nvm - i see thats happening in set_pet

test/policies/adopter_profile_policy_test.rb
```diff
@@ -0,0 +1,166 @@
...
+  include PetRescue::PolicyAssertions
+
+  context "#new?" do
+    setup do
+      @policy = -> { AdopterFosterProfilePolicy.new(AdopterFosterProfile, user: @user) }
+    end
+
+    should "be an alias to :manage?" do
+      assert_alias_rule @policy.call, :new?, :create?
+    end
```
nsiwnf, line 14: Seems like the test should be `be an alias to :create?`

db/seeds/01_alta.rb
```diff
@@ -25,7 +25,7 @@
     user_id: @user_staff_one.id
   )
 
-  @staff_account_one.add_role(:admin, @organizaton)
+  @user_staff_one.add_role(:admin, @organization)
```
nsiwnf, line 28: Good catch 😬

test/models/role_test.rb
```diff
@@ -1,14 +1,7 @@
...
-    should have_and_belong_to_many(:staff_accounts).join_table(:staff_accounts_roles)
-    should belong_to(:resource).optional
-  end
-
-  context "validations" do
-    should validate_inclusion_of(:resource_type).in_array(Rolify.resource_types).allow_nil
-  end
+  # test "the truth" do
+  #   assert true
+  # end
```
nsiwnf, line 6: Do we need tests here?

nsiwnf (approved):
lgtm 🎉 - sorry for all the one off comments btw

app/controllers/application_controller.rb
mononoken: Good catch! There were a few other old methods that I could remove as well and just removed in 73a4192a6547a1661bfa1585c5333713eafe3e53.

app/controllers/matches_controller.rb
mononoken: Not sure how that happened

app/controllers/organizations/adoption_application_reviews_controller.rb
mononoken: We would need `with` here, at least with how I had the policy written. I thought we could use namespace here as well. However, the lookup pattern is looking at the model of the object passed which does not match the controllers name, so if we do:
```ruby
    authorize! AdopterApplication,
      context: {organization: current_user.organization},
      namespace: Organizations
```
It would look for the policy `Organizations::AdopterApplicationPolicy`, not `Organizations::AdoptionApplicationReviewPolicy`. 

However, you got me thinking, we should just rename the policy to match the model in the first place, and then we can also use the namespace lookup. I've implemented that in b90e8319dd392c227b12d854e4a35fd5afef078a. This is also better since the policy actually didn't match the controller anyway because the controller name uses "adoption" not "adopter".

mononoken: **BUT**, this brings up a different point for the authorized_scope usage here. I was forcing the scope to use the `Organizations::AdopterApplicationReviewPolicy` for getting the `relation_scope` here before. If I remove that `with`, it actually is going to go for the `Organizations::PetPolicy` to check the `relation_scope` there. In this particular case, both policies are using the same defaults from `ApplicationPolicy`, so it does not make an immediate difference.

It makes me almost wonder if @q is grabbing the wrong relation here. This is an `index` for `AdoptionApplicationReviews` but it's not indexing applications, it's indexing pets (with the apps merged with pets). But AdopterApplications already have a pet reference so couldn't we just use ransack on the adopter applications themselves? 

I've made a commit to remove the `with` from the scope and change the controller test to check that the controller scope is grabbing from `Organizations::PetPolicy`. c47110f004294a11a77694b485bf0db990e49a91

Not sure how I feel about it. I'll sleep on it and would love to here your thoughts @nsiwnf

app/controllers/organizations/invitations_controller.rb
mononoken: I found the controllers that do not match with a model the hardest to write policies for, like this one. 

You are correct. There is no existing record to check, so all this authorization is doing is checking that the user has the `:invite_staff` permission, and that the user's organization matches with the current organization (via the `pre_check :verify_organization!`.

`StaffAccount` does not actually get used here. We could have passed anything, a string `"invitation"` or symbol `:invitation`. However, I felt it was best to pass the class of the model that is associated with the action. So in this case, we are building a new `StaffAccount` so I passed that class to the policy. 

Action Policy by default always passes a user context from `current_user`. The `User` model handles `permission?` via the `authorizable` concern. In this way, policies do not know which roles to check for. All they know is that user's reply to `permission?` 

I personally really like this pattern and don't think policies need to know about roles at all. All roles do for this implementation, is store a list of permissions. Then, the policies check for those permissions. In the future if it was desired, we could even add the permissions as an attribute to users so that user's could have individual sets of permissions that could be unique. Note, I wrote `authorizable` so the permissions and roles implementation is not part of Action Policy.

Hope that helps. Let me know if I can further clarify

test/models/role_test.rb
mononoken: I don't think we need to test the roles. They are being implemented by rolify, and they're mostly just data.

However, now that you mention this, I think we should write tests for `authorizable`'s public methods.

mononoken:
> lgtm 🎉 - sorry for all the one off comments btw

No problem at all! I appreciate your thoroughness a ton! I felt a little guilty making a big PR knowing y'all would have to review it.

app/controllers/matches_controller.rb
mononoken: Yup, we could use `authorized_scope` in any action really.

In some ways, I think scoping everywhere would be ultimate security. That could potentially avoid ever loading unauthorized data, rather than grabbing data and then denying access. But I have read that it is an easy path to lower performance. Because of this, I opted only to use scoping for `index` actions only.

app/controllers/organizations/invitations_controller.rb
nsiwnf: Thanks for the explanation!

test/models/role_test.rb
nsiwnf: Agreed^

app/controllers/matches_controller.rb
nsiwnf: oh yeah that makes sense

app/controllers/organizations/adoption_application_reviews_controller.rb
nsiwnf: 😵‍💫  Yeah, i see how this page indexing is weird - and yeah, seems like we should be able to group by pet through adoption applications - that might be more within the integrity of this controller., but honestly, I think it's fine as it is if the end result is what we expect! Also, I totally missed the "reviews" part of the policy 🤦. The rename to `AdopterApplicationPolicy` makes more sense to me tho!

mononoken: Yeah, I think I agree with you to leave the ransack search as is. I looked at a bit, and I can see why it was done this way. The page is using a partial from pets to show each pet's associated applications together. Changing it to index the applications, we would ultimately end up mapping the applications into pets anyway to keep using the shared partial.

app/controllers/concerns/organization_scopable.rb
```diff
@@ -15,20 +15,17 @@ def set_tenant
...
 
   def after_sign_in_path_for(resource_or_scope)
-    if resource_or_scope.staff_account&.deactivated?
-      adoptable_pets_path
-    else
+    if allowed_to?(
+      :index?, Pet, namespace: Organizations,
+      context: {organization: Current.organization}
+    )
       pets_path
```
kasugaijin, line 22: We should probably redirect to the dashboard_index_path, now we have a solid grasp on what the dashboard looks like. I think we set it to pets_path before the dashboard index was a thing.

app/controllers/organizations/staff_controller.rb
```diff
@@ -1,21 +1,19 @@
...
   def deactivate
-    if @staff_account.user != current_user
-      @staff_account.deactivate
-      respond_to do |format|
-        format.html { redirect_to staff_index_path, notice: "Staff account deactivated." }
-        format.turbo_stream { render "organizations/staff/update" }
-      end
-    else
-      redirect_to staff_index_path, alert: "You can't deactivate yourself."
+    @staff_account.deactivate
```
kasugaijin, line 13: What's stopping a staff user deactivating themself here, now?

kasugaijin: Ah nevermind I see it in the staff policy!

app/policies/adopter_application_policy.rb
```diff
@@ -0,0 +1,44 @@
+class AdopterApplicationPolicy < ApplicationPolicy
+  authorize :pet, optional: true
```
kasugaijin, line 2: Do we need to change the naming of adopter_profile to adopter_foster_profile here?

app/policies/organizations/adopter_application_policy.rb
```diff
@@ -0,0 +1,16 @@
+class Organizations::AdopterApplicationPolicy < ApplicationPolicy
```
kasugaijin, line 1: Should we name this policy AdopterApplicationReviewPolicy?

kasugaijin (commented):
Phew what a PR! Great work @mononoken this is a huge upgrade. I have left a couple of small comments. Otherwise, I think it is good to go. There might be one or two things we missed in the review here given the PR's size, but we can fix those as we find them (if any).

app/policies/organizations/adopter_application_policy.rb
nsiwnf: `AdopterApplicationPolicy` makes more sense since it's still related to how applications are managed - https://github.com/rubyforgood/pet-rescue/pull/466#discussion_r1525688918

kasugaijin: Ok I was initially a little confused as we have a controller AdopterApplicationsController that enables an adopter to make an application then we have the AdopterApplicationReviews controller for reviewing said applications on the staff side. This current naming is not aligned in that sense?

kasugaijin: Ok I see there was discussion on this already.

app/policies/adopter_application_policy.rb
mononoken: I think I will update this to just `verify_profile!` instead to decouple the method from the account type naming conventions. Any objections to that?

app/controllers/concerns/organization_scopable.rb
mononoken: Updated the redirect to dashboard now in ada4ef41cc10004f50d5153d23fbbb6c54873383

mononoken: On that note too, I think the dashboard routes could probably be cleaned up. We don't need the full set of RESTful routes like destroy and create for dashboards I am thinking. Not in the scope of this PR but I'll make an issue for it. Probably a good first issue.

app/controllers/organizations/adoption_application_reviews_controller.rb
```diff
@@ -1,9 +1,15 @@
 class Organizations::AdoptionApplicationReviewsController < Organizations::BaseController
-  before_action :active_staff
+  before_action :set_adopter_application, only: %i[edit update]
+
   layout "dashboard"
 
   def index
-    @q = Pet.org_pets_with_apps(current_user.staff_account.organization_id).ransack(params[:q])
+    authorize! AdopterApplication,
+      context: {organization: current_user.organization}
```
nsiwnf, line 8: Just noticed this - should we have a `before_action authenticate_user` or do this in case someone tries to access this page without logging in?
`context: {organization: Current.organization}`

mononoken: You are right! That should have been Current.organization and was def a mistake.

Issue #455, opened 2024-01-31
Pet Task Epic: As staff I can use a toggle to update the pet task list to show/hide completed tasks


kasugaijin:
As staff I can use a toggle to update the pet task list to show/hide completed tasks.
This should be implemented dynamically with Hotwire.
Toggle should be designed and placed as shown.

Refer to the theme for styling: https://geeksui.codescandy.com/
See Tasks V2 for design: https://www.figma.com/file/0jVgYASUJy0KiX3BVc3dFM/Tasks?type=design&node-id=0%3A1&mode=design&t=vyTdeotEtckQCTFB-1

Completed shown
![image](https://github.com/rubyforgood/pet-rescue/assets/95949082/57d41fa7-5685-485d-933a-80f35659f6e3)

Completed hidden
![image](https://github.com/rubyforgood/pet-rescue/assets/95949082/ac660fff-09f7-4685-a65c-33e2907ac8df)

sparshalc:
I'd love to give this a try!

kasugaijin:
All yours @sparshalc

sparshalc:
It would be wise to examine pull request #460 before moving forward with this matter, correct? @kasugaijin

kasugaijin:
@sparshalc i left a review on that PR. Not sure if you saw it? I think there has been an issues with visibility or notifications when someone requests a review a few times now.

kasugaijin:
@sparshalc oh weird! I will copy and paste it to a comment!

kasugaijin:
@sparshalc this issue is now unblocked!

PR #470, opened 2024-02-19
Add checkbox to show/hide completed tasks.

sparshalc:
This pr addressed #455  . I have updated the relevant code and added a checkbox to show/hide completed tasks.

image
image


kasugaijin (approved):
Looks good! I was originally thinking of turbo stream to update the collection, but I now prefer your implementation to just use one collection and hide/show tasks. More efficient.

PR #472, opened 2024-02-21
471 Sort Pet Tasks

### LINKED ISSUE #471: Pet Task Epic: As staff I see uncompleted pet tasks above completed pet tasks.
Acceptance criteria:
- Uncompleted pet tasks show above completed pet tasks. 
- Uncompleted tasks with a due date show above uncompleted tasks without a due date, and are ordered by due date (ascending) so the closest (or most overdue) due dates show higher. 
- Uncompleted tasks without a due date are ordered with most recent being placed higher (created_at and/or updated_at) (open to alternative ideas here also)
- Completed tasks are ordered the same as above based on due date, or date created/edited.

Currently we are ordereing tasks by making a query in the view. Now that we have a bit more complexity, it's probably better to move this to the Task model and do the ordering of the collection with a scope or method that we can call in the view.

jmilljr24:
Resolves #471 

# ✍️ Description

This adds a method to the Task model to sort the order in which tasks are returned to the view. I've added scopes for the different criteria then concat the tasks into a single list in the method. I'm not sure if there is a way to do this all in a single query (or if that would provide much benefit) because some parts of the list are ascending and some descending. I am open to suggestions.
    

- [x] Uncompleted pet tasks show above completed pet tasks.

- [x] Uncompleted tasks with a due date show above uncompleted tasks without a due date, and are ordered by due date (ascending) so the closest (or most overdue) due dates show higher.

- [x] Uncompleted tasks without a due date are ordered with most recent being placed higher (created_at and/or updated_at) (open to alternative ideas here also)

- [x] Completed tasks are ordered the same as above based on due date, or date created/edited.

# 📷 Screenshots/Demos

[Screencast from 02-21-2024 07:46:34 AM.webm](https://github.com/rubyforgood/pet-rescue/assets/16829344/d10809e0-2a6b-4bf4-b7c7-1a619505e3c6)

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -24,9 +24,19 @@ class Task < ApplicationRecord
 
   validates :name, presence: true
 
-  default_scope { order(created_at: :asc) }
+  scope :is_not_completed, -> { where(completed: false).or(where(completed: nil)) }
+  scope :is_completed, -> { where(completed: true) }
+  scope :has_due_date, -> { where.not(due_date: nil).order(due_date: :asc) }
+  scope :no_due_date, -> { where(due_date: nil).order(updated_at: :desc) }
 
   def overdue?
     due_date < Time.current if due_date
+  end
+
+  def self.list_ordered
+    Task.is_not_completed.has_due_date +
+      Task.is_not_completed.no_due_date +
+      Task.is_completed.has_due_date +
+      Task.is_completed.no_due_date
   end
 end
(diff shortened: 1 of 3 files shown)


kasugaijin (approved):
I think this works fine! It could be done with SQL, but this is way easier to comprehend and although it makes more db queries, the collections will always be small and I favour readbility over efficiency here. Plus, it would take me ages to figure out how to do it in SQL :D lol

PR #475, opened 2024-02-24
Refactor AdopterApplications routes

mononoken:
# 🔗 Issue
Not for a specific issue. I made these changes while working to refactor the authorization in these controllers for the authorization PR.

# ✍️ Description
I got a little confused with the routing of AdopterApplications, so I refactored it and think these changes fit the Rails conventions better. Specifically, I found `patch "my_application", to: "adopter_applications#update"` tripped me up, since it seems to have been setup as if for a singular resource route, but AdopterApplications are not a singular resource. I also thought it was odd that `:id` was in the strong params for AdopterApplications, which seems to have been done because of how the routing used to be.

I converted buttons that were performing POST and PATCH actions to forms as well. I also adjusted `AdoptablePets#show` to work with the forms better.

# 📷 Screenshots/Demos
This does not change any UI aspect and behavior should remain unchanged.


kasugaijin (approved):
Yes this is more intuitive.

mononoken:
Resolved the merge conflict

PR #492, opened 2024-03-03
Update Seeds for Org Profile Coutnry, State, Phone to fix issues

kasugaijin:
…y State gem so the value populates as expected in the form for Org profile edit. Also use a real phone number so it passes the phonelib gem validation on form save. The existing number was failing second validation.

# 🔗 Issue

# ✍️ Description

Fixes the issue where the country and state are not pre-populated in the form selects on the Org Profile edit page. The reason was because we use the City State gem to load the select and the seed data needs to match the options provided by the gem. 

Also use an actual valid phone number in seeds to prevent the validation error when saving the form in the UI.

# 📷 Screenshots/Demos

### DIFF
--- a/db/seeds/01_alta.rb
+++ b/db/seeds/01_alta.rb
@@ -1,6 +1,6 @@
 orga_location = Location.create!(
-  country: "USA",
-  province_state: "Nevada",
+  country: "US",
+  province_state: "NY",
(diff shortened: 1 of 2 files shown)


nsiwnf (approved):
lgtm

Issue #449, opened 2024-01-25
Foster epic: Create AdopterFosterAccount


kasugaijin:
~~Blocked by https://github.com/rubyforgood/pet-rescue/issues/448~~

~~We differentiate user types in this app by associating `User` with either `AdopterAccount` or `StaffAccount`. We need to add a third user type `FosterAccount`.~~

To add the Foster we will rename the AdopterAccount to AdopterFosterAccount as well as its corresponding table and code references.

Acceptance criteria
- Rename the adopter_accounts table to adopter_foster_accounts
- Rename the model 
- Keep same model associations but update names (belongs_to user, User has one adopter_foster_account)
- Update code references of adopter_account and AdopterAccount
- Tests pass

ErinClaudio:
I have started this but I will have follow-up questions.

kasugaijin:
@ErinClaudio for sure
Note you'll want to wait for this issue https://github.com/rubyforgood/pet-rescue/issues/448 to be merged so we can add the association to the adopter_foster_profile on FosterAccount model.

kasugaijin:
Draft PR in progress: https://github.com/rubyforgood/pet-rescue/pull/452

ErinClaudio:
https://gist.github.com/ErinClaudio/372b5d6e6bb03b3d604ab934a8c563ac

ErinClaudio:
https://gist.github.com/ErinClaudio/372b5d6e6bb03b3d604ab934a8c563ac

PR #497, opened 2024-03-04
Add FosterAccount association to AdopterFosterProfile

ErinClaudio:
# 🔗 Issue

# ✍️ Description

# 📷 Screenshots/Demos


mononoken:
Hey, Erin!
I took a look at your branch. My main comment that I think is affecting this, but not the direct cause of the error, is that this association seems incorrect:
```ruby
# app/models/adopter_foster_profile.rb:59
  belongs_to :foster_account, class_name: “User”, foreign_key: :adopter_account_id
```
I think this should be a simple reference to a `FosterAccount`. If that is not the intention, then it should definitely be named something else since `FosterAccount` is a model in the app.
However, that creates another issue. `belongs_to` associations inherently create presence validations in Rails. That means `AdopterFosterProfile` would require both an `AdopterAccount` **AND** a `FosterAccount`. I think the intention is for one or the other to be required, not both (correct me if I am wrong).
If the intention is for either an adopter_account OR foster_account, I think a [polymorphic association](https://guides.rubyonrails.org/association_basics.html#polymorphic-associations) for an account could work here. With that, each profile would have one associated account, of either foster type or adopter type.
However, if the intention is for each Profile to have a reference to both, then we could add `optional: true` to the foster_account association. That would mean every fosterer would have to be an adopter too though, which I am not sure is intended or not.

ErinClaudio:
@mononoken , thank you for your feedback. After considering your input and reviewing the ticket, I agree that a user should be either an AdopterAccount, FosterAccount, or StaffAccount, not "And." I will proceed with your suggested changes, including creating a polymorphic association. Thank you. @kasugaijin , I'm double-checking with you before moving forward with these adjustments.

ErinClaudio:
@kasugaijin Yup I will go that route

mononoken:
> @ErinClaudio So there are cases where a User could be an Adopter and a Fosterer Can't we just do this?
> 
> ```
> class AdopterFosterProfile < ApplicationRecord
>   belongs_to :location, dependent: :destroy
>   belongs_to :adopter_account
>   **belongs_to :foster_account**
>   accepts_nested_attributes_for :location
>   validates_associated :location
> ```

Since this is the case, we do need to make one or both accounts optional too. `belongs_to` creates a presence validation by default. That means these associations would cause `AdopterFosterProfile` to be invalid when a User only has an `AdopterAccount` (and I think this is probably behind the seed file errors you were getting). 

> ...
> However, if the intention is for each Profile to have a reference to both, then we could add `optional: true` to the foster_account association. That would mean every fosterer would have to be an adopter too though, which I am not sure is intended or not.

I'm thinking then, both accounts should be optional then, and we could write a custom validation to check the presence of at least one account.

kasugaijin:
You know @mononoken was right, polymorphic might be a better suit here. 

https://guides.rubyonrails.org/association_basics.html#polymorphic-associations

```
# Model A
class AdopterAccount < ApplicationRecord
  has_one :adopter_foster_profiles, as: :parent
end

# Model B
class FosterAccount < ApplicationRecord
  has_one :adopter_foster_profiles, as: :parent
end

# Model AdopterFosterProfile
class AdopterFosterProfile < ApplicationRecord
  belongs_to :parent, polymorphic: true
end
```

ErinClaudio:
@mononoken @kasugaijin Thank you

kasugaijin:
@ErinClaudio check out this draft PR I made to help you get the seeds to build - see the video as well for explanation.
https://github.com/rubyforgood/pet-rescue/pull/502

ErinClaudio:
@kasugaijin working on this now with your direction. I will close this PR after I have made the required changes

PR #524, opened 2024-03-10
Refactor matches

mononoken:
# 🔗 Issue
No issue.
Similar to  #475, I thought `Match` could benefit from refactoring. Note that I will be cleaning it up further, getting rid of `same_organization?` and `get_pet_id` in the Authz PR too.

# ✍️ Description
This started with me changing the routes to the Rails convention. I personally found it a little confusing that sometimes we refer to adoptions as "adoptions" and sometimes as a "match". This PR tries to at least make it so that things using `Match` refer to it as a "match". This includes things like the strong params name.

The other portion of this PR works to get `AdopterApplication` logic out of `MatchesController`. I think this business logic belongs in the models. I also like setting withdraw for an individual object more as an instance method rather than as a class method like `self.set_status_to_withdrawn` was doing. Feels more intuitive.

## Using `update!`
One thing I would like to bring attention to and get opinions on, is that I opted to use `update!` instead of `update` for these new methods. I was of two minds on it. The main difference is that `update!` is not silent and will raise an error. That seems undesirable in some cases.

However, where would this failure occur? Currently, `retire_applications` is called in `AdopterApplication#create` after a valid creation is performed. If something was going wrong where the applications could not update following a successful match, I think we would want to know about that. I also think the use cases here are fairly reliable and can't think of a case where I would expect them to fail.

## Using mocks
I would also be curious of opinions on how I wrote the model tests too. I think mocking works well for some of those tests but would love to hear other's thoughts.

### DIFF
--- a/app/controllers/matches_controller.rb
+++ b/app/controllers/matches_controller.rb
@@ -10,20 +10,18 @@ class MatchesController < ApplicationController
     ))
 
     if @match.save
-      set_statuses_to_adoption_made
+      @match.retire_applications
+
       redirect_back_or_to dashboard_index_path, notice: "Pet successfully adopted."
     else
       redirect_back_or_to dashboard_index_path, alert: "Error."
     end
   end
 
-    @successful_application = @match.adopter_account
-      .adopter_applications
-      .where(pet_id: @match.pet_id)[0]
-    AdopterApplication.set_status_to_withdrawn(@successful_application)
-
   def destroy
     if @match.destroy
+      @match.withdraw_application
+
       redirect_to pets_path, notice: "Adoption reverted & application set to 'Withdrawn'"
     else
       redirect_to pets_path, alert: "Failed to revert adoption"
@@ -36,15 +34,6 @@ class MatchesController < ApplicationController
     params.permit(:pet_id, :adopter_account_id)
   end
 
-  # set status on all applications for a pet
-  def set_statuses_to_adoption_made
-    @applications = Pet.find(params[:pet_id]).adopter_applications
-    @applications.each do |app|
-      unless app.status == "withdrawn"
-        app.status = "adoption_made"
-        app.save
-      end
-    end
   def set_pet
     @pet = Pet.find(match_params[:pet_id])
   end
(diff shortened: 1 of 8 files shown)


app/controllers/matches_controller.rb
```diff
@@ -31,25 +30,22 @@ def delete
...
+  def set_pet
+    @pet = Pet.find(match_params[:pet_id])
+  end
+
+  def set_match
+    @match = Match.find(params[:id])
   end
 
   def get_pet_id
     return params[:pet_id] if params[:pet_id]
```
nsiwnf, line 46: I _think_ this should be `match_params[:pet_id]`?
```suggestion
    return match_params[:pet_id] if match_params[:pet_id]
```

app/controllers/matches_controller.rb
```diff
@@ -31,25 +30,22 @@ def delete
...
 
-    Match.find(params[:match_id]).pet_id
+    Match.find(params[:id]).pet_id
```
nsiwnf, line 48: Where is this id coming from again?

nsiwnf (commented):
Running into a few errors locally - some are unrelated to this PR. Are you able to go through various flows on the application_review page?

mononoken:
> Thanks for the cleanup on this! Taking a look now 👀
> 
> Yeah, at some point we thought "Match" made more sense than "Adoption" since we wanted to support "Foster" as well. "Match a pet with an adopter/foster"

Thank you! That makes sense.

> Running into a few errors locally - some are unrelated to this PR. Are you able to go through various flows on the application_review page?
Good catch. Your suggested change on line 46 fixes this I think. I have added that now. I just realized I was not getting this error because I had been running the server off of the authz branch instead of this branch and the authz branch already got rid of `get_pet_id` and `same_organization?`. Woops! Should be working now.

app/controllers/matches_controller.rb
mononoken: The id is the id passed to the `MatchesController` action. Like for example, if you wrote: 
```ruby
<%= button_to "Delete Match", match_path(@match), method: :delete %>
```
The `id` of `@match` would be given to params. It actually would not exist for actions like `create` where there is no existing `@match`, but in those cases it should be getting the `pet_id` from the early return statement. 

I removed this method in the authz branch actually. It's only used in `same_organization?` in this controller, and that logic has been moved to the policy in that branch.

nsiwnf (approved):
lgtm!

PR #525, opened 2024-03-10
Fix adoptable_pet issue and add tests

### LINKED ISSUE #522: Fix: Staff should be able to see and adoptable_pet
**Expected behavior**
As a staff or admin user
when I click on an "adoptable_pet" from the "adoptable_pets" page
I can see that pet's adoptable_pet page

**Current behavior**
As a staff or admin user
when I click on an "adoptable_pet" from the "adoptable_pets" page
I get an internal service error

**Notes** 
I think we just need to skip [this part](https://github.com/rubyforgood/pet-rescue/blob/main/app/controllers/adoptable_pets_controller.rb#L15-L24) when the current_user is staff

```ruby
NoMethodError at /adoptable_pets/3

undefined method `id' for nil:NilClass
          adopter_account_id: current_user.adopter_account.id
```

nsiwnf:
# 🔗 Issue
Fixes #522 

# ✍️ Description
Fixes linked issue and adds integration tests. 
Expectation is that the adoptable_pets views are consistent across users - except for applicants being able to see their own application status 

I don't know how to DRY up these tests - should this be a spec?

### DIFF
--- a/app/controllers/adoptable_pets_controller.rb
+++ b/app/controllers/adoptable_pets_controller.rb
@@ -12,7 +12,7 @@ class AdoptablePetsController < Organizations::BaseController
       redirect_to adoptable_pets_path, alert: "You can only view published pets."
     end
 
-    if current_user
+    if current_user&.adopter_account
       @adoption_application =
         AdopterApplication.find_by(
           pet_id: @pet.id,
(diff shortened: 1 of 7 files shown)


kasugaijin (approved):
LGTM! 
I am personally fine with generic minitest, but someone else might have a preference to handle this differently. We do use another approach in other tests.

kasugaijin:
WE probably don't need to displat the Become an Adopter button, top right, if the user is staff
image

PR #535, opened 2024-03-11
Create foster account3

ErinClaudio:
# 🔗 Issue

# ✍️ Description

# 📷 Screenshots/Demos


ErinClaudio:
@kasugaijin I have more to do but I think this is closer to what you wanted.

db/migrate/20240311173819_remove_profiles_and_add_parent_from_adopter_foster_profiles.rb
```diff
@@ -0,0 +1,8 @@
+class RemoveProfilesAndAddParentFromAdopterFosterProfiles < ActiveRecord::Migration[7.0]
+  def change
+    remove_column :adopter_foster_profiles, :adopter_account_id, :bigint
+    remove_column :adopter_foster_profiles, :foster_account_id, :bigint
```
kasugaijin, line 4: is this removal needed in this particular branch? I had it because I was on your branch where you had already added this column in a migration.

kasugaijin:
yes this looks like it! I think all is left is a switch and replace for any cases of adopter_foster_profile.adopter_account now is adopter_foster_profile.parent

ErinClaudio:
@kasugaijin Some changes made. tests failing. I will be in the meeting tonight

db/migrate/20240311173819_remove_profiles_and_add_parent_from_adopter_foster_profiles.rb
ErinClaudio: I added this back in

kasugaijin:
@ErinClaudio Hmm weird - [my PR has 8 errors](https://github.com/rubyforgood/pet-rescue/actions/runs/8196586683/job/22417120134?pr=502) vs 28 errors in this one so there must be a difference in code that's causing that?

db/migrate/20240312150459_add_back_columns_to_adopter_foster_profiles.rb
```diff
@@ -0,0 +1,6 @@
+class AddBackColumnsToAdopterFosterProfiles < ActiveRecord::Migration[7.0]
```
kasugaijin, line 1: Do we need this migration? adopter_account_id is already on adopter_foster_profiles table. Then we have a migration above to remove both. 

I think it makes more sense to just remove the adopter_account_id from adopter_foster_profiles.

db/migrate/20240311173819_remove_profiles_and_add_parent_from_adopter_foster_profiles.rb
kasugaijin: Ah. Why is that?

db/migrate/20240312150459_add_back_columns_to_adopter_foster_profiles.rb
ErinClaudio: agreed

nsiwnf:
@ErinClaudio let's use the term `caretaker` instead of `parent` since "parent" has a different meaning in programming

ErinClaudio:
I am going to reset this branch and run the migrations without the residual code. I feel like this will leave the code in a more maintainable/cleaner state.

app/models/adopter_foster_account.rb
```diff
@@ -0,0 +1,21 @@
...
+#
+#  index_adopter_foster_accounts_on_user_id  (user_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (user_id => users.id)
+#
+class AdopterFosterAccount < ApplicationRecord
+  belongs_to :user
+  has_one :adopter_foster_profile, as: :caretaker
```
ErinClaudio, line 20: I added this in per this [work](https://github.com/rubyforgood/pet-rescue/blob/2bc73b48781e6b432d059a7d8dc1d7f412b24ea0/db/migrate/20240304183316_add_foster_account_id_to_adopter_foster_profiles.rb) @ben. not sure if it is needed.

ErinClaudio:
Ok, I am happy to continue working on this. But before continuing on I wanted to make sure that this current stopping point is correct. Thank you for the feedback @kasugaijin @mononoken @nsiwnf 

next steps resolving the seeding issue and tests( which we did earlier)

ErinClaudio:
My [notes](https://gist.github.com/ErinClaudio/822992ce207da5e014ae1d08f8b079fe)

app/models/adopter_foster_account.rb
kasugaijin: So we can remove the as caretaker now as we no longer need the polymorphic association. It's a direct association now where

`AdopterFosterAccount` `has_one :adopter_foster_profile`
`AdopterFosterProfile` `belongs_to :adopter_foster_account`

app/models/adopter_foster_profile.rb
```diff
@@ -2,57 +2,63 @@
...
 # Foreign Keys
 #
-#  fk_rails_...  (adopter_account_id => adopter_accounts.id)
+#  fk_rails_...  (adopter_foster_account_id => adopter_foster_accounts.id)
 #  fk_rails_...  (location_id => locations.id)
 #
 class AdopterFosterProfile < ApplicationRecord
   belongs_to :location, dependent: :destroy
   belongs_to :adopter_account
+  belongs_to :caretaker, polymorphic: true, optional: true
```
kasugaijin, line 61: So this now is `belongs_to :adopter_foster_account`

db/migrate/20240314171529_create_adopter_foster_accounts.rb
```diff
@@ -0,0 +1,10 @@
+class CreateAdopterFosterAccounts < ActiveRecord::Migration[7.0]
```
kasugaijin, line 1: Instead of creating a new table, we could just rename the `adopter_accounts` table to `adopter_foster_accounts`

app/models/user.rb
```diff
@@ -58,6 +58,7 @@ class User < ApplicationRecord
 
   has_one :staff_account, dependent: :destroy
   has_one :adopter_account, dependent: :destroy
+  has_one :adopter_foster_profile, dependent: :destroy
```
kasugaijin, line 61: rename line 60 to `adopter_foster_account` and delete line 61

db/schema.rb
```diff
@@ -98,8 +97,14 @@
     t.text "visit_dates"
     t.text "referral_source"
     t.bigint "location_id", null: false
-    t.index ["adopter_account_id"], name: "index_adopter_foster_profiles_on_adopter_account_id"
+    t.string "parent_type"
+    t.bigint "parent_id"
+    t.bigint "adopter_account_id"
+    t.bigint "foster_account_id"
+    t.bigint "adopter_foster_account_id"
```
kasugaijin, line 104: 101, 102 and 103 are no longer required on this table

kasugaijin:
> Ok, I am happy to continue working on this. But before continuing on I wanted to make sure that this current stopping point is correct. Thank you for the feedback @kasugaijin @mononoken @nsiwnf
> 
> next steps resolving the seeding issue and tests( which we did earlier)

It might be the easiest to just start from Main and make the changes. You don't have to, but it might be easier because you have a number of no longer required changes in that PR that you have to undo. Basically, what we need comparing changes to main, is to rename AdopterAccount to AdopterFosterAccount for the model, table, and code references. So a migration to rename the Adopter Accounts table to Adopter Foster Accounts table, and a migration to rename the adopter accounts id column on AdopterFosterProfiles table to adopter foster accounts id. Then a search and replace for any calls to .adopter_account in the code to .adopter_foster_account

nsiwnf:
@ErinClaudio thanks for being flexible with this work - I know we're pivoting a bit and it's not ideal to have many redos. Agree with @kasugaijin's suggestion about starting from main and doing renames

ErinClaudio:
Agreed I will start from a new branch. I will delete this PR and Branch after the new branch is merged.

Issue #521, opened 2024-03-10
Pet Task Epic: When I see the form to create a new task I see a Cancel button that closes the form


kasugaijin:
Currently when you create or edit a task on Pet > Tasks tab, the inline form appears, but there is no way to close the form if you don't want to save the task. 

Acceptance criteria:
- add new button to the right of the 'Save Task' button at the bottom of the inline form that is 'Cancel'
- when you click cancel, the inline form closes (using turbo stream to close the form)
![image](https://github.com/rubyforgood/pet-rescue/assets/95949082/55910781-c8b4-47ce-a084-7e7999b72bad)

Cancel instead of Delete but this style - see styling used in the Files or Photos tab for this button.
<img width="127" alt="image" src="https://github.com/rubyforgood/pet-rescue/assets/95949082/6e3eb3ac-e036-40dd-871e-f2d909412922">

MooseCowBear:
I can do this. Maybe we also want to display validation errors? I think currently the task form doesn't report them to the user.

kasugaijin:
@MooseCowBear 100%!

PR #541, opened 2024-03-12
521 task form cancel button

MooseCowBear:
# 🔗 Issue

#521 

# ✍️ Description

Adds a cancel button to the tasks form that responds with a turbo stream response. 
And adds validation errors message to form. 

So adding a cancel button as a link to the pet show page like `<%= link_to "Cancel", pet_path(task.pet, active_tab: "tasks"), data: {turbo_stream: true} %>` technically resulted in a turbo stream response but the whole page was replaced as you can see: 

Screenshot 2024-03-12 at 3 15 42 PM
Screenshot 2024-03-12 at 3 15 53 PM

In order to get only the relevant task to replace itself, I ended up moving the task portion of the view to its own partial and adding a cancel action to the tasks controller that replaces the form with that partial if the task has an id (so it must be existing) and removes the new task form if the task doesn't. 

The logs for that look like: 

Screenshot 2024-03-12 at 3 36 39 PM

So, significantly better, but more involved. 

# 📷 Screenshots/Demos

Screenshot 2024-03-12 at 2 46 16 PM


config/routes.rb
```diff
@@ -30,6 +30,8 @@
       post "activate", to: "staff#activate"
       post "update_activation", to: "staff#update_activation"
     end
+
+    get "task/cancel", to: "tasks#cancel"
```
kasugaijin, line 34: I believe we can keep the routes together by doing
```
resources :tasks, only: [:new, :create, :edit, :update, :destroy] do
  collection do
    get "cancel", to: "tasks#cancel"
  end
end
```

kasugaijin:
I cannot comment with any degree of nuance on your use of Hotwire, but it makes sense how you handled it and I don't see anything obviously wrong. It looks like it does what it is supposed to do. Maybe someone else with more Hotwire knowledge can comment. I checked out the branch and played with it and it looks good to me. Just one small comment on routes.

config/routes.rb
MooseCowBear: I completely forgot about collection!

nsiwnf:
I'm also not as familiar with hotwire, but if it seems to be working properly, I say we go ahead an merge and then revisit cleanup later

kasugaijin:
> I'm also not as familiar with hotwire, but if it seems to be working properly, I say we go ahead an merge and then revisit cleanup later

Yes my sentiments, too.

kasugaijin (approved):
LGTM!

mononoken:
This is some really clean work, @MooseCowBear !

However, when I had briefly messaged with @kasugaijin about this issue in slack, I thought we could achieve this by just adding one line to the form:
```ruby
    <%= link_to "Cancel", "", class: "classes" %>
```
Are there some benefits to the routes I am not thinking about? I will try to look at this more tonight or tomorrow.

MooseCowBear:
@mononoken I originally thought something like you suggested would work because I had used that approach before too. But it ended up re-rendering basically the whole page instead of just the single task frame. My current Turbo understanding is not good enough to say why.

mononoken:
> @mononoken I originally thought something like you suggested would work because I had used that approach before too. But it ended up re-rendering basically the whole page instead of just the single task frame. My current Turbo understanding is not good enough to say why.

I see that now in your post and replicated that with trying my suggestion above. Good catch!

I like the ingenuity to your solution, but I still am also wary about the routes and associated. It feels wrong to me to add these, when we could have also achieved this cancel toggle with simple JavaScript. 

I also am afraid of what this pattern entails for future development. If we add these routes for a cancel button to work on a form here, we may end up doing that in other spots in the app too. For those reasons, I think if we merge this, we should revisit this if more of this pattern is required in the future. I checked one of my apps that used your original approach for a cancel button, and it does not do this so I suspect something with our implementation is causing a conflict.

I actually am thinking maybe JavaScript is the way to go with a cancel button, and we could use Stimulus for that. For example, what if a user clicks the `Add New Task` button, which loads the form with turbo, clicks `Cancel`, which removes that form, and then clicks `Add New Task` again. Yeah, that's an indecisive user, but it's also requiring loading and offloading the form each time. If we just used JS, we could hide the form on clicking `Cancel` instead so no more loading. What do y'all think? @MooseCowBear @kasugaijin @nsiwnf ? Let me know if I am thinking about this too much 😅

kasugaijin:
> > @mononoken I originally thought something like you suggested would work because I had used that approach before too. But it ended up re-rendering basically the whole page instead of just the single task frame. My current Turbo understanding is not good enough to say why.
> 
> I see that now in your post and replicated that with trying my suggestion above. Good catch!
> 
> I like the ingenuity to your solution, but I still am also wary about the routes and associated. It feels wrong to me to add these, when we could have also achieved this cancel toggle with simple JavaScript.
> 
> I also am afraid of what this pattern entails for future development. If we add these routes for a cancel button to work on a form here, we may end up doing that in other spots in the app too. For those reasons, I think if we merge this, we should revisit this if more of this pattern is required in the future. I checked one of my apps that used your original approach for a cancel button, and it does not do this so I suspect something with our implementation is causing a conflict.
> 
> I actually am thinking maybe JavaScript is the way to go with a cancel button, and we could use Stimulus for that. For example, what if a user clicks the `Add New Task` button, which loads the form with turbo, clicks `Cancel`, which removes that form, and then clicks `Add New Task` again. Yeah, that's an indecisive user, but it's also requiring loading and offloading the form each time. If we just used JS, we could hide the form on clicking `Cancel` instead so no more loading. What do y'all think? @MooseCowBear @kasugaijin @nsiwnf ? Let me know if I am thinking about this too much 😅

I think if a simpler solution will solve an issue then it should be what we choose, generally. 

If we use JS to hide the form, when the user has 'hidden' the form and clicks Create Task again, it will load a new chunk of HTML via hotwire for the new form, right? Play this a few times and it would result in a load of redundant, hidden HTML. Granted, it is not a likely use case or a huge issue, but hiding isn't perfect from that sense, either. I think you raise a good point with needing a route. I have not worked with hotwire enough to comment. My sentiments are we keep momentum on things. Technical debt is ideally not incurred, but factoring in developer happiness, it is also a pain to implement entirely new approaches. I would defer to @MooseCowBear on whether or not you are happy with your implementation or think JS is better? I am happy as it is.

MooseCowBear:
@kasugaijin @mononoken As a general rule, I'm never opposed to looking for better solutions. Which actually might be a reason for me not to make calls like this when we are trying to get an MVP going...

I think in this case, though, it is at least not obvious to me that using JS will be a solution to the debt incurred by the route. Specifically, I'm wondering about how it would work for updating existing tasks. We'd need to remove the form from the page (or hide it, and then the issue Ben raised occurs), but if the form is replacing the task, we'd also need to re-add the task to the page if the cancel button is clicked -- which I think would involve making a request to the backend to get the task/task partial anyway and then we'd need to figure out which existing route to get that from or add a new route which puts us in a situation like the one we are in. 

Alternatively, we could have the page draw a hidden form for every task when it loads (plus one for the new task) and use a Stimulus controller to toggle the visibility of the task/form pair (edit/new to make the form visible and hide the relevant task if there is one, and cancel to hide the form/make the task visible if there is one). So this would essentially give up on turbo-frames. I suppose the response could still be a turbo-stream, only it would replace both the task and the form to make it hidden again and clear the fields (or add the task in the case of creating)? I'm not sure what would happen to the form in the case of a 422 response if we went this way. We might need a turbo-stream response to handle that and show the validation errors instead of the usual `render :new` or `render :edit`. (Or maybe there would be a way to leave the turbo-frames and not have them interfere? Without actually trying it I'm not sure how the two would interact).

Because I am a curious person (perhaps to a fault), I'm happy to create a few branches and investigate. But if there are other things that should be given priority right now, I can spend my time on those and come back to this at a later point.

kasugaijin:
@MooseCowBear thats totally fine. I don’t want to put you on the spot! Just wanted to give you the option rather than make the call without you having a say. In this case, I think we should move on. Your current solution is fine, and we have a lot more work to complete, and I am of the opinion that momentum is key. We can always dedicate time to this particular technical question in a meeting and determine a preferred approach in future. There will be other places we will want similar functionality to cancel a form, but not a whole lot of cases. Maybe one or two more areas in the app.

What do you think @mononoken?

mononoken:
I agree with accepting this solution. I wanted to voice my concerns to see what y'all think, but I also was questioning if I should even say something when it's still a great solution. Thanks for bringing up the issues with my proposed solution as well, and thanks y'all for the discussion!

kasugaijin:
> @MooseCowBear @kasugaijin I agree with accepting this solution. I wanted to voice my concerns to see what y'all think, but I also was questioning if I should even say something when it's still a great solution. Thanks for bringing up the issues with my proposed solution as well, and thanks y'all for the discussion!

Thanks @mononoken please keep doing it. I sometimes act in haste and it’s good to have alternative opinions and discussions.

jmilljr24:
If you end up coming back to this, I believe the issue is that PetsController#show doesn't have a respond to format.turbo_stream so it defaults to html.

```  
def show
    @active_tab = determine_active_tab
    return if pet_in_same_organization?(@pet.organization_id)

    redirect_to pets_path, alert: "This pet is not in your organization."
  end
```

I was confused as well playing with this because I was focused on the task controller but you are actually requesting the turbo stream from the pets controller with `<%= link_to "Cancel", pet_path(task.pet, active_tab: "tasks"), data: {turbo_stream: true} %>`

This is a crude example that just updates the tasks_list with turbo when you click cancel. You could add a dom_id to the individual tasks if you wanted to target the specific task your are editing/cancelling. (I did just the tasks_list for proof of concept)

```
  def show
    @active_tab = determine_active_tab
    if @active_tab == "tasks"
      respond_to do |format|
        format.html
        format.turbo_stream { render turbo_stream: turbo_stream.replace("tasks_list", partial: "organizations/pets/tasks/tasks", locals: {task: @task}) }
      end
    end
    return if pet_in_same_organization?(@pet.organization_id)

    redirect_to pets_path, alert: "This pet is not in your organization."
  end
```

This would remove the need for the new route and cancel method.

kasugaijin:
Ah nice find @jmilljr24 . If you want to make a PR for it, please go ahead. It will be a good learning opportunity for a few of us I think :)

PR #551, opened 2024-03-15
Create AdopterFosterAccount

ErinClaudio:
# 🔗 Issue

# ✍️ Description

# 📷 Screenshots/Demos


ErinClaudio:
@kasugaijin 

Please review and let me know if this approach seems appropriate.

This adjustment involved only altering the model name, so I updated the wording in the model files accordingly while keeping the relationships unchanged.

Planned Steps:

1. Rename AdopterAccount to AdopterFosterAccount across the entire application.
2. This renaming will necessitate a series of migrations. Execute these migrations and address any resulting errors.
3. Ideally, all tests should pass after these changes. If not, we will tackle the failing tests next.

Let me know your thoughts.

ErinClaudio:
[notes](https://gist.github.com/ErinClaudio/91edbc855c11d8ebfc1da415ab12c3fd)

db/migrate/20240315160350_rename_adopter_accounts_to_adopter_foster_accounts.rb
```diff
@@ -0,0 +1,7 @@
+class RenameAdopterAccountsToAdopterFosterAccounts < ActiveRecord::Migration[7.0]
+  def change
+    if table_exists?(:adopter_account)
+      rename_table :adopter_account, :adopter_account
```
kasugaijin, line 4: table names should be plural I believe and the second `adopter_account` needs to be `adopter_foster_accounts`

db/schema.rb
```diff
@@ -61,8 +54,14 @@
     t.index ["pet_id"], name: "index_adopter_applications_on_pet_id"
   end
 
+  create_table "adopter_accounts", force: :cascade do |t|
```
kasugaijin, line 57: see, this table has the wrong name, should be adopter_foster_accounts

db/schema.rb
```diff
@@ -98,8 +97,14 @@
     t.text "visit_dates"
     t.text "referral_source"
     t.bigint "location_id", null: false
+    t.string "parent_type"
```
kasugaijin, line 100: This stuff from the old PR using polymorphic association should not be here. I think this is because you are using the same database from your old PR and it is including this in the schema. I suggest dropping and recreating the table so it builds only with the migrations you want it to for this PR.

db/schema.rb
```diff
@@ -119,6 +124,13 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "foster_accounts", force: :cascade do |t|
```
kasugaijin, line 127: Same as above. Do not want this on the schema.

db/schema.rb
```diff
@@ -258,13 +270,14 @@
 
   add_foreign_key "active_storage_attachments", "active_storage_blobs", column: "blob_id"
   add_foreign_key "active_storage_variant_records", "active_storage_blobs", column: "blob_id"
-  add_foreign_key "adopter_accounts", "users"
-  add_foreign_key "adopter_applications", "adopter_accounts"
+  add_foreign_key "adopter_applications", "adopter_accounts", column: "adopter_account_id"
```
kasugaijin, line 273: Down here, need to make sure adopter_accounts becomes adopter_foster_accounts sand remove references to foster_accounts

app/models/adopter_account.rb
```diff
@@ -1,23 +0,0 @@
-# == Schema Information
```
kasugaijin, line 1: Need to make the model/file replacing this - adopter_foster_accounts

kasugaijin:
Yes your steps seem appropriate. I have left a series of comments. One thing, while you have started a new branch, it looks like you have not reset your database and you have committed schema changes from the old PR where we were going to use polymorphic association. Note that your database, unless you create a new one for a separate branch, or recreate it entirely, will still have the schema from whatever migrations you have run on it. In this case, it has the changes from the migrations in the old PR. So please drop and recreate your db, paying a mind to which migrations you have in this branch, and ensure you only have the migrations from main branch + whatever you need for this PR. That will re-write the schema and remove the unwanted changes we have above. Thanks!

db/schema.rb
ErinClaudio: I had to reset it because of making commits to the wrong branch. Monday - I think this cleaner now

ErinClaudio:
@kasugaijin I hope I've left this ticket in a decent spot for you. The schema should now reflect the correct changes. The next steps involve updating the seed file and modifying all instances of the adopter_accounts.

kasugaijin:
> @kasugaijin I hope I've left this ticket in a decent spot for you. The schema should now reflect the correct changes. The next steps involve updating the seed file and modifying all instances of the adopter_accounts.

Thanks @ErinClaudio! You did most of the work. Just two quick commits after.

nsiwnf:
Great job! Would it be possible to combine all the migrations into one migration file? I think we can also rename columns which would help retain data

kasugaijin:
> Great job! Would it be possible to combine all the migrations into one migration file? I think we can also rename columns which would help retain data

Done. I split the table rename and column renames into two files. I also added foreign key constraints for the belong_to relationships.

app/policies/adopter_foster_profile_policy.rb
mononoken, line 1: I think we should update the permissions in `authorizable.rb` to reflect the name change from `AdopterProfile` to `AdopterFosterProfile`. That way we are consistent with naming everywhere.

You should be able to just go into `app/models/concerns/authorizable.rb` and change every reference of `adopter_profile` to `adopter_foster_profile`.

Then we can do the same in this file, `app/policies/adopter_foster_profile_policy.rb` too.

app/policies/profile_review_policy.rb
mononoken, line 1: Also update `view_adopter_profiles` permission name to `view_adopter_foster_profiles` here too, per https://github.com/rubyforgood/pet-rescue/pull/551/files#r1533320405

mononoken (commented):
lgtm! I think we should update the authorization permission names to reflect the `AdopterProfile` to `AdopterFosterProfile` name change, but I can do that in a separate PR.

kasugaijin:
> lgtm! I think we should update the authorization permission names to reflect the `AdopterProfile` to `AdopterFosterProfile` name change, but I can do that in a separate PR once this is merged.
> 
> Thank you for all the work, Erin!

Good point! This branch needs main merging in and then running find and replace again. I can do that.

kasugaijin:
@mononoken I made the change to add unique constraint on index for adopter foster profile on adopter foster account. 
With find and replace, I only found one filename to change to the new naming. Am I missing something else that needs changing?

FYI I am finishing this off for Erin as he is away for 10 days - @ErinClaudio is aware I was taking over :)

mononoken:
> @mononoken I made changes to policy permission names (commit below) and added unique constraint to the index as suggested.
> 
> FYI I am finishing this off for Erin as he is away for 10 days - @ErinClaudio is aware I was taking over :)


Thanks for taking over, @kasugaijin ! The new changes look good.

kasugaijin:
> > @mononoken I made changes to policy permission names (commit below) and added unique constraint to the index as suggested.
> > FYI I am finishing this off for Erin as he is away for 10 days - @ErinClaudio is aware I was taking over :)
> 
> Thanks for taking over, @kasugaijin ! The new changes look good.

@mononoken feel free to approve and merge if you want to!

mononoken (approved):

kasugaijin:
Nice! This unblocks a lot of work!

PR #563, opened 2024-03-18
Refactor pet task form cancel

jmilljr24:
This is in response to discussions on PR #541. If these changes are not worth another merge feel free to disregard.

# ✍️ Description

1. Refactor the task form cancel button. Removes the need for additional routing for a cancel action.
2. Fixes a bug where if a form error is displayed, the cancel button would not work.

Minimal changes were made to the work @MooseCowBear did. I removed the cancel action from the TasksController and associated routing. I added a turbo_stream response to the PetsController and copied the code from cancel.turbo_stream.erb to a new file show.turbo_stream.erb.

The original PR has a bug with the error message displayed due to the turbo stream removing the turbo_frame_tag the cancel button would target. I resolved this by adding an additional turbo_frame_tag to the form partial. Not the cleanest but it works with no issues. One potential better implementation I thought of was for the error message to display as a flash which matches how it's implemented elsewhere in the website. I feel that's outside the scope of this PR. 

# 📷 Screenshots/Demos

Turbo_stream response
[Screencast from 03-18-2024 02:20:01 PM.webm](https://github.com/rubyforgood/pet-rescue/assets/16829344/96eb7fa6-387c-4727-9a8e-7ad70a3f6eb7)

Cancel works for new and edit form. Cancel works after error displayed.
[Screencast from 03-18-2024 01:58:34 PM.webm](https://github.com/rubyforgood/pet-rescue/assets/16829344/3fb58cf6-8b2e-425d-8e23-301c6304b05f)


kasugaijin:
@mononoken @MooseCowBear what do you think of @jmilljr24 implementation?
To my untrained hotwire eye, no longer requiring the new route is preferable and this appears more in line with the hotwire way?

mononoken:
Thank you for looking into this @jmilljr24 !

I do like this approach better. I was actually running into an issue last night while we merged the Action Policy PR with the cancel button via routes approach. The route exposed the action and thus it needed to be authorized. But, because the cancel route was acting as a kind of middle-route between either a new **or** edit form, the policy logic would require conditionals and maybe some more thought.  This approach fixes that issue.

This PR made me notice some other things going on with how Turbo and Tasks were setup originally in the project. I think a further improvement we could consider, would be to move the task partials in the pets views, `_task` and `_tasks`, to the tasks views, as `show` and `index`, and use turbo to stream those frames from there. That would avoid the task logic leaking into the `Organizations::PetsController` since we could move it into the `TasksController`. So for example, we would have:
```erb
# app/views/organizations/pets/tabs/_tasks.html.rb
<%= turbo_frame_tag dom_id(@pet, :tasks), src: pet_tasks_path(@pet) %>

# app/views/organizations/tasks/index.html.erb
<%= turbo_frame_tag dom_id(@pet, :tasks), class: "css-classes" do %>
  # stuff inside the turbo frame currently in app/views/organizations/pets/tasks/_tasks.html.rb
<% end %>
```
My gut tells me everything in `views/organizations/pets/tasks/` really should be in `views/organizations/tasks/` instead. But that is all seeking clean code. Everything works as is in this PR. Anyone have thoughts on this suggestion? @jmilljr24 @kasugaijin @MooseCowBear

jmilljr24:
I have a working solution using @mononoken 's suggestions. Added the index/show to TasksController, along with reorganizing the files/folders as mentioned. I can push it up if that's the agreed decision. The only major thing that I have no idea on is the authorization for the new actions. I did some copy and paste to make it work but there is a high probability it's wrong.

Overall I'm happy with this change and it seems more inline with convention. One minor thing to note (I'm not sure how to fix) is that with this change to the TasksController, you can see the delay it takes for the task lists to render after switching tabs on the pets page.
[Screencast from 03-19-2024 07:34:52 AM.webm](https://github.com/rubyforgood/pet-rescue/assets/16829344/fcd67f25-efbd-4275-bb4d-a8949a5ccb8f)

It has something to do with making two request. First the clicking on the tab does a full page render then the task list request comes through.
Screenshot from 2024-03-19 07-46-45

kasugaijin:
> I have a working solution using @mononoken 's suggestions. Added the index/show to TasksController, along with reorganizing the files/folders as mentioned. I can push it up if that's the agreed decision. The only major thing that I have no idea on is the authorization for the new actions. I did some copy and paste to make it work but there is a high probability it's wrong.
> 
> Overall I'm happy with this change and it seems more inline with convention. One minor thing to note (I'm not sure how to fix) is that with this change to the TasksController, you can see the delay it takes for the task lists to render after switching tabs on the pets page. [Screencast from 03-19-2024 07:34:52 AM.webm](https://github.com/rubyforgood/pet-rescue/assets/16829344/fcd67f25-efbd-4275-bb4d-a8949a5ccb8f)

Push it up and we will have a look!

mononoken:
Thank you for the extra work, @jmilljr24 ! I will take a look at it tonight and check out the authz part too.

config/routes.rb
```diff
@@ -14,11 +14,7 @@
 
     resources :home, only: [:index]
     resources :pets do
-      resources :tasks, only: [:new, :create, :edit, :update, :destroy] do
-        collection do
-          get "cancel", to: "tasks#cancel"
-        end
-      end
+      resources :tasks, only: [:new, :index, :show, :create, :edit, :update, :destroy]
```
mononoken, line 17: We can get rid of the only now as these are all the default actions.

app/views/organizations/tasks/index.html.erb
```diff
@@ -23,13 +23,11 @@
           <%= turbo_frame_tag Task.new %>
         </div>
         <ul class="list-group">
-          <% @pet.tasks.list_ordered.each do |task| %>
-            <%= turbo_frame_tag task do %>
-              <%= render partial: "organizations/pets/tasks/task", locals: { task: task } %>
-            <% end %>
+          <% @tasks.each do |task| %>
+            <%= render partial: "organizations/tasks/task", locals: { task: task } %>
           <% end %>
```
mononoken, line 28: Beautiful little part of Rails is that you can actually do this with just:
```ruby
<%= render @tasks.list_ordered %>
```

app/controllers/organizations/pets_controller.rb
```diff
@@ -21,6 +21,13 @@ def edit
 
   def show
     @active_tab = determine_active_tab
+    return unless @active_tab == "tasks"
+
+    @task = params[:task_id] ? @pet.tasks.find(params[:task_id]) : @pet.tasks.build
+    respond_to do |format|
+      format.html
+      format.turbo_stream
+    end
```
mononoken, line 30: I think we can get rid of this task logic from the PetsController now. Let me know if I am forgetting something.

app/controllers/organizations/tasks_controller.rb
```diff
@@ -1,28 +1,39 @@
...
+  before_action :set_task, only: %i[show edit update destroy]
 
   def new
     authorize! Task, context: {pet: @pet}
 
     @task = @pet.tasks.build
   end
 
+  def index
+    authorize! Pet, context: {organization: Current.organization}
```
mononoken, line 12: This was close to correct, but with this setup it uses the PetPolicy, not the TaskPolicy. Update this to:
```ruby
authorize! Task, context: {pet: @pet}
```
Action Policy infers the policy based on the class of the first parameter object given to `authorize!`, so now this will grab the correct policy.

The policy was not setup for `index` though since it did not exist so please also update this:
```ruby
# app/policies/organizations/task_policy.rb
  alias_rule :create?, :new?, :index?, to: :manage? # Add `:index?` to line 8 like shown
```
This small change will use the default `manage?` check for index?, which will work fine for this.

app/controllers/organizations/pets_controller.rb
mononoken: Hmm, hold off on this comment. I think I may have different thoughts now.

I meant to get more review progress on this tonight, but I ended up falling a little ill after the meeting. Hopefully, I will feel better tomorrow and finish it up then. I think I may have a fix for the delay you noticed.

jmilljr24: I originally did remove it but it must have slipped back in with a merge on accident. As of right now it does nothing but I'm curious on you fix for the visual delay.

jmilljr24: For some reason I thought I removed this but I guess not. I will have to rework a few things. The edit form cancel is all going through the tasks_controller but the new form cancel was still relying on this.

I'm looking forward to your thoughts on the delay.

app/controllers/organizations/tasks_controller.rb
```diff
@@ -40,10 +58,10 @@ def update
           Organizations::TaskService.new(@task).create_next
         end
 
-        format.turbo_stream { render turbo_stream: turbo_stream.replace("tasks_list", partial: "organizations/pets/tasks/tasks", locals: {task: @task}) }
+        format.turbo_stream { render turbo_stream: turbo_stream.replace("tasks_pet_#{@pet.id}", partial: "organizations/pets/tabs/tasks", locals: {task: @task}) }
```
mononoken, line 61: We want to broadcast to the specific task, not the whole tab. Instead of this, I think you could remove all that and have this:
```ruby
# app/views/organizations/tasks/update.turbo_stream.erb
<%= turbo_stream.replace @task do %>
  <%= render @task %>
<% end %>
```
Now, you should be able to open multiple edits on the index, and click "Update Task" on one, and the rest will stay open.

app/views/organizations/tasks/show.turbo_stream.erb
```diff
@@ -0,0 +1,3 @@
+<%= turbo_stream.update @task do %>
```
mononoken, line 1: Let's also create a `show.html.erb` to support the html response path as well. I think it can just be:
```erb
<%= render @task %>
```

app/controllers/organizations/tasks_controller.rb
jmilljr24: I was going off the prior implementation and only updated the target to use the new dom id. Happy to change this, the only consideration is that the task list will no longer update to the correct order until a refresh.

For example, the task list order is overdue, upcoming then completed. If you change the first overdue task to a future date it will still be the first on the list until a refresh. I can see pro's and cons to both ways.

app/controllers/organizations/pets_controller.rb
mononoken: Last night, I was thinking the delay behavior must have to do with lazy loading or instant links. Poking around, I haven't quite figured out how to eager load the tasks though with the tab. However, I'm not sure we really have to worry about it. Right now, your solution is just loading the tasks in a separate request from the tab 'template', which is fine since I don't see any n+1 issues.

app/controllers/organizations/tasks_controller.rb
```diff
@@ -1,28 +1,46 @@
...
   def new
     authorize! Task, context: {pet: @pet}
 
     @task = @pet.tasks.build
   end
 
+  def index
+    authorize! Task, context: {pet: @pet}
+
+    if params[:cancel] == "true"
```
mononoken, line 14: Thanks for moving this over. Good stuff!

app/controllers/organizations/tasks_controller.rb
```diff
@@ -1,28 +1,46 @@
...
+  before_action :set_task, only: %i[show edit update destroy]
 
   def new
     authorize! Task, context: {pet: @pet}
 
     @task = @pet.tasks.build
   end
 
+  def index
+    authorize! Task, context: {pet: @pet}
```
mononoken, line 12: 👍

app/controllers/organizations/pets_controller.rb
jmilljr24: > Last night, I was thinking the delay behavior must have to do with lazy loading or instant links. Poking around, I haven't quite figured out how to eager load the tasks though with the tab. However, I'm not sure we really have to worry about it. Right now, your solution is just loading the tasks in a separate request from the tab 'template', which is fine since I don't see any n+1 issues.

My first thought was lazy/eager load but couldn't figure anything obvious out. I agree I don't thing the delay is a huge issue right now. I was thinking maybe in a future revision each tab would be updated with turbo instead of the whole page re-rendering, which I think would possibly help (or maybe not haha). Just a thought.

app/controllers/organizations/tasks_controller.rb
```diff
@@ -1,28 +1,46 @@
...
+
   def create
     authorize! Task, context: {pet: @pet}
 
     @task = @pet.tasks.build(task_params)
 
     if @task.save
       respond_to do |format|
-        format.turbo_stream { render turbo_stream: turbo_stream.replace("tasks_list", partial: "organizations/pets/tasks/tasks", locals: {task: @task}) }
+        format.turbo_stream { render turbo_stream: turbo_stream.replace("tasks_pet_#{@pet.id}", partial: "organizations/pets/tabs/tasks", locals: {task: @task}) }
```
mononoken, line 38: Similar to `update`, I think we should be targeting the index, not the tab.

However, this made me realize I misguided you in my initial recommendation to move the stuff from the tabs to the task/index. Moving all the tasks stuff to the tasks views was the right move, but we should not have moved the entirety of `tabs/_tasks.html.erb` to `tasks/index.html/erb`. A lot of that html is specific to the tabs, and does not make sense for the tasks/index standing alone. And really, the turbo_frame for the pet tasks should encapsulate the rendered tasks only, not the whole template.

So instead, I think the better solution is this:
```erb
# move content from tasks/index.html.erb back to tabs/tasks
# and move the outermost turbo tags to only surround the tasks render

# app/views/organizations/pets/tabs/_tasks.html.erb
<div class="row">
  <div class="col-12 mb-4">
  ...
        <%= turbo_frame_tag dom_id(@pet, :tasks) do %>
          <%= render @tasks %>
        <% end %>
        ...
</div>

# app/views/organizations/tasks/index.html.erb
<%= turbo_frame_tag dom_id(@pet, :tasks) do %>
  <%= render @tasks %>
<% end %>

```
This will require `@tasks` to be defined in `PetsController#show`, which will also assist us in eager loading the tasks for the tab.
```ruby
# app/controllers/organizations/pets_controller.rb
  def show
    @active_tab = determine_active_tab
    @tasks = @pet.tasks
  end
```
And finally:
```erb
# app/views/organizations/tasks/create.turbo_stream.erb

<%= turbo_stream.update Task.new, "" %> 
<%= turbo_stream.append dom_id(@pet, :tasks), @task %> 
```
You can test things are working properly by opening an edit for one of the tasks, then open the new task button, and submit a new task. It should be appended at the end of the task list, and the edit form should remain open as well.

Let me know if you have any issues with this solution.

app/controllers/organizations/tasks_controller.rb
mononoken: Thanks, @jmilljr24 ! I did not consider that ramification. 

You are right, as I am getting ambitious past the original intent of the PR... I am a new collaborator and will try to work on focusing on scope on my future reviews. Sorry!

I think we can leave this as is then based on your points, unless you have already made my suggested changes. If we want to clean this up more, we can do so in a future PR.

jmilljr24: I haven't touched anything yet. Love your insight thought and happy to do whatever is decided on!! If you don't mind, could you please indicate on the other revisions requested so there is no ambiguity on what you still want changed since the conversations are starting to overlap.

app/controllers/organizations/tasks_controller.rb
mononoken: Ignore this. See my reply to your comment on `update` https://github.com/rubyforgood/pet-rescue/pull/563/files#r1533089025

Again, sorry for getting ambitious in this review and missing the sorting issue!

mononoken:
@jmilljr24 I marked the comments that no longer are applicable resolved. We can revisit those in a separate PR if we decide to change up how turbo is wired up here. 

Let's still add the html file for show to keep that no-JS path working. (https://github.com/rubyforgood/pet-rescue/pull/563#discussion_r1533052383)

Thanks again for your time and the clean work on this one!

test/system/tasks_test.rb
```diff
@@ -9,7 +9,7 @@ class TasksTest < ApplicationSystemTestCase
   end
 
   test "creates a recurring task with a due date without redirecting" do
-    due_date = (Date.today + 1.day)
+    due_date = (Date.today + 3.day)
```
mononoken, line 12: Think you have a change for a separate issue leaking in here.

mononoken (approved):
Looks good, @jmilljr24 ! Thanks for the clean work on this one.

Issue #556, opened 2024-03-18
Fix: All adopter applications should be visible to staff


nsiwnf:
**Actual**
On the `adopter_applications_review` page, clicking "New Adoption" causes the applications associated with that pet to disappear from the page. Further, they are not included in filter results for pet name or application status 'adopted'. 

**Expected**
As staff, when I click "New Adoption" for an application on the `adopter_applications_review` page, I should still be able to see all the applications for that pet.

**Notes**
Might need to fix ransack query

MooseCowBear:
I can work on this

PR #571, opened 2024-03-20
556 all adopter applications visible to staff

MooseCowBear:
# 🔗 Issue

#556 

# ✍️ Description

Changes the `self.org_pets_with_apps` method to include pets that have been adopted so they show up on the adoption applications review index page and adds tests for the updated method. 

Right now the order of the pets changes after the adoption has been made. In the recording below you can see that Jasmine moved the bottom of the page after the adoption was made, do we want to specify an ordering for the pets? 

# 📷 Screenshots/Demos

https://github.com/rubyforgood/pet-rescue/assets/82609260/279e7a22-c2c4-4762-8380-4cf6867a4642


test/models/pet_test.rb
```diff
@@ -39,4 +39,22 @@ class PetTest < ActiveSupport::TestCase
...
+      pet_with_app = create(:pet, :adoption_pending)
+      pet_without_app = create(:pet)
+      res = Pet.org_pets_with_apps(ActsAsTenant.current_tenant.id)
+
+      assert res.include?(pet_with_app)
+      assert_not res.include?(pet_without_app)
+    end
+
+    should "include pets that have been adopted" do
+      adopted_pet = create(:pet, :adopted, :adoption_pending)
```
kasugaijin, line 54: technically the status here would be `successful_applicant` for an adopted pet. But there is no enforcement of this...something we should address down the road.

test/system/adoption_creation_test.rb
```diff
@@ -0,0 +1,22 @@
...
+  setup do
+    user = create(:staff)
+    @pet = create(:pet)
+    create(:adopter_application, :successful_applicant, pet_id: @pet.id)
+
+    sign_in user
+  end
+
+  context "creating an adoption" do
+    should "should display adopeted pet's applications after it has been adopted" do
```
kasugaijin, line 13: typo in test name `adopeted`

kasugaijin (commented):
Looks good! Just a couple comments.
 I remember now, the applications page in the original app does not show adopted pets. I think down the road we might want to rethink how we differentiate between adopter and unadopted pets. Perhaps we would want to not show adopted pets by default, and only when that state is selected in the search. Over time an org will have hundreds+ of adopted pets. But this is for future planning.

MooseCowBear:
@kasugaijin I fixed the tests. 

As for differentiating adopted and not yet adopted, I can think of at least a couple ways to do it. We could do it in the db query, something like: `Pet.org_pets_with_apps(old_id).where.missing(:match)` for not yet adopted and `where.associated(:match)` for  adopted. Or we could stick with the db query for all pets with applications and then add a default filtering to  the results in the controller action if there's no specified application status. Just some preliminary thoughts.

kasugaijin:
@MooseCowBear thank you! Yes those two options would work. I think it depends on how big the collections are, and we might consider caching also once things start to get large. TBD.

kasugaijin (approved):

Issue #573, opened 2024-03-21
Repair hero background image


Developer3027:
The hero section for org/root no longer shows the intended image. This was originally just the URL link from the Figma file. This link has an expire code built in. Add the intended image to the local environment, (assets/images) and rework code the Rails way for future image grab from the db.

Developer3027:
I have reworked using the asset_path helper. Dog-run.png from the Figma file is added to the assets/images folder and with using the asset_path helper the image is evaluated and supplied by rails.

PR #574, opened 2024-03-21
573 hero background fix

Developer3027:
# 🔗 Issue
The background image for the hero section uses a Figma location link which has expired. We need a way to allow org's to supply their own image anyway.
[issue 573](https://github.com/rubyforgood/pet-rescue/issues/573)

# ✍️ Description
Add intended default image and rework code to allow for future rails evaluation.

# 📷 Screenshots/Demos

Screenshot from 2024-03-20 21-43-25


kasugaijin (approved):
Yes this is a good placeholder. I was missing that hero image on the prod site :)

Issue #534, opened 2024-03-11
Remove 422 route as "restricted access"


mononoken:
Currently, this route is defined:
```ruby
# routes.rb
match "/422", to: "errors#restricted_access", via: :all
```

I am not sure where this definition came from but `422` should be "Unprocessable Content" (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422).

I think this could have potentially been a typo for `402` which is "Payment required" which seems closer to "restricted access". Ultimately, though "restricted access" makes me think of `403 forbidden`.

All that said, I don't think this route is currently in use. We could delete it (and associated view and controller action).

 Anyone know what this was for and whether we are good to remove?

kasugaijin:
I do not recall why I added that specific route...and incorrectly named haha. I think we will want to figure out what routes/errors we want to handle moving ahead and refactor accordingly. Always nicer to show custom, branded errors rather than the heroku default.

PR #575, opened 2024-03-21
Change 422 to correct status name, Unprocessable Content

mononoken:
# 🔗 Issue
#534 

# ✍️ Description
This error code seems like it was incorrectly named at some point.

https://developer.mozilla.org/en-US/docs/web/http/status/422


kasugaijin (approved):
Good catch!

Issue #555, opened 2024-03-18
Add validations around adopter applications


nsiwnf:
Currently our seed files allows for some undesired scenarios around adoptions and adoption applications. We need some validations checking for uniqueness to prevent this from occurring. 

**Acceptance Criteria**
- an `adopter` to can have more than one adoption applications but can only have **1 application per `pet`**
- a `pet` can only have 1 `match` (multiple adopters cannot "adopt" the same pet)

PR #578, opened 2024-03-21
555 adopter app validations

mononoken:
# 🔗 Issue
#555 

# ✍️ Description

Added validations to model and to schema so that the Rails app checks for uniqueness of this relationship.

> a pet can only have 1 match (multiple adopters cannot "adopt" the same pet)

Added uniqueness validation to model and schema for pets on matches.

I also updated the seed file, so that it will redo invalid adopter application sampling attempts.

I did not make any UI changes, as currently adopter applications disappear from the index view when an adoption is made.

### DIFF
--- a/db/migrate/20240322214921_index_adopter_applications_on_account_and_pet.rb
+++ b/db/migrate/20240322214921_index_adopter_applications_on_account_and_pet.rb
@@ -1,0 +1,7 @@
+class IndexAdopterApplicationsOnAccountAndPet < ActiveRecord::Migration[7.0]
+  def change
+    add_index :adopter_applications, [:pet_id, :adopter_foster_account_id],
+      unique: true,
+      name: "index_adopter_applications_on_account_and_pet"
+  end
+end
--- a/db/seeds/01_alta.rb
+++ b/db/seeds/01_alta.rb
@@ -261,13 +261,19 @@ ActsAsTenant.with_tenant(@organization) do
   )
 
   10.times do
-    AdopterApplication.create!(
+    adopter_application = AdopterApplication.new(
       notes: Faker::Lorem.paragraph,
       profile_show: true,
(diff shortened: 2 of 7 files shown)


db/seeds/01_alta.rb
```diff
@@ -261,13 +261,19 @@
...
+      status: rand(0..4),
+      adopter_account: AdopterAccount.all.sample,
       pet: Pet.all.sample
     )
+
+    if adopter_application.valid?
+      adopter_application.save!
+    else
+      redo
+    end
```
mononoken, line 276: I think this method is fine for ensuring 10 adopter applications are generated at least, but it does have the one potential problem where if it's not possible for 10 unique applications to exist, it will create an infinite loop. Another easy solution would be to just try 10 times and skip the invalid ones. Could also use a flag to only retry an iteration one time. I think this current one is fine though for a seed file.

db/migrate/20240322214921_index_adopter_applications_on_account_and_pet.rb
```diff
@@ -0,0 +1,7 @@
+class IndexAdopterApplicationsOnAccountAndPet < ActiveRecord::Migration[7.0]
+  def change
+    add_index :adopter_applications, [:pet_id, :adopter_foster_account_id],
+      unique: true,
+      name: "index_adopter_applications_on_account_and_pet"
```
mononoken, line 5: I had to manually name the index because, with the new adopter_foster_account name, this default index name was getting too long.

db/seeds/01_alta.rb
nsiwnf: Yeah, Im in favor of avoiding infinite loops but agree this is fine for now

nsiwnf (approved):
lgtm, ty!

PR #585, opened 2024-03-22
508 Add recurring to Default Pet Tasks

### LINKED ISSUE #508: Pet Task Epic: As staff I can create a default pet task that is recurring and set the
Staff can currently create Default Pet Tasks. These are tasks that are automatically created on new pet, when that pet is created. This is for common tasks that always apply, and saves the staff from having to manually create them on each pet.

Tasks now have recurring functionality. That is, staff can create/edit a task on a pet and set it as recurring, and also set a next_due_date_in_days value. When this recurring task is marked complete, a new, identical task is created, and its due date will be today + the next_due_date_in_days value. 

This ticket is to add ~~two~~  a new attribute to Default Pet Tasks that will create recurring tasks on a pet. 

Acceptance criteria:
Add to DefaultPetTask (table, model, form)
- Migration to add recurring, boolean ~~and next_due_date_in_days, integer~~ to the default_pet_tasks table
- Add recurring, checkbox, and ~~next_due_date_in_days, positive integer input~~ to the default pet tasks form
- Add new attr to accepted params in controller, and if it makes sense, validate that due_in_days is positive on the model
- Update the TaskService to create the first task. Due date will be Time.current + due_in_days, and Task's next_due_in_days will be the Default Task's due_in_days
- Update tests

jmilljr24:
Resolves #508 

# ✍️ Description

1. Migration to add recurring, boolean to the default_pet_tasks table
2. Add recurring, checkbox input to the default pet tasks form
3. Add new attr to accepted params in controller (validation of due_in_days on the model was previously implemented)
4. Update the TaskService to include recurring and next_due_date_in_days as applicable. (due_date previously implemented)
5. Update tests

I also added a functioning check box for recurring to the index page of the default pet tasks. The table already had a spot for it. If it adds unneeded complexity or doesn't make sense I can take it out. Just need to know how you'd rather have the status displayed.

Let me know if I missed the mark anywhere!

# 📷 Screenshots/Demos

Screenshot from 2024-03-22 11-34-06

[Screencast from 03-22-2024 11:37:49 AM.webm](https://github.com/rubyforgood/pet-rescue/assets/16829344/f36749d2-ae70-42d4-b402-f8726f0da9b7)


kasugaijin:
I will get time to review in detail later but does it create a due date in the new recurring tasks? I did not see one populated in the task form.

kasugaijin:
Looks good per the gif - will review in detail soon. 
That column for recurrence is actually intended for a boolean Yes or No rather than to allow editing of the actual recurring db value. There are some issues including it in the table A) I could see someone turning it off but not also deleting due in days via the edit form, so they would have a task with a due date and B) They could set a non recurring task there but not enter a due in days value. Could you please set the value in that column to Yes or No and change the table column heading to 'Recurring'?
Also, do you think we need to use JS to ensure that if recurrence is checked true in the form, a due in days value is also entered? Or we could handle this in a custom model validation, also. We have somethins similar in a JS controller for the Task form, but it does behave a little differently. I guess from a UX perspective both validations is better, but you tell me how much you want to do here, and we can create a new issue if this is scope creep!

jmilljr24:
I believe I understand your intent with the enforcing the recurring with due_in_day. Maybe I'm missing something as this differs from how staff can create a task on a pet.

A new task can be made by staff with no due_date and with recurring true. Is that intended? A default pet task with due_in_days as nil and recurring true would create the same thing that staff can do with a new task. 

I'll change the UI table to yes/no as that's simpler but I want to make sure the default_pet_tasks can generate the same combinations as if staff added a new task to a pet individually.

The current Tasks combinations that work right now are as follows:

1. No due_date, No recurring, No next_due_date_in_days
2. Only due_date, No recurring, No next_due_date_in_days
3. With due_date, recurring and next_due_date_in_days
4. With recurring, No due_date, No next_due_date_in_days

The video shows all those combines. These were created from default_pet_tasks but there is nothing stopping you from creating them manually on the pet the same way with the current tasks implementation.

The Task form gives an error if you try to add recurring without a next_due_date_in_days if you have a due_date entry which makes sense. However that is not an issue with default_pet_tasks because next_due_date_in_days is created based off the due_in_days param.

kasugaijin:
> I believe I understand your intent with the enforcing the recurring with due_in_days. Maybe I'm missing something as this differs from how staff can create a task on a pet.
> 
> A new task can be made by staff with no due_date and with recurring true. Is that intended? A default pet task with due_in_days as nil and recurring true would create the same thing that staff can do with a new task.
> 
> I'll change the UI table to yes/no as that's simpler but I want to make sure the default_pet_tasks can generate the same combinations as if staff added a new task to a pet individually.
> 
> The current Tasks combinations that work right now are as follows:
> 
> 1. No due_date, No recurring, No next_due_date_in_days
> 2. Only due_date, No recurring, No next_due_date_in_days
> 3. With due_date, recurring and next_due_date_in_days
> 4. With recurring, No due_date, No next_due_date_in_days
> 
> The video shows all those combines. These were created from default_pet_tasks but there is nothing stopping you from creating them manually on the pet the same way with the current tasks implementation.
> 
> The Task form gives an error if you try to add recurring without a next_due_date_in_days if you have a due_date entry which makes sense. However that is not an issue with default_pet_tasks because next_due_date_in_days is created based off the due_in_days param.

Ah yes you are right! We can create tasks that are recurring with no due date. If we create a recurring task with a due date, then we require a next due in days value. That's where I slipped...going on memory instead of looking at the code.

Ok so I have checked it out and played with it and it looks great. Forget the validations - not needed for this. 

Could you just make the small change to the Index table, then?

jmilljr24:
Screenshot from 2024-03-22 22-01-30

app/services/organizations/default_pet_task_service.rb
```diff
@@ -7,12 +7,15 @@ def initialize(pet)
 
   def default_tasks_array
     @default_pet_tasks.map do |task|
+      next_due = (task.recurring.!) ? nil : task.due_in_days
```
kasugaijin, line 10: This makes me want a scope on default pet task that allows us to call task.recurring? Instead. You don’t have to do it in this PR. But would prefer to read it like English when we can :) Go Ruby

app/views/organizations/default_pet_tasks/index.html.erb
```diff
@@ -36,7 +34,7 @@
                     <%= task.due_in_days ? pluralize(task.due_in_days, "day") : "" %>
                   </td>
                   <td>
-                    -
```
kasugaijin, line 39: Actually you use it here and since it’s a Boolean we can call it directly.

kasugaijin (approved):
Looks great @jmilljr24 just one small non blocking comment.

app/services/organizations/default_pet_task_service.rb
jmilljr24: Yes that looks much better.

kasugaijin:
Thank you @jmilljr24

Issue #600, opened 2024-03-28
Assign Roles on Account Creation


mononoken:
## Description
Currently, roles are not assigned in any flows outside of the seed files. This means if you sign up for a new account currently, you are never assigned a role and thus do not have the necessary permissions to navigate the site as would be expected.

We want new users to be given these roles during their account creation.

## Acceptance Criteria
_Note: I removed the fosterer from this criteria. That work will move to a new issue_
- [x] When signing up for an account as an adopter, the User is assigned the `adopter` role
- [x] When User is invited as staff, the User is assigned the `staff` role
- [x] When User is invited as admin, the User is assigned the `admin` role
- [x] Test each of these separate role assignments:
  - [x] adopter
  - [x] staff
  - [x] admin

PR #602, opened 2024-03-28
600 assign roles

mononoken:
# 🔗 Issue
#600 

# ✍️ Description
This implements assigning the role of `:adopter` to `Users` when they register as an adopter. Currently, users who sign up for an account do not get this assignment, and thus do not have adopter permissions to navigate the site.

It also adds tests for this behavior, as well as for the existing assignment of the `:staff` and `:admin` roles from the invitations.

This does not implement the assignment of the `:fosterer` role as that role has not been added yet, and we also have not created the workflow for staff inviting fosterers yet.

I think we should leave the issue open for now to remember that we need to do this for fosterers. I've marked the tasks related to this PR as completed in the issue.

Some small naming/typo fixes are also included.

### DIFF
--- a/app/controllers/organizations/invitations_controller.rb
+++ b/app/controllers/organizations/invitations_controller.rb
@@ -15,11 +15,13 @@ class Organizations::InvitationsController < Devise::InvitationsController
     authorize! StaffAccount, context: {organization: Current.organization},
       with: Organizations::InvitationPolicy
 
-    @user = User.new(user_params.merge(password: SecureRandom.hex(8)).except(:roles))
+    @user = User.new(
+      user_params.merge(password: SecureRandom.hex(8)).except(:roles)
+    )
+    @user.add_role(user_params[:roles], Current.organization)
     @user.staff_account = StaffAccount.new
 
     if @user.save
-      @user.add_role(user_params[:roles], Current.organization)
       @user.invite!(current_user)
       redirect_to staff_index_path, notice: "Invite sent!"
     else
--- a/test/controllers/organizations/invitations_controller_test.rb
+++ b/test/controllers/organizations/invitations_controller_test.rb
@@ -2,6 +2,43 @@ require "test_helper"
 require "action_policy/test_helper"
 
 class Organizations::InvitationControllerTest < ActionDispatch::IntegrationTest
+  context "#create" do
+    setup do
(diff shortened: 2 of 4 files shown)


app/controllers/organizations/invitations_controller.rb
```diff
@@ -15,11 +15,13 @@ def create
...
 
-    @user = User.new(user_params.merge(password: SecureRandom.hex(8)).except(:roles))
+    @user = User.new(
+      user_params.merge(password: SecureRandom.hex(8)).except(:roles)
+    )
+    @user.add_role(user_params[:roles], Current.organization)
     @user.staff_account = StaffAccount.new
 
     if @user.save
-      @user.add_role(user_params[:roles], Current.organization)
```
mononoken, line 22: I moved the add_role call to before the `save` call. That way, the role is associated with the user when the user validations happen. Plus, it also means if there is a validation error with adding the role, it will raise before the `User` is persisted.

test/controllers/organizations/invitations_controller_test.rb
```diff
@@ -30,7 +67,7 @@ class Organizations::InvitationControllerTest < ActionDispatch::IntegrationTest
             first_name: "John",
             last_name: "Doe",
             email: "john@example.com",
-            roles: "admin"
+            roles: "staff"
```
mononoken, line 70: I just changed this because I just happened to see we used "admin" as a param in another test for this controller, so I figured it is kind of nice to have both values used in separate tests. This is not an error.

nsiwnf (approved):
lgtm - did you just want to leave the issue open or the PR as well for foster-related work?

mononoken:
> lgtm - did you just want to leave the issue open or the PR as well for foster-related work?

Thanks for reviewing, @nsiwnf ! I think we can merge this PR but leave the issue open. Or we can move the fosterer part of the issue to its own separate issue if you would prefer.

kasugaijin:
> > lgtm - did you just want to leave the issue open or the PR as well for foster-related work?
> 
> Thanks for reviewing, @nsiwnf ! I think we can merge this PR but leave the issue open. Or we can move the fosterer part of the issue to its own separate issue if you would prefer.

Nice work! @mononoken I'd say let's merge this and move the foster piece to a new issue.

PR #604, opened 2024-03-28
Create adopter applications scope to organization and ensures no dupl…

kasugaijin:
# 🔗 Issue

# ✍️ Description
Seed data was creating adopter apps that were on pets or for adopter foster accounts in different orgs. This was causing errors, and obviously not desirable. This fixes to scope them to org and ensure no duplicate apps.

# 📷 Screenshots/Demos


kasugaijin:
We will take a different approach instead - need to declare association for every model with Org using acts_as_tenant. Will make a PR to do so.

Issue #605, opened 2024-03-29
Scope models to organization


kasugaijin:
We have a number of models that are not scoped to the tenant (organization), so within the tenant context, if you call Model.all you will get all records for all organizations. This is not desirable and we should leverage the acts_as_tenant gem to scope these resources accordingly. 

Resources to scope:
- AdopterApplication
- AdopterFosterAccount
- AdopterFosterProfile
- Donation
- Task

Acceptance criteria
- Migration to organization_id to each corresponding model table
- Update each model to belong to org `acts_as_tenant(:organization)`
- No broken tests!

PR #606, opened 2024-03-29
605: Scope resources to Org

kasugaijin:
… and remove the donations logic as it is not required for MVP

# 🔗 Issue
https://github.com/rubyforgood/pet-rescue/issues/605

# ✍️ Description
**We will need to reset the db in production as I am adding a new column to tables with null false and not backfilling data. Doesn’t make sense to backfill seed data when we have the luxury of resetting**

Adds org id column to resource tables, and association on models, for resources that should be scoped under Org. This should scope resources to the Org when calling Resource.all within the tenant context, instead of retrieving all records from all orgs, as was happening in the seed files.

Also removes the donation logic/tests as we do not need that for MVP.

I note that we set the tenant/org context 
in the base controller under controllers/organizations. This means we have other controllers that are not scoped within this context, and there is risk that a developer could call Resource.all, expecting it to be scoped, especially if the policy context is not applied. 

What do we think about moving all organization specific logic to the controllers/organizations folder so that all logic carried out here is within the tenant context? I think it will alleviate aforementioned risks, and intuitively it makes sense we organize things this way. This would mean the following (not ruled out) moving into that directory:
image

This would require a bit of work to reorganize files, and update routes, controller class names, policies but not a huge untertaking. Thoughts?


nsiwnf:
>What do we think about moving all organization specific logic to the controllers/organizations folder so that all logic carried out here is within the tenant context?
Yeah, I like this idea 👏

Seems like theres a few things happening in this PR - can we split them out?

kasugaijin:
I will close this PR and make three new PRs to split out the above code.

Issue #590, opened 2024-03-23
Fix: Parameterize contact [org] mailer


nsiwnf:
### What
Currently the "contact" form triggers the `ContactsMailer` which has hard-coded values for org details. 

route ex `/alta/contacts`

### Acceptance Criteria
- Update the values in this workflow to use the current organization contact info
- Update contact_forms test or add new tests for this workflow

MooseCowBear:
I can work on this

PR #607, opened 2024-03-30
590 parameterize contact mailer

MooseCowBear:
# 🔗 Issue

#590 

# ✍️ Description

Replaces hard coded organization info with info from current org, and adds tests for mailer and contacts controller. 

I also switched the route for the create action to use post instead of get, and added a scope to the new contact form to send parameters in the form that the strong params want so we can pass `contacts_params` to `Contact.new` and `ContactsMailer.with` instead of each individually. 

In the controller, I had to replace `current_tenant.slug` with `Current.organization.slug` to get it to work I think because the contacts controller doesn't include `OrganizationScopable`. Should contacts be scoped to the organization? 

# 📷 Screenshots/Demos


nsiwnf (approved):
This looks good to me! I think we can do the scope-to-org in a separate PR 👍

PR #614, opened 2024-03-31
Add Org ID to resources to scope them

kasugaijin:
# 🔗 Issue
https://github.com/rubyforgood/pet-rescue/issues/605

# ✍️ Description
Tests in this PR will fail until this PR has been reviewed and merged: https://github.com/rubyforgood/pet-rescue/pull/612
I also suggest we review and merge this PR beforehand: https://github.com/rubyforgood/pet-rescue/pull/613

**We will need to reset the db in production as I am adding a new column to tables with null false and not backfilling data. Doesn’t make sense to backfill seed data when we have the luxury of resetting**

This PR scopes resources to Org Id so when we call Resource.all in the tenant context it should only return those records scoped to the tenant.

I note that we set the tenant/org context
in the base controller under controllers/organizations. This means we have other controllers that are not scoped within this context, and there is risk that a developer could call Resource.all, expecting it to be scoped, especially if the policy context is not applied.

What do we think about moving all organization specific logic to the controllers/organizations folder so that all logic carried out here is within the tenant context? I think it will alleviate aforementioned risks, and intuitively it makes sense we organize things this way. This would mean the following (not ruled out) moving into that directory:
image

This would require a bit of work to reorganize files, and update routes, controller class names, policies but not a huge untertaking. Thoughts?

# 📷 Screenshots/Demos


mononoken (approved):
Looks good. I'm glad we caught this now. Thanks for adding these, Ben!

And to reiterate from the meeting, I agree with moving the files to `organizations/` in an additional issue.

kasugaijin:
> Looks good. I'm glad we caught this now. Thanks for adding these, Ben!
> 
> And to reiterate from the meeting, I agree with moving the files to `organizations/` in an additional issue.

issue made: https://github.com/rubyforgood/pet-rescue/issues/621

kasugaijin:
@nsiwnf can I lave this to you to merge and reset the prod db?

Issue #592, opened 2024-03-25
Org Page Content Epic: As staff I can upload a Hero Image and up to two About Us images


kasugaijin:
Currently blocked by https://github.com/rubyforgood/pet-rescue/issues/480

As staff, you can go to the dashboard and see the Page Text navbar option, where you can update the Hero Text and About Us text. This also needs to allow staff to update their Hero Image (1 image) and About Us Images (1 or 2 images) on the org's root/home page.

Implementation Criteria:
- Use the PageText model for the attachments: `has_one_attached :hero_image`and `has_many_attached :about_us_image`
- On the PageText form add attachment inputs, clearly labelled. One for Hero Image (allows one attachment), One for About Us images (allows one or two attachments). 
- Add model validations to ensure the files are images, not massive (see the Pets Controller for example)
- Add controller actions on PageTextsController for each type of attachment to attach (can use existing attachments controller for purge)
- Staff should be able to upload and purge (delete) an image. 
- Resuse partials where it makes sense e.g., the upload input, and the table to show the uploaded images (see the partials used in the Pet tabs for photos and files)
- Add permissions to the policy (may not be needed based on what is added in https://github.com/rubyforgood/pet-rescue/issues/480).
- On org root/home, render the Org.page_text.hero_image if it exists, and render either one or both About Us images, if they exist. If there is only one image, then use an existing default image for the second.
- Tests 

As next steps (different issue) we will want to look into recommending image dimensions and handling resizing of image variants. Feel free to include in this issue if you want to.

I may have forgotten something here - hit me up in Slack!

ErinClaudio:
Can I give this one a crack? @kasugaijin 
If yes I would like to write a breakdown of my path forward before beginning any coding.

kasugaijin:
@ErinClaudio yes sounds good!

ErinClaudio:
@kasugaijin 

For clarification, the functionality for updating the Hero Image and About Us Images should be integrated directly within the 'Page Text' navbar option, as opposed to introducing a separate button?

ErinClaudio:
My notes the project
https://gist.github.com/ErinClaudio/393ca48b7f2a77b8d65d5ab36a7f120c

kasugaijin:
> @kasugaijin
> 
> For clarification, the functionality for updating the Hero Image and About Us Images should be integrated directly within the 'Page Text' navbar option, as opposed to introducing a separate button?

[...]

PR #618, opened 2024-04-02
Update with Image frontend buttons

ErinClaudio:
# 🔗 Issue
https://github.com/rubyforgood/pet-rescue/issues/592

# ✍️ Description

- Overview: Update the dashboard to allow staff to manage Page Text options, including updating Hero Text, About Us text, and corresponding images on the organization's root/home page.
- Implementation:
  - Utilize PageText model with has_one_attached :hero_image and has_many_attached :about_us_images.
  - Add form inputs for attachments on the PageText form: one clearly labeled for Hero Image (single attachment) and one for About Us images (one or two attachments).
  - Implement model validations for image file type and size restrictions, referencing the Pets Controller for examples.
  - Update PageTextsController to handle image attachments and purging.
  - Ensure staff can upload and purge images.
  - Reuse partials for upload inputs and image display tables, similar to those used in Pet tabs.
  - On the org's root/home page, render the hero_image and one or both about_us_images as applicable, with a default image for the second if only one is uploaded.
- Testing:
  - Modify tests to accommodate changes to the page_text model. Pay special attention to these modifications.

I do not like how the about us images upload but I could use a little feedback

# 📷 Screenshots/Demos


config/routes.rb
```diff
@@ -54,6 +54,11 @@
   get "/terms_and_conditions", to: "static_pages#terms_and_conditions"
   get "/cookie_policy", to: "static_pages#cookie_policy"
 
+  resources :adopter_applications, path: "applications",
```
kasugaijin, line 57: These two resource routes should not be added here - they have been moved under organization scope. Looks like these came from an old branch?

app/controllers/organizations/page_texts_controller.rb
```diff
@@ -8,14 +8,14 @@ def update
...
   end
 
   private
 
   def page_text_params
-    params.require(:page_text).permit(:hero, :about)
+    params.require(:page_text).permit(:hero, :about, :hero_image, about_us_images: [])
   end
 
   def set_page_text
```
kasugaijin, line 21: This is slightly unrelated but I found a bug while testing - could you please change line 22 from 
`@page_text = current_user.organization.page_text` to
`@page_text = PageText.first` ?
I tried changing the slug from Baja to Alta and if gives the following error because the user context doees not exist after redirecting to the Alta site 
```
undefined method `organization' for nil:NilClass
    @page_text = current_user.organization.page_text
```

Instead we want to load the PageText and let the policy check for `pre_check :verify_organization!` check the user's org matches the PageText's org.

app/views/organizations/home/index.html.erb
```diff
@@ -144,14 +144,19 @@
...
             <span>
-                <%= image_tag('danielle_2.jpg', 
+                <%= image_tag('danielle_2.jpg',
                     class: 'rounded-5 w-100 px-4 px-lg-0') %>
             </span>
         </div>
         <div class="col-lg-4 px-4 order-last order-lg-0">
             <p class='justify bigger'>
               <%= Current.organization.page_text&.about || "#{Current.organization.name} was founded by an incredible group of ladies in April of 2020. Our initial goal was to have both a rescue and a spay/neuter clinic, however, we quickly realized that it would be more efficient to separate into two organizations."%>
             </p>
```
kasugaijin, line 154: Can we apply a similar approach to the Hero image for the about us images? I.e., if we have an uploaded image, we use that, otherwise, we use the default?

The about us is a bit different because we have two images, but we could check the number of about us images attached, if 0, we use both default images, if 1, we replace one of the default images, if 2, we replace both of the default images. Let me know if you have any questions.

app/views/organizations/page_texts/_form.html.erb
```diff
@@ -3,34 +3,50 @@
...
+                                class: 'form-control' %>
+          </div>
+          <div class="form-group">
+            <%= form.file_field :hero_image,
+                                label: "Hero Image",
+                                class: 'form-control'%>
+          </div>
+          <div class="form-group mt-4">
+            <%= form.text_area :about, label: "About Us", placeholder: "About Us text", class: 'form-control' %>
+            <small class="form-text text-muted">You can upload up to 2 images. </small>
```
kasugaijin, line 25: Can we move this text below the input instead? And can we also add the image spec requirements i.e., `Images must be .png or .jpeg under 2MB`

app/views/organizations/page_texts/_form.html.erb
```diff
@@ -3,34 +3,50 @@
...
-            </div>
-          </form>
+          <div class="form-group">
+            <%= form.text_field :hero,
+                                label: "Hero",
+                                autofocus: true,
+                                placeholder: "hero text",
+                                class: 'form-control' %>
+          </div>
+          <div class="form-group">
```
kasugaijin, line 18: Like below, can we also add a hint below the input for the image spec requirements i.e., `Images must be .png or .jpeg under 2MB`

test/system/registration_test.rb
```diff
@@ -4,6 +4,13 @@ class RegistrationTest < ApplicationSystemTestCase
   setup do
     @user = create(:staff)
     @organization = @user.organization
+    @page_text = create(:page_text, organization: @organization)
```
kasugaijin, line 7: For these 3 system tests, we can remove `Current.organization = @organization`. 
Also, to DRY it up a little, given these are required to set up the PageText that is called in the root page, we could move this logic to a helper method, and just call that method in each setup. That will make any future changes to the page text easier to update in the test setup.

You should be able to create a new file in `test/helpers` that is a module, name something like PageTextSetup or the likes, and put a method in there e.g., `def set_up_page_text` and include that module in these system test classes, then call that method in the setup.

test/controllers/organizations/home_controller_test.rb
```diff
@@ -3,6 +3,11 @@
 class Organizations::HomeControllerTest < ActionDispatch::IntegrationTest
```
kasugaijin, line 3: For all of these tests, I gather you had to create a PageText for each org so there are no errors when the view calls for the org's page text.

Instead of creating them in the setup of each test class, can't we just create the PageText when we create an organization in the factories? Then any time we create an org via factory, which is pretty much most tests, there will be an associated PageText, and there won't be any need to have to add this to every test setup. 

For example, check out the factory for organizations, and you will see that we pass a `profile` object to the Organization. We could also pass it a PageText object in the same way, so then there is no need to create one after the fact.

Does that make sense? Then we should be able to get rid of all these additions to the test setups.

kasugaijin:
Hey @ErinClaudio the implementation looks good! Just a couple small details to tweak. 

For tests, I think we can handle setting up the PageText differently - see my comment on that. 
I also think we could add a controller test for uploading an image.

Also, when I run the tests locally, this test keeps failing for me (though in the pipeline it passes, so maybe there's something going on locally for me...does it pass for you locally?

```
Organizations::PageTextsControllerTest#test_: PATCH #update should update page text.  [test/controllers/organizations/page_texts_controller_test.rb:53]:
Expected: "canine caped crusader"
  Actual: "Eius id minus et."
```

kasugaijin (approved):
Good job @ErinClaudio! This is good chunk of work that adds value to the customization of the home page for each org. Thank you for seeing it through!

PR #620, opened 2024-04-02
Adopter foster dashboard

kasugaijin:
# 🔗 Issue
https://github.com/rubyforgood/pet-rescue/issues/619

# ✍️ Description
Adds the layout for Adopter Foster Dashboard. Will need updating once we have adopter role added to add conditional links. 
I also refactored the avatar helper to give the option to create a small, or XL avatar.

Work to refactor the top navbar and get rid of the avatar dropdown will be in a separate PR.

# 📷 Screenshots/Demos
image

With initial avatar
image

pets


app/views/layouts/adopter_foster_dashboard.html.erb
```diff
@@ -0,0 +1,118 @@
...
+                    <div class="d-flex align-items-center">
+                      <div class="me-2 position-relative d-flex justify-content-end align-items-end mt-n5">
+                        <%= render_avatar_partial(xl: true) %>
+                      </div>
+                      <div class="lh-1">
+                        <h2 class="mb-0"><%= "#{current_user.first_name} #{current_user.last_name}" %></h2>
+                      </div>
+                    </div>
+                    <div>
+                      <a href="add-course.html" class="btn btn-primary d-none d-md-block">Sample Button</a>
```
mononoken, line 29: Think we should make this href blank for now, since this is otherwise fetching a non-existent path.

mononoken (approved):
lgtm. Thanks for building this, Ben! I just have the one comment about a link.

I have some thoughts from a design standpoint:
- Do we want the staff dashboard to reflect this design as well or do we want to upkeep two different dashboard layouts?
- Having two hamburger menus *feels* wrong to me, especially when some of the content is duplicated in both (such as log out).
- I think it feels wrong because the user has to learn which hamburger menu is the one they want. They have no way of knowing which contains the link they want without opening both. We don't want users opening a hamburger menu and realizing they meant to open the other one. That's frustrating for UX. These considerations may affect #622

kasugaijin:
> lgtm. Thanks for building this, Ben! I just have the one comment about a link.
> 
> I have some thoughts from a design standpoint:
> 
> * Do we want the staff dashboard to reflect this design as well or do we want to upkeep two different dashboard layouts?
> * Having two hamburger menus _feels_ wrong to me, especially when some of the content is duplicated in both (such as log out).
> * I think it feels wrong because the user has to learn which hamburger menu is the one they want. They have no way of knowing which contains the link they want without opening both. We don't want users opening a hamburger menu and realizing they meant to open the other one. That's frustrating for UX. These considerations may affect [Refactor top navbar #622](https://github.com/rubyforgood/pet-rescue/issues/622)

Thanks @mononoken!
I think each dashboard design works better for each user type/flows/data. The hard work has been done now in setting it up so the cost going ahead should be minimal.

I also do not love two hamburger menus. It works, but I agree it is not the most user friendly. Though, I have seen it before and as a user, you just have to figure it out. I agree we do not need to duplicate any links, so we can manage that easily enough. In terms of only having one hamburger, that would mean reducing the number of links in the top navbar (on dashboard view) to the degree where they would fit logicially in the top nav on mobile view. I cannot see a way round the sidebar hamburger. 

Another point of consideration, though, is that we chose this theme in the beginning to speed up development flow. One of the core assumptions we undertook in that decision was that we take what the theme gives us as devs, and roll with it. It keeps design and implementation consistent, and acts as a source of truth so to speak. This is valuable when building out to an MVP, quickly. In that sense, I would say we leave this for MVP (it does work, it just isn't lovely) and solve this if it becomes a problem based on user feedback.

For reference - from the theme itself
image

mononoken:
Considering those points for later sounds good to me. Thanks for the thorough response, Ben 🙏

PR #625, opened 2024-04-03
Small policy and factory fixes

mononoken:
# 🔗 Issue
General fixes not associated with any issue.

# ✍️ Description
When I was playing around with refactoring the policies, as discussed in the last meeting, I found a few things to fix in the policies and factories.  I will leave comments to explain some and feel free to ask about the others.


app/policies/adoptable_pet_policy.rb
```diff
@@ -6,7 +6,7 @@ class AdoptablePetPolicy < ApplicationPolicy
   end
 
   def show?
-    allow! if permission?(:manage_pets)
+    allow! if allowed_to?(:manage?, record, namespace: Organizations)
```
mononoken, line 9: This utilizes Action Policy to call a policy rule from within another policy. The permission check was fine as well, but this means that other logic besides the permission check are also carried over.

test/factories/pets.rb
```diff
@@ -14,12 +14,12 @@
...
     trait :adoption_pending do
-      adopter_applications { build_list(:adopter_application, 3, :adoption_pending) }
+      adopter_applications { build_list(:adopter_application, 3, :adoption_pending, pet: instance) }
     end
 
     trait :adopted do
-      match { association :match }
-      adopter_applications { build_list(:adopter_application, 1, :successful_applicant) }
+      match { association :match, pet: instance }
+      adopter_applications { build_list(:adopter_application, 1, :successful_applicant, pet: instance) }
```
mononoken, line 22: This was an especially annoying bug. I realized that `FactoryBot.create(:pet, :adopted)` was instantiating 2 pets, not 1. You can try this in console in `main`:
```ruby
Pet.count # 10
FactoryBot.create(:pet, :adopted)
Pet.count #12
```
A second pet is instantiated within the association since the instance is not passed to it.

test/policies/adoptable_pet_policy_test.rb
```diff
@@ -4,12 +4,28 @@
 class AdoptablePetPolicyTest < ActiveSupport::TestCase
   include PetRescue::PolicyAssertions
 
-  def setup
-    @policy = -> { AdoptablePetPolicy.new(@pet, user: @user) }
+  context "relation_scope" do
```
mononoken, line 7: This test was missing previously. The `relation_scope` was already defined previously.

app/policies/organizations/invitation_policy.rb
```diff
@@ -1,5 +1,6 @@
 class Organizations::InvitationPolicy < ApplicationPolicy
   pre_check :verify_organization!
+  pre_check :verify_active_staff!
```
mononoken, line 3: I figured some of these admin only actions should still have `verify_active_staff!` incase an admin's staff account was deactivated for some reason.

kasugaijin (approved):
LGTM

Issue #623, opened 2024-04-03
Tenant Footer: Change text and add link to the application root


kasugaijin:
We want people browsing an org (tenant) site, who can see the tenant footer, to know that this website is provided by Pet Rescue application, and that they can also get a website just like it, and be able to click a link to go to the application root page

![image](https://github.com/rubyforgood/pet-rescue/assets/95949082/a5edda9c-340b-451d-8300-e285aeb1bfde)

Acceptance Criteria
- change `We provide a platform for Pet Rescues and Animal Service Organizations` to something like `If you are a pet adoption or foster organization and want a website like this one, for free, click here.` The click here should be a link to the organization root i.e., `petrescue.com`.

Gabe-Torres:
Can I be assigned this issue? 🐶

kasugaijin:
@Gabe-Torres @jeevikasirwani made a pull request for this issue so it’s already been spoken for (I could add them on this issue as assigned, sorry for that). We will have more good first issues coming up soon!

Gabe-Torres:
Gotcha, No problem! Thank you

PR #626, opened 2024-04-03
fix:Tenant Footer: Change text and add link to the application root #623

jeevikasirwani:


kasugaijin (commented):
Thank you!
Could you please use a rails link_to helper and route helper instead of static url? e.g., `<%= link_to "click here", root_path %>.` You might want to check Rails Routes to confirm the path. 

And also please include a screenshot for UI changes.

kasugaijin:
@all-contributors
please add @jeevikasirwani for code.

kasugaijin (commented):
can you please use a rails route helper instead of that static url? Check rails routes to determine the correct path, it should look something like `root_path`. 

So the final product should look something like 
```
<p> If you are a pet adoption or foster organization and want a website like this one, for free, <%= link_to 'click here', root_path %></p>
```

jeevikasirwani:
yes i did make the changes ,instead of root path i have added the link ,in my recent commit

<p>
              If you are a pet adoption or foster organization and want a website like this one, for free,
              <%= link_to 'click here', 'https://petrescue.com' %>
            </p>

kasugaijin:
> yes i did make the changes ,instead of root path i have added the link ,in my recent commit
> 
> If you are a pet adoption or foster organization and want a website like this one, for free, <%= link_to 'click here', 'https://petrescue.com' %>

Yes I see that thanks. But we need to use the rails route helper method because this url will not work unless it’s in that specific domain. We need this to work on any domain whether local or on our testing site in heroku. Please look up the rails routes to figure out what helper to use and ask in slack if you’re having issues.

kasugaijin:
To access the routes of your app you can either do it in the console or the browser. I prefer the browser. With your app server running, visit http://localhost:3000/rails/info/routes in your browser. 

The route helper we want is `/` i.e. the home route
image

As you can see, the helper method (left most column) is the `root_path`. So you can use that instead of your hard coded URL

jeevikasirwani:
thank for for help ,i have change please check if thats the solution needed

config/routes.rb
```diff
@@ -41,6 +41,7 @@
   match "/500", to: "errors#internal_server_error", via: :all
 
   root "root#index"
+  get "/petrescue_home", to: redirect("https://petrescue.com")
```
kasugaijin, line 44: You can delete this

app/views/layouts/shared/_footer.html.erb
```diff
@@ -7,7 +7,11 @@
         <div class="mb-4">
           <h3> Pet Rescue </h3>
           <div class="mt-4">
-            <p> We provide a platform for Pet Rescues and Animal Service Organizations </p>
+          <p>
+              If you are a pet adoption or foster organization and want a website like this one, for free,
+              <%= link_to 'Home', petrescue_home_path %>
```
kasugaijin, line 12: change the path to `root_path`

kasugaijin (approved):
Looks good thank you!

PR #630, opened 2024-04-03
Move specific controllers to organizations scope and update tests

### LINKED ISSUE #621: Scope organization specific controllers under organization
We set the tenant in `app/controllers/organizations/base_controller.rb` using `OrganizationScopable`. The controllers scoped under `controllers/organizations` inherit from `Organizations::BaseController`. This means the logic in these controllers is executed in the tenant scope, e.g., calling `Task.all` will return all tasks for current organization only. This helps to keep data secure between organizations.

We have some routes/controllers that are used to create common Organization workflows, but they are not currently scoped under Organization. This is not ideal because it means that calling resources in these controllers could return all records from all organizations (e.g., `Task.all` will return all tasks from all orgs), presenting a risk of data bleeding. So, we need to move these routes, controllers under the organization scope. 

Acceptance criteria:
- move the list of controllers and associated routes (not crossed out) below into `controllers/organizations` and inherit from `Organizations::BaseController` instead of `ApplicationController`
- Rename the classes appropriately for their new scope
- Fix tests (renaming)
image

jmilljr24:
resolves #621 

# ✍️ Description

This moves specific controllers under the scope of organizations for appropriate tenant scope. The following controllers were moved:

- adoptable_pets
- adopter_applications
- adopter_foster_profiles
- matches
- profile_reviews

The associated routes were scoped to organizations and tested were updated.

One thing I am a bit confused on is regarding Action Policy for adopter_applications_controller. There are two preexisting policies called `adopter_application_policy.rb` and `organizations/adopter_application_policy.rb`. I'm not sure exactly where `organizations/adopter_application_policy.rb` is used (adoption_application_reviews_controller?) 

Because action policy defaults to inferring the policy from the controller name, moving adopter_applications_controller under organizations causes it to try and use `organizations/adopter_application_policy.rb`. I used the explicit `with:` option to specify the correct policy which I don't immediately see an issue with. If I'm missing something or there is a better way to do this I'm all ears.


kasugaijin (commented):
Thanks @jmilljr24! Should we also move the respective policies under the organization scope? For example, we have moved AdopterFosterProfilesController under Org, but its policy remains outside of the Org scope. I would have guessed the policy would not work if it is not scoped the same as it's calling controller given it relies on the controller name to find it?

You have a good question about the adopter applications policy. @mononoken will have to remind me of the reason for the naming (I think there was discussion around this). We have two adopter application related controllers, one for adopters to make applications (AdopterApplicationsController), and one for staff to review them (AdoptionApplicationReviewsController). Now these will be in the same scope, I think we will need to rename the Policy classes if we move them also to the same scope?

jmilljr24:
> Thanks @jmilljr24! Should we also move the respective policies under the organization scope? For example, we have moved AdopterFosterProfilesController under Org, but its policy remains outside of the Org scope. I would have guessed the policy would not work if it is not scoped the same as it's calling controller given it relies on the controller name to find it?
> 
> You have a good question about the adopter applications policy. @mononoken will have to remind me of the reason for the naming (I think there was discussion around this). We have two adopter application related controllers, one for adopters to make applications (AdopterApplicationsController), and one for staff to review them (AdoptionApplicationReviewsController). Now these will be in the same scope under Org, I think we will need to rename the Policy classes if we move them also to the same scope?

Moving the policies under org was my original intention and that's were I discovered the conflict with `adopter_application`. Then I guess I didn't go back and try moving the others. Sorry about that. I will look into moving those those and we can reconvene on the adopter application policies when we get more info from Ken.

kasugaijin:
@jmilljr24 no problem at all! Sounds good.

mononoken:
> One thing I am a bit confused on is regarding Action Policy for adopter_applications_controller. There are two preexisting policies called `adopter_application_policy.rb` and `organizations/adopter_application_policy.rb`. I'm not sure exactly where `organizations/adopter_application_policy.rb` is used (adoption_application_reviews_controller?)
> 
> Because action policy defaults to inferring the policy from the controller name, moving adopter_applications_controller under organizations causes it to try and use `organizations/adopter_application_policy.rb`. I used the explicit `with:` option to specify the correct policy which I don't immediately see an issue with. If I'm missing something or there is a better way to do this I'm all ears.

Just looking at the controllers tests I wrote, these are where each policy is currently used on main:
`Organizations::AdopterApplicationPolicy` -> `Organizations::AdopterApplicationsReviewController`
`AdopterApplicationPolicy` -> `AdopterApplicationsController`

It's important to remember, the policy lookup from calling `authorize!` is not looking at the controller name, just the namespacing. It's actually using the name of the class of the object passed to the method. So:
```ruby
class Organizations::AdoptionApplicationReviewsController < Organizations::BaseController
  ...  
  def set_adopter_application
    @application = AdopterApplication.find(params[:id])
    authorize! @application
    # This is looking for @application.class.name + "Policy" with the Organizations scope
    # "Organizations::AdopterApplicationPolicy"
  end
```
You can read more here: https://actionpolicy.evilmartians.io/#/lookup_chain?id=policy-lookup

Now that we are moving one of these controllers in the same scope, they will try to use the same policy lookup.

Two solutions I can think of:
1. Specify using `with` as Justin mentioned. 
2. Further scope these controllers

I think there may also be a third option for using Action Policy to specify the policy class to use within a method in a controller too. I don't see it in a quick look, but I will look deeper tonight.

Now that these are both in `Organizations`, it is further illuminating what Ben said, that really the `reviews` is how staff interact with `AdopterApplication` and the other controller is how the adopter interacts with them. We could add a `staff` and `adopter` scope to `organizations`. I think this option may be the better option, especially if we can think of other controllers that would benefit from this scoping.

Scoping deeper gives me pause but the `organizations` scope is so broad, I think it's okay in this case. What do y'all think @jmilljr24 and @kasugaijin ?

kasugaijin:
Ah right thanks for the explanation @mononoken that makes sense now. Thinking about name spacing, I think there is benefit to having staff, adopter and foster namespaces inside of organizations. Relevant contollers live there and generic controller to org operations live in the org namespace. It’s a much bigger task to do. So @jmilljr24 what is your thoughts on this? We could stick with declaring with for now and add more namespaces later, or if you think it should be done at once, you can go for it.

jmilljr24:
@mononoken thank you for the detailed explanation. That definitely helps me understand it better. This is pushing the boundaries of my knowledge so I can't speak on the best way forward. (I have learned a lot just working through this so your guidance has been much appreciated!!)

I will push up the change with the other policies scoped under org that don't have the conflict. @kasugaijin I feel like a much bigger change would probably require too much of your undivided attention for me to do it successfully without a lot of help. So for now can we go the current structure?

kasugaijin:
> @mononoken thank you for the detailed explanation. That definitely helps me understand it better. This is pushing the boundaries of my knowledge so I can't speak on the best way forward. (I have learned a lot just working through this so your guidance has been much appreciated!!)
> 
> I will push up the change with the other policies scoped under org that don't have the conflict. @kasugaijin I feel like a much bigger change would probably require too much of your undivided attention for me to do it successfully without a lot of help. So for now can we go the current structure?

Yes that's fine! We can add this as a future task for code organization.

kasugaijin (approved):
LGTM!

Issue #633, opened 2024-04-04
Organization Profile: As staff I can save URLs on my organization profile. 


kasugaijin:
Organizations will benefit from being able to save URLs on their profile for things like social media and donation links. This issue is to enable this to happen. 

Acceptance Criteria:
- Migration on the organizations table to remove the about_us column, and to add three new text columns for facebook_url, instagram_url, donation_url. I did consider using string datatype, but by default is limited to 255 chars, and the jury seems to be out on the best datatype, I say text just because we can accept longer URLs if required. But, open to your ideas.
- Add the new URL input fields on the Organization Profile form. 
- On the organization profile model, add a URL validation for each attribute to make sure it is a valid URL being saved. Do not allow save if not a valid URL. 
- I should be able to add URLs in the UI, then retrieve them in the console from the OrganizationProfile as a dev.

MooseCowBear:
I'd like to work on this. Do we know how we want to go about validating urls? If not, I can look into options.

kasugaijin:
@MooseCowBear all yours. I think Rails 7 has a built in module for validating URLs. Or we can write a custom validator class, but I don't think we need to.

PR #637, opened 2024-04-09
633 organization profile urls

MooseCowBear:
# 🔗 Issue

#633 
# ✍️ Description

Adds Facebook, Instagram, and donation urls to organization profile and removes the `about_us` column. I updated existing tests and the seed files to remove references to about us. 

# 📷 Screenshots/Demos

This is what it looks like with the url fields beneath.

Screenshot 2024-04-09 at 6 05 01 PM

Is that too much empty space?

Another option: 

Screenshot 2024-04-09 at 5 38 58 PM


kasugaijin:
Looking good @MooseCowBear ! I think I prefer the second layout.

app/views/organizations/organization_profiles/_form.html.erb
```diff
@@ -71,6 +66,9 @@
 
       <!-- Third Column -->
       <div class="col-lg-6">
+        <%= form.text_field :facebook_url, label: "Facebook", class: 'form-control' %>
```
kasugaijin, line 69: can we add a placeholder value for each? Something like https://www.example.com just so they know its for URLs and to include the protocol.

kasugaijin (approved):
LGTM!

Issue #599, opened 2024-03-28
Add Fosterer Role


mononoken:
~~This is pending some discussion on the state of the policy model tests in Slack.~~
Per Slack and meeting discussion, we are going to leave policy unit tests as is for readability and ease of use. We will add fosterer role checks to all authz unit tests as indicated in acceptance criteria.

## Description
We are implementing users as fosterers and have created shared `AdopterFosterAccount` and `AdopterFosterProfile`. However, we need the `fosterer` role to actually assign the permissions for `fosterers` separately from `adopters`.

## Acceptance Criteria
- [x] User can be assigned a `Fosterer` role using rolify
- [x] `FOSTERER_PERMISSIONS` created and linked to fosterer role.
- [x] Fosterer role added to authorization tests

mononoken:
For documentation, here is the main Slack message pending this issue. If anyone happens to have thoughts on this here, please share them!

---
I am making an issue and starting to work on the fosterer role that we discussed in yesterday’s meeting. This got me thinking about how the authz models are currently tested:
Right now, for the most part, all policy models are tested something like this:
```ruby
class PolicyTest
  context "#action?" do
    context "no user"
    context "adopter"
    context "staff"
    context "admin"
  end
end
```
I like this method as it documents clearly what each role can do given the policy. While it creates a lot of duplication in tests, I personally don’t think duplication in tests is a problem.

While I was working on the authz branch, I thought of a different approach that I am reconsidering now.

Currently, each role is  just a way to fetch a list of preassigned permissions per role:
```ruby
  ADOPTER_PERMISSIONS = %i[
    create_adopter_applications
    view_adopter_applications
    withdraw_adopter_applications
    create_adopter_foster_profiles
    manage_adopter_foster_profiles
  ].freeze
```
Because of this, another approach would be to test like this:
```ruby
class PolicyTest
  context "#action?" do
    context "user has x permission" do
      before do
        stub permission to user
      end
    end
  end
end
```
This would make some policy tests way slimmer as we are directly testing the permission behavior. The policy tests no longer have any idea that roles even exist, just like the policy objects themselves. We could also then test the roles separately to assert each role has the expected permissions.

[...]

PR #641, opened 2024-04-13
599 add fosterer role

mononoken:
# 🔗 Issue
#599

# ✍️ Description
This implements the `:fosterer` role for authz purposes. It does so by implementing `FOSTERER_PERMISSIONS`. It also adds the fosterer factory and a fosterer context to many of the policy unit tests.

Adding the fosterer factory led me to refactor the user factory slightly in regards to how roles were assigned in our factories. It would be worth making sure that those changes make sense.

## Role assignment refactor
Roles were previously being added through the account factories. This refactoring moves the role assignment to the user factory.

The roles are associated now with `User` records, not `Account` records. Coupling the accounts with the roles was making testing for role related behavior clunky, especially since adopter and fosterers share the same account type, `AdopterFosterAccount`. 

For consistency, I made these changes to staff too.

## Bug note:
I noticed before these changes that running just the `MatchPolicyTest` file caused the test runner to hang:
```
bin/rails test -v test/policies/organizations/match_policy_test.rb
```
It does not happen when running the full test suite. It seems to be related to this issue and is fixable by one of the suggestions in that thread: https://github.com/rails/rails/issues/48468

I made note of this in slack. If we can replicate this error and it starts affecting other test files, we can implement the fix from that issue but otherwise I think we can ignore it for now.

### DIFF
--- a/app/models/concerns/authorizable.rb
+++ b/app/models/concerns/authorizable.rb
@@ -8,6 +8,7 @@ ADOPTER_PERMISSIONS = %i[
 ].freeze
 
 FOSTERER_PERMISSIONS = %i[
+  view_adopter_foster_dashboard
   create_adopter_foster_profiles
   manage_adopter_foster_profiles
 ].freeze
--- a/app/models/concerns/authorizable.rb
+++ b/app/models/concerns/authorizable.rb
@@ -7,6 +7,11 @@ ADOPTER_PERMISSIONS = %i[
   manage_adopter_foster_profiles
 ].freeze
 
+FOSTERER_PERMISSIONS = %i[
+  create_adopter_foster_profiles
+  manage_adopter_foster_profiles
+].freeze
+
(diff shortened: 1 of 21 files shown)


app/models/concerns/authorizable.rb
```diff
@@ -6,6 +6,11 @@
   manage_adopter_foster_profiles
 ].freeze
 
+FOSTERER_PERMISSIONS = %i[
+  create_adopter_foster_profiles
```
kasugaijin, line 10: can we also add `view_adopter_foster_dashboard` here, too? I just merged that in.

kasugaijin (approved):
LGTM! Just one comment to add a permission.

mononoken:
> LGTM! Just one comment to add a permission.

Thank you, @kasugaijin ! I've added the permission now.

Issue #622, opened 2024-04-03
Refactor top navbar


kasugaijin:
Remove avatar
Put relevant dropdown links (dashboard, sign out) in navbar as nav link

mehranjavid:
Hi @kasugaijin, Can i work on this issue?

mononoken:
Not sure if this would fall under the scope of this PR. However, I noticed today that if you are logged in as an adopter, the `Adopt` link in the navbar (which sends to user sign up form) results in an `ArgumentError`.

I feel like this link should change its path to the adoptable pets index when the user is already signed in.

We should probably add some authorization logic to prevent authenticated users from going to sign_up form (and probably sign_in) if it's going to raise an error too. I can create a separate issue for that if you agree.

kasugaijin:
Thanks @mononoken let’s keep that in this issue! I agree on your suggestions.

PR #642, opened 2024-04-13
622 refactor top navbar

kasugaijin:
# 🔗 Issue
https://github.com/rubyforgood/pet-rescue/issues/622

# ✍️ Description
Refactors top nav to replace links with relevant links and remove avatar and place the avatar dropdown links in navbar. 
Also fixes the error we were getting if a signed in user tries to access sign up page.

https://www.loom.com/share/8b171d6bc07f417e890e615f39c68cca?sid=db9f76dc-bb09-4448-9d76-6721aa1f0e87

# 📷 Screenshots/Demos

### DIFF
--- a/app/views/layouts/shared/_navbar.html.erb
+++ b/app/views/layouts/shared/_navbar.html.erb
@@ -19,75 +19,39 @@
         <%= active_link_to 'Sign Up', new_user_registration_path, class: 'btn btn-primary mx-2' %>
         <%= active_link_to 'Log In', new_user_session_path, class: 'btn btn-primary' %>
       <% end %>
-
-      <div class="dropdown ms-2 list-unstyled position-relative" data-controller="dropdown">
-        <button class="btn p-0 border-0" id="nav-dropdown-button" data-action="dropdown#toggle" data-dropdown-target="button" aria-expanded="false">
-          <% if user_signed_in? %>
-           <%= render_avatar_partial %>
-          <% end %>
-        </button>
-
-        <%# Use "aria-dropdown-menu" here to avoid the theme's default hover-open behavior for accessibility concerns. %>
-        <div class="aria-dropdown-menu position-absolute end-0" data-dropdown-target="menu" hidden aria-hidden="true" aria-labelledby="nav-dropdown-button" style='z-index: 1000'>
-          <% if user_signed_in? %>
-            <ul class="list-unstyled">
-              <% if current_user.adopter_foster_account %>
-                <li>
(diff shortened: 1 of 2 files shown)


app/views/layouts/shared/_navbar.html.erb
```diff
@@ -19,75 +19,39 @@
...
         <% if user_signed_in? %>
-          <li class="nav-item d-block d-lg-none">
-            <%= button_to 'Log Out', destroy_user_session_path, method: :delete, data: { turbo: false }, class: 'nav-link' %>
-          </li>
+          <% if current_user.adopter_foster_account %>
+            <li class='nav-item'>
+              <%= active_link_to 'Dashboard', adopter_foster_dashboard_index_path, class: 'nav-link' %>
+            </li>
+          <% end %>
+          <% if current_user.staff_account && !current_user.staff_account.deactivated? %>
```
mononoken, line 50: An alternative approach would be to utilize ActionPolicy here. That would keep authz logic out of the view. It usually works great for using policies to check logic like this because it can infer the correct policy from the record passed. However, since `dashboard` is not a record, it's a bit wordy in this case:
```ruby
<% if allowed_to?(:index?, with: Organizations::DashboardPolicy, context: {organization: Current.organization}) %>

```
This makes it so instead of writing a conditional for the user having the right attributes, we check simply if the user passes the policy rule and let the policy handle the logic.

app/views/layouts/shared/_navbar.html.erb
```diff
@@ -19,75 +19,39 @@
...
+          <%= active_link_to 'Contact', new_contact_path, class: 'nav-link' %>
         </li>
         <li class="nav-item">
           <%= active_link_to 'Donate', donate_path, class: 'nav-link' %>
         </li>
         <% if user_signed_in? %>
-          <li class="nav-item d-block d-lg-none">
-            <%= button_to 'Log Out', destroy_user_session_path, method: :delete, data: { turbo: false }, class: 'nav-link' %>
-          </li>
+          <% if current_user.adopter_foster_account %>
```
mononoken, line 45: See https://github.com/rubyforgood/pet-rescue/pull/642/files#r1564502741

In this case it would be:
```ruby
<% if allowed_to?(:index?, with: Organizations::AdopterFosterDashboardPolicy, context: {organization: Current.organization}) %>
```

mononoken (approved):
Looks good! Good catch on the devise method. Thanks, @kasugaijin !

app/views/layouts/shared/_navbar.html.erb
kasugaijin: Ah yes good call. I made these two changes!

Issue #664, opened 2024-04-24
Redirect adopter/fosterer to dashboard on login


mononoken:
## Description
When an adopter or fosterer logs in, we want them to redirect to their dashboard. Currently, they redirect to the adoptable pet index.

## Acceptance Criteria
- [x] Adopter is redirected to `adopter_foster_dashboard_index_path` on login
- [x] Fosterer is redirected to `adopter_foster_dashboard_index_path` on login
- [x] Staff is redirected to `dashboard_index_path`
- [x] Conditional utilizes policies (using `allowed_to?` helper)

PR #665, opened 2024-04-24
664 Redirect to dashboard

mononoken:
# 🔗 Issue
#664 

# ✍️ Description
This updates `after_sign_in_path_for` so that adopters and fosterers are redirected to their dashboard now upon login. It also updates the current if statement so that the staff dashboard redirect is using the `DashboardPolicy` instead of `PetPolicy` (which was old logic).


kasugaijin (approved):
Thanks @mononoken !

PR #666, opened 2024-04-24
659 Staff can CRUD FAQs from the UI

### LINKED ISSUE #659: Org FAQ: Enable Staff to CRUD FAQs via the UI.
This PR https://github.com/rubyforgood/pet-rescue/pull/657/files added the table and model for FAQs, which allows an Org to CRUD their own frequently asked questions/answers. Now we need to be able to handle this in the UI. Much of this implementation is very similar to how we handle the organization's Default Pet Tasks, so refer to that.

Note, we have an `order` column on FAQ, but we do not need to handle that in this PR.

Acceptance criteria:
- Add Controller with REST actions, and routes, scoped under Organization, authorize the actions using the policy.
- Add PolicyClass and a new Staff level permission `manage_FAQs`
- Add link in Dashboard side nav to 'FAQ'
- Add views to allow CRUD of an FAQ
- Show FAQs in an index table, where I can edit or delete an FAQ
- Tests

jmilljr24:
Resolves #659 

# ✍️ Description

This adds a new page for staff to CRUD FAQs. I added the ability to edit the FAQ inline with turbo_streams. The new form is a redirect for now. I could implement something with turbo however I got to thinking about when ordering is eventually implemented and potential issues without knowing the details.

There is also one minor change to the static FAQ route name since it was conflicting with the now implemented FAQ routing. I updated files where applicable except for the `sign_up_mailer` since the url is still hard coded as `"https://www.bajapetrescue.com/faq"` which I assume will be updated at a later time.

# 📷 Screenshots/Demos

[Screencast from 04-24-2024 09:12:23 AM.webm](https://github.com/rubyforgood/pet-rescue/assets/16829344/23ec1244-7710-4846-9ee1-73bca9dd6f8e)


app/controllers/organizations/faqs_controller.rb
```diff
@@ -0,0 +1,76 @@
+class Organizations::FaqsController < Organizations::BaseController
+  layout "dashboard"
+  before_action :context_authorize!, only: %i[index new create]
```
kasugaijin, line 3: Any reason to leave out Update and Destroy?

app/policies/organizations/faq_policy.rb
```diff
@@ -0,0 +1,10 @@
+class Organizations::FaqPolicy < ApplicationPolicy
+  pre_check :verify_organization!
+  pre_check :verify_active_staff!
+
+  alias_rule :new?, :create?, :index?, to: :manage?
```
kasugaijin, line 5: Same here, should we add update and destroy?

app/controllers/organizations/faqs_controller.rb
jmilljr24: I'm still trying to fully wrap my head around action policy, however I reviewed default_pet_task to get a better understanding and this is how it is set up there.

`context_authorize!` calls `authorize! DefaultPetTask,
      context: {organization: Current.organization}`

For show edit update destroy  uses `set_faq` which calls `authorize! @faq`

jmilljr24:
> Looks really good @jmilljr24 I also played around with it in the browser and works well. I like the use of turbo for destroy and update and agree we can leave create as is for now. Just one question about whether we want to include update and destroy in the policy.
> 
> For the static FAQ, I will soon make an issue to remove this, and create a new route, controller, view for unauthenticate users and adopters and fosterers to view. This will be in the top Navbar and use the collection of FAQs created here.

Maybe @mononoken can shed some light on this. I added a reply above. I spent quite a while trying to understand the action policy stuff better but I'm missing something.

kasugaijin:
> > Looks really good @jmilljr24 I also played around with it in the browser and works well. I like the use of turbo for destroy and update and agree we can leave create as is for now. Just one question about whether we want to include update and destroy in the policy.
> > For the static FAQ, I will soon make an issue to remove this, and create a new route, controller, view for unauthenticate users and adopters and fosterers to view. This will be in the top Navbar and use the collection of FAQs created here.
> 
> Maybe @mononoken can shed some light on this. I added a reply above. I spent quite a while trying to understand the action policy stuff better but I'm missing something.

Yes it makes sense now. When we want to set an instance eg for update or destroy show show we have the setter method so it makes sense to authorize there. And for new instances we have the other callback for new, etc.

app/policies/organizations/faq_policy.rb
jmilljr24: I think I have a better understand now of what's going on after reading the docs a few times. ActionPolicy::Base sets `manage?` as default so that it what `update` and `destroy` use. `new?, create?, index?` are already defined in the superclass by default so the alias is used to set them to `manage?` here.

[ActionPolicy Aliases](https://actionpolicy.evilmartians.io/#/aliases)

kasugaijin: Thanks @jmilljr24 I too had to check the docs out for that. Not obvious at first!

kasugaijin (approved):

mononoken:
Looks great! Sounds like you figured out the confusion with the defaults in Action Policy. Let me know if you need clarification on anything else.

Since the default rules get used frequently, maybe it would be beneficial to show them in our `ApplicationPolicy`. We could show them in a comment. And, if we decided to overwrite any of these defaults, we could uncomment them:
```ruby
# Action Policy defaults
# default_rule :manage?
# alias_rule :new?, to: :create?
# def index?() = false
# def create?() = false
# def manage?() = false
```

jmilljr24:
@kasugaijin Would you like me to push up a commit with the comments suggested before this is merged?

kasugaijin:
> @kasugaijin Would you like me to push up a commit with the comments suggested before this is merged?

Yes I think that is a great idea @mononoken . Auth is a layer most devs will need to touch, so the more we can help in this regard the better. We even have a readme.md file in our `app/policies` folder at work for that reason. We don't have to go that far, but maybe add a line in the comments with a link to the Action Policy documentation, too please @jmilljr24 .

Issue #668, opened 2024-04-27
Org FAQ: Enable users to see an org's FAQs


kasugaijin:
As a site user (authenticated or not) I should be able to see a link in the top navbar to FAQs and click it so see a list of an org's FAQs.

PR #673, opened 2024-04-28
Add UI for users to see an org's FAQ

kasugaijin:
# 🔗 Issue
https://github.com/rubyforgood/pet-rescue/issues/668

# ✍️ Description
This adds the UI for users to see an org's FAQ regardless of authentication status. 

# 📷 Screenshots/Demos
https://www.loom.com/share/685c8aede60441bdb5b4dfbfe166edba?sid=4bf552e0-956f-4d5d-b923-3edb8633aba7


app/views/organizations/public_faq/index.html.erb
```diff
@@ -0,0 +1,39 @@
...
+    </div>
+  </div> <!--container-->
+</header>
+
+<section class="pb-5" id="public_faq">
+  <div class="container">
+    <div class="row justify-content-center">
+      <div class="col-lg-8 col-md-10 col-12">
+        <div class="accordion accordion-flush" id="accordionExample">
+          <% @faqs.each_with_index do |faq, index| %>
```
jmilljr24, line 16: This all works well, just wondering when the index would come in to play. If index is not needed, could we simplify this and render as a collection `render partial: "faq", collection: @faqs`. The new `public_faq/_faq` partial could then use the dom_id helper. `id="<%= dom_id(faq, :heading) %>"`

Other small nitpick, maybe change "accordionExample" and "#accordionExample" with something more specific.

jmilljr24 (changes requested):
Just a small nitpick and a question. Otherwise looks good!

app/views/organizations/public_faq/index.html.erb
kasugaijin: @jmilljr24 yes rendering partials would make more sense. The index is required for the JS to work so that each FAQ Q and A id matches but is different from the rest. I have not see the dom_id helper but obviously there is something like this as it's a common use case. Good catch!

jmilljr24: Sorry I should have highlighted this line `<% faq_id = "faq_#{faq.id || index + 1}" %>`. The faq should always have an id so the `index + 1` would ever be used.

kasugaijin: Yes I see. Haha I really biffed this one. I rushed it. I will fix it.

kasugaijin: @jmilljr24 updated to use a question and answer component that we can use elsewhere.

app/views/components/_question_and_answer.html.erb
```diff
@@ -0,0 +1,13 @@
...
+  <h3 class="mb-0 fw-bold">
+    <a href="#" class="d-flex align-items-center text-inherit active collapsed" data-bs-toggle="collapse" data-bs-target="#collapse_<%= faq.id %>" aria-expanded="false" aria-controls="collapse_<%= faq.id %>">
+      <span class="me-auto"><%= faq.question %></span>
+      <span class="collapse-toggle ms-4">
+          <i class="fe fe-plus text-primary"></i>
+        </span>
+    </a>
+  </h3>
+</div>
+<div id="collapse_<%= faq.id %>" class="collapse" aria-labelledby="heading_<%= faq.id %>" data-bs-parent="#accordionExample">
```
jmilljr24, line 11:
```suggestion
<div id="collapse_<%= faq.id %>" class="collapse" aria-labelledby="heading_<%= faq.id %>" data-bs-parent="#accordion-public-faq">
```

jmilljr24 (approved):
I'll do a better job of commenting on the applicable lines so this is my fault. There is another lingering "accordionExample" that I should have been more explicit about in my initial review. Feel free to change if you want but I don't think it affects anything.

Otherwise it looks good to me!

app/views/components/_question_and_answer.html.erb
kasugaijin: Ah yes. I changed that also!

jmilljr24 (approved):
lgtm

Issue #171, opened 2023-09-06
Display of an Organization's information


jadekstewart3:
When I visit an organization's root page I see in the footer (replacing the default content that is currently in the footer):
- The organization's address
- The organization's phone number

![image](https://github.com/rubyforgood/pet-rescue/assets/95949082/1eb835d9-417c-4dda-ba83-851999b2647a)

sean-dickinson:
@jadekstewart3 I'd be happy to take this one. Just looking into it and I'm wondering what should be shown for an address? It looks like the `Location` model doesn't actually have a street address just a city, state/province, zipcode and country.

jmilljr24:
@sean-dickinson All yours!

>It looks like the `Location` model doesn't actually have a street address just a city, state/province, zipcode and country.

City, state/province, zipcode and country is good. `organization_profile` model has the phone number. Let me know if you have any more questions.

PR #680, opened 2024-05-01
feat: replace dummy contact info with org info

sean-dickinson:
Adds the organization contact information to the footer partial

# 🔗 Issue
https://github.com/rubyforgood/pet-rescue/issues/171

# ✍️ Description
- Replaces the dummy contact info in the footer with the org's contact info

# 📷 Screenshots/Demos
image

### DIFF
--- a/app/views/layouts/shared/_footer.html.erb
+++ b/app/views/layouts/shared/_footer.html.erb
@@ -76,9 +76,9 @@
         <!-- contact info -->
         <div class="mb-4">
           <h3 class="fw-bold mb-3">Get in touch</h3>
-          <p>339 McDermott Points Hettingerhaven, NV 15283</p>
-          <p class="mb-1">Email: <a href="#">support@geeksui.com</a></p>
-          <p>Phone: <span class="text-dark fw-semibold">(000) 123 456 789</span></p>
+          <p><%= Current.tenant.location.city_town %>, <%= Current.tenant.location.province_state %> <%= Current.tenant.location.country %></p>
+          <p class="mb-1">Email: <a href="mailto:<%= Current.tenant.profile.email %>"><%= Current.tenant.profile.email %></a></p>
+          <p>Phone: <span class="text-dark fw-semibold"><%= Current.tenant.profile.phone_number %></span></p>
         </div>
       </div>
     </div>
--- a/app/views/layouts/shared/_footer.html.erb
+++ b/app/views/layouts/shared/_footer.html.erb
@@ -78,7 +78,9 @@
           <h3 class="fw-bold mb-3">Get in touch</h3>
           <p><%= Current.tenant.location.city_town %>, <%= Current.tenant.location.province_state %> <%= Current.tenant.location.country %></p>
           <p class="mb-1">Email: <a href="mailto:<%= Current.tenant.profile.email %>"><%= Current.tenant.profile.email %></a></p>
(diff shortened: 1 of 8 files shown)


test/system/home_page_test.rb
```diff
@@ -8,9 +8,17 @@ class HomePageTest < ApplicationSystemTestCase
   end
 
   test "renders custom hero and about text from PageText or default text" do
-    PageText.create(hero: "Super Pets for a good paws", about: "All about us")
+    create(:page_text, hero: "Super Pets for a good paws", about: "All about us")
+
     visit home_index_path
     assert_text "Super Pets for a good paws"
     assert_text "All about us"
+
```
sean-dickinson, line 16: I wasn't sure if this was the best place to test this. Happy to put it elsewhere if there's a better place for it.

app/views/layouts/shared/_footer.html.erb
```diff
@@ -76,9 +76,9 @@
         <!-- contact info -->
         <div class="mb-4">
           <h3 class="fw-bold mb-3">Get in touch</h3>
-          <p>339 McDermott Points Hettingerhaven, NV 15283</p>
-          <p class="mb-1">Email: <a href="#">support@geeksui.com</a></p>
-          <p>Phone: <span class="text-dark fw-semibold">(000) 123 456 789</span></p>
+          <p><%= Current.tenant.location.city_town %>, <%= Current.tenant.location.province_state %> <%= Current.tenant.location.country %></p>
+          <p class="mb-1">Email: <a href="mailto:<%= Current.tenant.profile.email %>"><%= Current.tenant.profile.email %></a></p>
+          <p>Phone: <span class="text-dark fw-semibold"><%= Current.tenant.profile.phone_number %></span></p>
```
jmilljr24, line 81: What are your thoughts on formatting the phone number? 

```suggestion
          <p>Phone: <span class="text-dark fw-semibold"><%= Current.tenant.profile.formatted_phone %></span></p>
```

Screenshot from 2024-05-01 19-36-28

We have a method in the `adopter_foster_profile` model you could reference and add to `organization_profile` model

jmilljr24 (changes requested):
Thanks for working on this! It looks good. I think we can go without testing for this unless some else thinks its prudent. 

Do you think you could have the phone number formatted to make it a bit easier to read? We use the Phonelib gem which can do most of the work. Let me know what you think!!

test/system/home_page_test.rb
sean-dickinson: Yeah that makes sense, but I always worry about adding extra cases for a system test just given how slow they are comparatively. I think you're right that there's not much benefit here

app/views/layouts/shared/_footer.html.erb
sean-dickinson: Sounds good, I'll take a look and pull it into a concern for reuse

sean-dickinson: Hey @jmilljr24 I went ahead and refactored this into a concern. I included the `before_save` normalizing in there as well. Let me know what you think.

jmilljr24:
Awesome job! Thanks @sean-dickinson 

Can you please run `rails standard:fix` then we will be all set.

jmilljr24:
@all-contributors
please add @sean-dickinson  for code.

jmilljr24 (approved):
lgtm!

Issue #678, opened 2024-05-01
Rename foster_application_reviews to manage_fosters


kasugaijin:
We currently have a route, controller, and views for `foster_application_reviews`. This naming is not accurate and needs to be changed to `manage_fosters`

Acceptance Criteria: 
- change route, controller, view
- update tests so they pass

LindseyB:
Can you assign this to me?

kasugaijin:
Done @LindseyB

PR #683, opened 2024-05-03
Rename `FosterApplicationReviews` => `ManagerFosters`

LindseyB:
# 🔗 Issue
Closes 

* https://github.com/rubyforgood/pet-rescue/issues/678

# ✍️ Description

Fairly straightforward rename. I started with the controller and the controller file name. Additionally, there's the associated views folder. It's not clear to me where all of these views are actually used, but there's not much there. I also updated the associated routes. 

Something I wanted to do, but was struggling with was just some basic test coverage for this controller just to ensure it renders, but I realized upon further digging because of how this is currently setup the page gives an error: **Action 'organizations/foster_application_reviews#index' hasn't been authorized** so even with the rename there's not a good way to test this since it will actually error. 

# 📷 Screenshots/Demos

Since it was just a rename I don't think there's anything worth screenshoting here.


jmilljr24:
@all-contributors
please add @LindseyB  for code.

jmilljr24 (approved):
Looks good! Thanks @LindseyB 

Just for context, Fosters are not fully built out so the authorization is not in place yet.

Issue #636, opened 2024-04-09
Invite fosterers


mononoken:
## Description
Fosterers are invited directly by staff. They do not sign up like adopters, who fill out forms for their account.

It is possible for a user to be both an adopter and a fosterer. This issue is only concerned with the case where a user does not yet exist and is being invited as a new user. A different flow will be used for existing users becoming fosterers.

A fosterer exists as a `User` with the `fosterer` role. They should also have an `AdopterFosterAccount`.

This will follow a similar flow to how users are currently invited to become a new staff user. See `./alta/users/invitation/new` path that is accessed from the staff dashboard link `Staff` on the sidebar. The role selection will not be required for this new form, as it will be auto-assigned as `fosterer`.

## Acceptance Criteria
- [ ] Create `Organizations::FosterInvitationsController` (use `Organizations::InvitationsController` as a guide)
- [ ] Views:
  - [x] Add dashboard sidebar link for `Fosterers` (add it above "Staff")
  - [x] Above link will redirect to a fosterers index (Note: these views for fosterers should mostly mimic the `Staff` link also found on dashboard sidebar)
  - [ ] Index should have a link to a form for inviting a new fosterer
- [ ] Create `Organizations::FosterInvititationPolicy` to authorize controller actions
  - [ ] Add `invite_fosterer` permission to `STAFF_PERMISSIONS` and any others as necessary to `authorizable.rb`
- [ ] Staff can fill a form to send invitation to fosterer by email
  - [ ] Form should not include role select. The role should be assigned in the controller.
- [ ] Fosterer receives an email inviting them to create a fosterer account
- [ ] Invited user should have:
  - [ ] User record with form attributes
  - [ ] Assigned the `fosterer` role using `rolify`
  - [ ] Associated `AdopterFosterAccount` is built for the User
- [ ] Tests
  - [ ] Unit test new policy class
  - [ ] Controller test for implementation of authorization in controller

## Additional Info
![fosterers-index](https://github.com/rubyforgood/pet-rescue/assets/81536479/a393bacb-2258-4c10-bd12-678f097c163e)
![fosterer-invite](https://github.com/rubyforgood/pet-rescue/assets/81536479/52822df6-a56a-45d2-90c3-c1762764eddc)

There are many components to this issue, but they should mostly be copies of existing actions. If there are any inconsistencies or confusion in this issue, please bring them to our attention here or on Slack!

kasugaijin:
Hey @mononoken to answer the Questions:
[...]

PR #684, opened 2024-05-03
636 invite fosterers

mononoken:
I have the invitation functionality working, but I have some parts such as the authz unfinished. However, before I go further, I wanted to make sure other's are happy with the approach I took in regards to the invitations.

# 🔗 Issue
#636 

# Summary of work
There were issues setting this up with devise with what I initially thought this would look like. I was not able to get devise_invitable to work with multiple invitation controllers, so I instead have one main controller hooked up to the gem with multiple sub controllers and forms pointing to this one. The main controller requires conditionals to account for the multiple paths within it. You may be able to follow along just with the comments I added to this PR, but if you need more info I have much more below. I have some alternatives I thought of at the bottom too.

# ✍️ Description
This PR will allow staff to invite fosterers. Fosterers will receive an email to join the app, similar to how staff are invited to join the application.

There was a lot of pain with devise setting this up. Initially, I thought I would rename `Organizations::InvitationsController` to `Organizations::StaffInvitationsController` and then copy it as `Organizations::FostererInvitationsController`. However, devise_invitable does not support using multiple controllers.

I did try to work around it and force the devise routing and helpers to play nicely with multiple invitation controllers, but I was not able to get it to work. It's probably skill issues, but I found it a huge pain trying to debug the errors it was coming up with when trying to setup both controllers through devise_invitable. I was also not able to find any examples online where someone had setup multiple controllers to inherit from the devise_inivitables default.

This led me to this solution. Instead of having two invitation controllers, we have a main controller `InvitationsController` that is wired up with devise_invitable. Then we have two other controllers for fosterers and staff invitations that only have `new` actions. This allows us to have separate forms with separate authorization for each. However, ultimately both of these forms are posting to the main `InvitationsController#create` action.

Another way to think of it is that `InvitationsController` is the controller for creating a `User` via an invitation. The secondary controllers `FostererInvitationsController` and `StaffInvitationsController` simply act as middle-men to generate the correct form.

Thus, we have to account for multiple paths within the `create` action. Right now, that's fairly simple with a case statement with two happy paths. Hopefully, we don't have to extend it too much in the future or else this could get messy.

The devise routes generated with `devise_for` don't know what we are doing of course, so they are still generating a route for the default devise_invitable `new` action. That route is no longer used since we have these two sub-controllers taking over `new`. To account for this, I made the policy for the `InvitationsController` always deny, so no user can access that action. The alternative would be to manually generate all the devise routes, as I don't think there is a way to disable single routes within it. So that would look something like:
```ruby
devise for :users,
  skip: [:invitations],
  controllers: {
    sessions: "sessions path",
    registration: "registration path"
  }
get/patch/put/post for every devise_invitable route except new
```

Some alternative approaches I thought of:
1. Instead of multiple controllers, all logic lives in the `InvitationsController`, and we edit the form with different form partials that are called based on conditionals. I do not like this approach, since it places too much logic in the views.
2. Similar to 1, except instead of partials we use scoped views. I think this would work. However, as I understand from the docs, I think the intention of scoped views is for when you have multiple devise models, which we do not.
3. Write our own Invitations layer. Not ideal for time, and it would require thorough work intermingling with the authentication layer.

As far as alternatives go, I think 2 is the most promising, but I don't think we gain much from it either. I am also hesitant to use a tool for something it doesn't sound like it was intended for. And finally, I think it may make the authorization layer messier for this area.

I would love it if someone has a cleaner idea too!

# 📷 Screenshots/Demos
[Demo of fosterer invitation flow](https://vimeo.com/942792787?share=copy)


app/controllers/organizations/invitations_controller.rb
```diff
@@ -3,29 +3,50 @@ class Organizations::InvitationsController < Devise::InvitationsController
...
-    @user = User.new(
-      user_params.merge(password: SecureRandom.hex(8)).except(:roles)
-    )
-    @user.add_role(user_params[:roles], Current.organization)
-    @user.staff_account = StaffAccount.new
-
-    if @user.save
-      @user.invite!(current_user)
-      redirect_to staff_index_path, notice: "Invite sent!"
+    case user_params[:roles]
```
mononoken, line 15: We use a case statement here to account for the different authorization paths necessary for the action as well as the separate results.

Fosterers can be invited by staff members but only an admin can invite staff and other admins.

app/controllers/organizations/invitations_controller.rb
```diff
@@ -3,29 +3,50 @@ class Organizations::InvitationsController < Devise::InvitationsController
...
+
+      if @user.save
+        @user.invite!(current_user)
+        redirect_to fosterers_path, notice: "Invite sent!"
+      else
+        render :new, status: :unprocessable_entity
+      end
     else
-      render :new, status: :unprocessable_entity
+      :deny
```
mononoken, line 49: This is a placeholder. We will want to redirect here with an error message in the event that someone is trying to post an invitation with any roles other than the ones in the cases above.

app/policies/organizations/invitation_policy.rb
```diff
@@ -3,6 +3,6 @@ class Organizations::InvitationPolicy < ApplicationPolicy
   pre_check :verify_active_staff!
 
   def create?
-    permission?(:invite_staff)
+    deny!
```
mononoken, line 6: This policy is setup to always deny where it is called. This is acceptable because the other `FostererInvitationPolicy` and `StaffInvtationPolicy` are used where necessary in the `InvitationsController`. This will also make sure anyone trying to go to the general invitations form page that is auto-generated by devise, will not be able to access the page.

The other actions such as edit are not authorized (since they mostly don't even have an authenticated user yet) and Action Policy is currently setup to not require authorization for Devise controllers.

app/views/organizations/fosterer_invitations/_form.html.erb
```diff
@@ -0,0 +1,32 @@
+<!-- form -->
+<%= bootstrap_form_with model: user, url: invitation_path(user) do |form| %>
```
mononoken, line 2: This form is called from the `FosterersInvitationsController` but it is posting to the `invitation_path(user)` which directs to `InvitationsController`, where devise_invitable is setup.

kasugaijin:
Hey @mononoken I sympathize with your pain haha. I have had to do similar things with Devise and ended up having to manage multiple user types in a single flow like this. I think your implementation is good and happy with continuing on with it. I'd just recommend adding a couple comments in each controller (and possibly the policies) to mention this is how we are handling it.

mononoken:
This should be ready for review now! I added a demo of the work at the bottom of the PR too.

app/policies/organizations/invitation_policy.rb
kasugaijin: Makes sense!

app/controllers/organizations/staff/invitations_controller.rb
```diff
@@ -3,29 +3,57 @@ class Organizations::Staff::InvitationsController < Devise::InvitationsControlle
...
     @user = User.new
-    @staff = StaffAccount.new(user: @user)
   end
 
   def create
-    authorize! StaffAccount, context: {organization: Current.organization},
-      with: Organizations::InvitationPolicy
+    # We have different return paths based on the roles provided in the params.
+    # If you extend this, make sure new paths have their own authz!
+    case user_params[:roles]
```
kasugaijin, line 18: Just a thought here - it may not work (but I think it should). But, I wonder for each new controller `FostererInvitationsController` and `StaffInvitationsController` if they instead inherited from the `InvitationsController` then we could use the same `#new` logic, and for `#create` we could put most (or all( of this case this logic in each respective controller (or call `super` if it made sense to pass some of the work to the parent action)? It might be a way for us to keep the logic in these controllers rather than have it split between the three..if that makes sense? I don't think there is an issue in inheriting from `InvitationsController` over `Organizations::BaseController`

kasugaijin (approved):
@mononoken I think this implementation exposes the limits of this gem for our given use case. Without looking in depth, it looks like the gem works best when user types are differentiated with devise resources, rather than a single devise resource (User) with roles. But, we can still shoehorn it to fit our needs. I tried to think of a different way to handle it that might be better, but I think what you have is reasonable, and I like it. I am approving now, but if you want to hold off merging until Tue meeting to see if we get any other ideas, feel free. Your call!

app/views/organizations/staff/staff/index.html.erb
```diff
@@ -19,7 +19,7 @@
           </div>
         </div>
       </div>
-      <%= link_to "Invite Staff", new_user_invitation_path, class: "btn btn-primary" %><br>
+      <%= link_to "Invite Staff", new_staff_staff_invitation_path, class: "btn btn-primary" %><br>
```
nsiwnf, line 22: Could we change this to `/staff/invitations/new` or will that be annoying to refactor?

nsiwnf (approved):
a little late to the review, but lgtm! I think any clean-up or re-org ideas can be done later if needed

app/views/organizations/staff/staff/index.html.erb
kasugaijin: Hmm yes it is a bit of a mouthful. I would vote we keep the `StaffInvitationsController` name, but just amend the path in the routes to be new_staff_invitation_path

mononoken:
Thanks y'all for the reviews @kasugaijin and @nsiwnf ! I agree that it would be nice to clean up some of the new paths that changed with the namespacing recently. Another one I didn't like using was `new_staff_fosterer_invitation_path`. I think it will be a good topic for the next meeting.

Issue #677, opened 2024-05-01
Code Organization: Add namespaces for staff and adopter_fosterer


kasugaijin:
The number of routes, controllers, policies and views scoped under `organization` is growing and becoming a little disorganized. This is reflected in some of the less than ideal naming of things. Therefore, we would benefit from some code organization through namespacing. See the code snippet below for what needs to go where.

Acceptance Criteria:
- Add namespaces for `staff` and `adopter_fosterer`
- Scope applicabled controllers, views, and policies for each user type under each namespace
- Public routes are not namespaced
- Update naming in tests, and wherever else it needs to be updated, reflecting the new namespace

Example of namespacing the routes
```
scope module: :organizations do
    # public routes 
    resources :home, only: [:index]
    resources :adoptable_pets, only: [:index, :show]
    resources :public_faq, only: [:index] # (change to :faq)

    namespace :staff do
      resource :organization_profile, only: %i[edit update]
      resource :page_text, only: [:edit, :update]
      resources :profile_reviews, only: [:show]

      resources :pets do
        resources :tasks
        post "attach_images", on: :member, to: "pets#attach_images"
        post "attach_files", on: :member, to: "pets#attach_files"
      end

      resources :default_pet_tasks
      resources :faqs
      resources :dashboard, only: [:index]
      resources :matches, only: %i[create destroy]

      resources :adoption_application_reviews, only: [:index, :edit, :update]
      resources :foster_application_reviews, only: [:index]
      resources :staff do
        post "deactivate", to: "staff#deactivate"
        post "activate", to: "staff#activate"
        post "update_activation", to: "staff#update_activation"
      end
    end

    namespace :adopter_fosterer do
      resource :adopter_foster_profile, except: :destroy, as: "profile" # change to :profile
      resources :adopter_foster_dashboard, only: [:index] # change to :dashboard
      resources :adopter_applications, path: "applications", only: %i[index create update]
    end    
  end
```

MooseCowBear:
I'd like to work on this

kasugaijin:
All yours @MooseCowBear - let me know if you need more details!

PR #685, opened 2024-05-03
677 add namespaces for staff and adopter fosterer

MooseCowBear:
# 🔗 Issue

#677 

# ✍️ Description

Updates routes to use staff and adopter_fosterer namespaces. Add directories for each namespace to controllers, views, and controller tests, and updates path/url helpers throughout. Also updates `locale/en.yml` to incorporate namespaces. 

I didn't split the integration tests into staff, adopter_fosterer directories. Wasn't sure how these are best organized. Let me know. I'm happy to make changes. 

# 📷 Screenshots/Demos


config/routes.rb
```diff
@@ -9,38 +9,45 @@
...
-      post "attach_files", on: :member, to: "pets#attach_files"
-    end
-    resources :default_pet_tasks
-    resources :faqs
-    resources :public_faq, only: [:index]
-    resources :dashboard, only: [:index]
-    resources :adopter_foster_dashboard, only: [:index]
-    resources :adopter_applications, path: "applications",
-      only: %i[index create update]
     resources :forms do
```
kasugaijin, line 15: Ah I think I missed this one in my ticket example. This should be in the staff namespace - it's for CRUDing the org forms and questions.

app/controllers/organizations/invitations_controller.rb
kasugaijin, line 1: This should also be under `staff` namespace

kasugaijin:
Thanks @MooseCowBear! I think we can leave the integration tests as is for now. I will check this branch out this weekend and have a good look.

config/routes.rb
```diff
@@ -60,8 +67,4 @@
   get "/cookie_policy", to: "static_pages#cookie_policy"
 
   resources :contacts, only: [:new, :create]
```
kasugaijin, line 69: I was thinking this could be scoped under Org - but is beyond the scope of this PR as it would require some logic changes, but being scoped there, the current org will be set for the context of the mailer. Though, I was in two minds whether this could also be used to create contacts for the non-org root i.e. from orgs who have questions about this app that comes to us devs. Not something to dwell on here. Just leaving a non-blocking thought.

kasugaijin (approved):
This looks really good @MooseCowBear thank you. It is a breath of fresh air now to see things scoped/organized better. What do you think?

I see we could do some cleanup in the views. Some things are not doing anything e.g., `organizations/settings` and the components and partials could be updated with naming to reflect the new namespaces e.g., rename `dashboard` to `staff_dashboard` but they can be done in another task.

MooseCowBear:
@kasugaijin Yes, I think the namespaces were a good idea.

config/routes.rb
MooseCowBear: @kasugaijin One thing we might want to consider for orgs contacting devs is allow them to add tags like question or bug or feature request or important, etc. to make it easier to triage any issues they are having -- maybe not as a top priority but at some point down the line. Which might be a reason to have separate routes/controllers.

kasugaijin: Yes that's a great idea! @MooseCowBear

PR #690, opened 2024-05-04
Authenticated users can delete their avatar

### LINKED ISSUE #681: Purging Attachments: Create a separate flow for purging user profile avatars
Staff can upload attachments and purge them via the `AttachmentsController` with associated policy that checks the org, and staff is active in `ActiveStorage::AttachmentPolicy` - this works fine for all attachments that staff manage under an org.

We also allow every authenticated user (Staff, Adopters, Fosterers) to add an avatar on their profile via the edit user registration path. Deleting of this attachment should be handled in its own flow. I am open to ideas, but I think a new controller (e.g. `PurgeAvatarController`) using the `destroy` action that also uses a new policy (e.g., `PurgeAvatarPolicy`) that checks the user is authenticated and deleting the attachment associated with their user account (not someone elses). With respect to scoping this route/controller/view, it should be accessible to all authenticated users, `staff` and `adopters` and `fosterers` so it won't belong in either of the staff or adopter_fosterer namespaces. So for now, just leave it outside the namespaces and we can figure out how to organize it later (the namespacing issue is currently in progress).

If you think this can be handled better than above after looking at it, let me know!

Acceptance criteria:
- Add new controller, route 
- Add new policy
- Adopter/fosterer can delete their own avatar, not someone elses
- Tests

jmilljr24:
# 🔗 Issue

Resolves #681 

# ✍️ Description

This adds a new method to the attachments controller to purge avatar's.
- Authorization to purge_avatar added to adopters, fosterers, staff
- `purge_avatar` method in the attachment controller prevents user from deleting avatar that is not theirs
-  `AttachmentsController` was moved out of Staff to Organizations
- "staff/" was added to the `staff_purge_attachment` url to match the staff namespace
- Tests added 

# 📷 Screenshots/Demos

[Screencast from 05-04-2024 11:14:41 AM.webm](https://github.com/rubyforgood/pet-rescue/assets/16829344/fd9047bf-2913-48d6-ae14-075157d33d0b)


kasugaijin (approved):
Looks good to me @jmilljr24 !

Issue #672, opened 2024-04-28
Fix: Pet settings dropdown on staff "Pets" page are cut off when only one pet is displayed


jmilljr24:
When only one pet is shown on the table, the setting's dropdown is cut off at the bottom so not all of the options are displayed.

The settings dropdown should display the same as when multiple pet's are in the table.

**Fix**
![Screenshot from 2024-04-26 16-26-07](https://github.com/rubyforgood/pet-rescue/assets/16829344/d6d6d021-de1e-45fb-a0c0-312216388ffb)

**Correct**
![Screenshot from 2024-04-26 16-27-25](https://github.com/rubyforgood/pet-rescue/assets/16829344/7625424c-6225-4977-b733-251a12e19907)

OliverLeighC:
Could I work on this?

jmilljr24:
@OliverLeighC All yours! Let me know if you have any questions.

OliverLeighC:
Looks like it's "expected behavior" for responsive tables 
https://github.com/tabler/tabler/issues/1343#issuecomment-1308569481

but I am going to see if I can get the suggested popper config fix to work

OliverLeighC:
Hmm... setting popper config to {strategy: "fixed"} reveals the full menu but doesn't behave nicely when scrolling 

https://github.com/rubyforgood/pet-rescue/assets/51760107/571bbcfc-206a-454f-a411-786570443ce5

Will keep investigating

jmilljr24:
> Hmm... setting popper config to {strategy: "fixed"} reveals the full menu but doesn't behave nicely when scrolling
> Clip.mov
> 
> Will keep investigating

I really appreciate you playing around with this. If none of the suggestions that you linked to give good results or if there will be some compromises, post an update and the team can discuss it.

jmilljr24:
@OliverLeighC Setting a  mini-height of the table might be the easiest option. The dropdown doesn't have a lot cut off so it might work out ok to set a min-height and adjust table headers to fit. But whatever works best!

OliverLeighC:
here is what it looks like with the min-height: 
<img width="1525" alt="image" src="https://github.com/rubyforgood/pet-rescue/assets/51760107/ec7cd3a7-a004-4311-afc9-087defc86325">
<img width="1534" alt="image" src="https://github.com/rubyforgood/pet-rescue/assets/51760107/092442f9-3bcf-4a0b-bcc5-bd63512f8c62">

more rows for comparison: 
[...]

PR #697, opened 2024-05-09
Issue 672 Pet Settings Dropdown

OliverLeighC:
# 🔗 Issue

https://github.com/rubyforgood/pet-rescue/issues/672

# ✍️ Description

Turns out it is a known issue with the responsive table that dropdowns get cut off (https://github.com/tabler/tabler/issues/1343#issuecomment-1308569481) 

I tried the popper config solution but it led to undesirable behavior when scrolling so I opted for setting the min height to give room for the dropdown to be displayed. 

# 📷 Screenshots/Demos

image
image
image
image

https://github.com/rubyforgood/pet-rescue/assets/51760107/534650b1-01bf-42f2-9b08-65ca117e4326


app/views/organizations/staff/pets/index.html.erb
```diff
@@ -42,7 +42,7 @@
       <div class="card">
         <!-- table -->
         <div class="table-responsive overflow-y-hidden">
-          <table class="table mb-0 text-nowrap table-hover table-centered">
+          <table style="height: 180px;" class="table mb-0 text-nowrap table-hover table-centered">
             <thead>
               <tr>
                 <th scope="col">Name</th>
```
jmilljr24, line 48:
```suggestion
          <table class="table mb-0 text-nowrap table-hover table-centered">
            <thead>
              <tr style="height: 70px;">
                <th scope="col">Name</th>
```

I'm thinking this will give just enough room for the dropdown with having the least amount of changes to the design. Thoughts?

jmilljr24 (changes requested):
Thanks for figuring out what the issue was! It looks like with the height applied to the table the single pet row height quite a bit larger. This is definitely betting than not being able to use the dropdown, but it does look a little jarring. 

If you don't mind, I think making the height of the row containing the header information slightly bigger will allow the dropdown to display fully even if it overlaps the header a bit. Let me know if I'm missing something.

app/views/organizations/staff/pets/index.html.erb
OliverLeighC: Good idea, 
image
image

jmilljr24:
@all-contributors
please add @OliverLeighC  for code.

jmilljr24 (approved):

jmilljr24:
@OliverLeighC looks good! Thank you

Issue #706, opened 2024-05-14
Adoption Applications: Add pagination.


kasugaijin:
We added pagination to `/staff/pets` and now we should add it to `/staff/adoption_application_reviews`.

I think that 5 per page would be good? Open to other ideas here as well. 

Refer to this PR for how it was done on staff pet index: https://github.com/rubyforgood/pet-rescue/pull/649

Acceptance criteria:
- Adoption application review page shows paginated cards, 5 per page
- No need to add to seed files as we have enough for more than one page already
- No need for test

atbalaji:
@kasugaijin I would love to work on this issue

kasugaijin:
@atbalaji all yours! No rush! Thank you

atbalaji:
/staff/adoption_application_reviews in this page i am getting errors due to existence of adopter_foster_accounts with no adopter_foster_profile in pre populated data. I want to know your thoughts on handling this?

kasugaijin:
> __/staff/adoption_application_reviews_ in this page i am getting errors due to existence of adopter_foster_accounts with no adopter_foster_profile in pre populated data. I want to know your thoughts on handling this?

You are working off the latest main branch, right?

atbalaji:
Can we close this issue?

PR #710, opened 2024-05-14
706 add pagination adoption application reviews

atbalaji:
# 🔗 https://github.com/rubyforgood/pet-rescue/issues/706

# ✍️ Description
-   Fixed bug in alta and baja seed data 
-   Added pagination to adoption application reviews and fixes failed test

# 📷 Screenshots/Demos
Screenshot 2024-05-15 043616

### DIFF
--- a/app/controllers/organizations/staff/adoption_application_reviews_controller.rb
+++ b/app/controllers/organizations/staff/adoption_application_reviews_controller.rb
@@ -21,7 +21,7 @@ class Organizations::Staff::AdoptionApplicationReviewsController < Organizations
       @pets_with_applications = filter_by_application_status(@pets_with_applications, status_filter)
     end
 
-    @pagy, @pets_with_applications = pagy(@pets_with_applications, items: AdopterApplication::REVIEWS_PER_PAGE)
+    @pagy, @pets_with_applications = pagy(@pets_with_applications, items: 10)
   end
 
   def edit
--- a/app/controllers/organizations/staff/adoption_application_reviews_controller.rb
+++ b/app/controllers/organizations/staff/adoption_application_reviews_controller.rb
@@ -1,6 +1,8 @@
 class Organizations::Staff::AdoptionApplicationReviewsController < Organizations::BaseController
   before_action :set_adopter_application, only: %i[edit update]
 
+  include ::Pagy::Backend
+
   layout "dashboard"
 
   def index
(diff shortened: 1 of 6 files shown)


app/controllers/organizations/staff/adoption_application_reviews_controller.rb
```diff
@@ -18,6 +20,8 @@ def index
       status_filter = params[:q]["adopter_applications_status_eq"]
       @pets_with_applications = filter_by_application_status(@pets_with_applications, status_filter)
     end
+
+    @pagy, @pets_with_applications = pagy(@pets_with_applications, items: AdopterApplication::REVIEWS_PER_PAGE)
```
kasugaijin, line 24: Can we just set the reviews per page in the controller instead? I see you have put it on the model as a constant, but this is a view related concept and does not really have anything to do with the core domain logic of an AdopterApplication, if you get me? I think specifying it on the controller is preferable.

app/models/adopter_application.rb
```diff
@@ -42,6 +42,8 @@ class AdopterApplication < ApplicationRecord
       message: "has already applied for this pet."
     }
 
+  REVIEWS_PER_PAGE = 5
```
kasugaijin, line 45: See above comment in controller.

test/integration/adoption_application_reviews_test.rb
```diff
@@ -33,7 +33,7 @@ class AdoptionApplicationReviewsTest < ActionDispatch::IntegrationTest
       get staff_adoption_application_reviews_path
 
       assert_response :success
-      AdopterApplication.all.each { |application| verify_application_elements application }
+      AdopterApplication.first(AdopterApplication::REVIEWS_PER_PAGE).each { |application| verify_application_elements application }
```
kasugaijin, line 36: Will need to also update this with a value after removing the constant from the model

db/seeds/01_alta.rb
```diff
@@ -117,41 +117,55 @@
     zipcode: "12345"
   )
 
+  user_profile_params = ->(adopter_foster_account) {
```
kasugaijin, line 120: I actually prefer this approach because it saves us from repeating the profile attributes over and over again. But, it is different from how it's being handled above. So, could you please create an issue to refactor the seed files to use this approach for all cases where we create an adopter foster profile? That would help clean the seeds up nicely!

app/controllers/organizations/staff/adoption_application_reviews_controller.rb
atbalaji: is it okay to set the constants in config/initializers/pagy.rb inside a Pagination module ?  Because I feel that controllers should not be the place to have these constants due to SOC as like models. Would love to hear your thoughts on this

kasugaijin: I don’t think there’s a need for a constant. Just pass the integer in the pagu method in the controller. That’s how I have seen it done in other apps

kasugaijin (approved):
LGTM! Thank you
Please make the issue to update the seed files to use a lambda for the adopter foster profile in both seed files.

Issue #615, opened 2024-04-01
As an admin user, I want to be able to change the roles of other staff and admin users


Ndshah82:
Admin users should be able to change the role of other staff between `staff` and `admin` for their organization.

Acceptance Criteria
**Backend**
- Add a new controller `UserRolesController` (scoped under Organization) with separate actions (and routes) for changing roles on another User e.g., if I am changing a role from `staff` to `admin` we have an action that will delete the `staff` role on the user and create an `admin` role on the user, and another action for the opposite. Each request should be authorized.
- Add a new policy class `UserRolePolicy` that has the general pre-checks (org and active staff) and checks the User making the request is an admin (i.e. has Admin role, and the Admin permissions include a new permission `change_user_role`.  The policy should also prevent a user from making the request on themself (see `StaffAccountPolicy` for reference).

**UI**
- If I am an Admin I should be able to see the staff index table, and in the Role column, see a select that displays the current role for that staff, and if I click the select, I can select the other role. This saves using Turbo upon selection.
- The role select on my own row in the table is disabled.
- User submitting the action is notified of change, or failure of change, via flash

![image](https://github.com/rubyforgood/pet-rescue/assets/95949082/803324c7-020d-434d-9082-b41fffd2ecc4)

Ndshah82:
@kasugaijin Can admin bump other admin down to staff?

kasugaijin:
Yes we only have the two roles at the moment. So an admin can jump another down. We toyed with the idea of super admin but don’t see it as necessary for MVP. Generally an org owner is only going to give admin status to someone they trust to do that role properly so shouldn’t be an issue.

mononoken:
@kasugaijin I think we can close this issue or need to change it up.

In regards to the above, I don't think admin are actually currently able to change roles. At least, I don't recall myself implementing any role interactions for users. They are able to deactivate each other, which is maybe what you meant. Do we want to implement the ability for admins to change roles directly? Currently, the only way to change roles is via the console.

kasugaijin:
@mononoken i think ideally we do have the ability to change role. The thinking behind it was to allow others to take on admin roles in the future if current admins are leaving the org.

mononoken:
[...]

PR #711, opened 2024-05-14
615 admin change roles of staff

jmilljr24:
# 🔗 Issue

#615 

# ✍️ Description

# 📷 Screenshots/Demos

[Screencast from 05-14-2024 07:13:36 PM.webm](https://github.com/rubyforgood/pet-rescue/assets/16829344/937b354c-bf50-4573-858e-03098c8bb71e)


app/controllers/organizations/staff/user_roles_controller.rb
```diff
@@ -0,0 +1,41 @@
+class Organizations::Staff::UserRolesController < Organizations::BaseController
+  include RoleChangeable
```
kasugaijin, line 2: I was thinking we'd have this as a model concern that is included on the user, so we can call user.change_role. Thoughts? I think this would better sit closer to the user model than in the controller as it operates on the user.

jmilljr24: Yes that makes more sense.

kasugaijin (approved):
Looks great. Thanks @jmilljr24

PR #724, opened 2024-05-17
Fix Staff deactivate toggle

### LINKED ISSUE #723: Fix: Staff deactivate toggle turbo-stream
On the Staff page there are two layouts, cards and a table. When toggling deactivate on the card, the table is not always updated.

When toggling deactivate on a card, both card and table should be updated via turbo-stream. Toggling via the table should do the same.

[Screencast from 05-17-2024 06:49:07 AM.webm](https://github.com/rubyforgood/pet-rescue/assets/16829344/9d5fc993-4794-4825-80d9-3b3cd76ecd17)

jmilljr24:
# 🔗 Issue

Resolves #723 

# ✍️ Description

This fixes the staff deactivate toggle by using replace_all and targeting a class instead of an id.

The page loads both cards and tables at the same time so one staff member would have two toggles with the same id on the page causing issues. By changing the id to a class and targeting all classes matching the target name this resolves the issue.

# 📷 Screenshots/Demos

[Screencast from 05-17-2024 07:07:23 AM.webm](https://github.com/rubyforgood/pet-rescue/assets/16829344/3ccc6588-f76c-446a-a1fa-84b25c076846)

### DIFF
--- a/app/views/organizations/staff/staff/_deactivate_toggle.html.erb
+++ b/app/views/organizations/staff/staff/_deactivate_toggle.html.erb
@@ -1,4 +1,4 @@
-<div id=<%= "staff_deactivate_toggle_#{staff.id}" %>>
+<div class=<%= "staff_deactivate_toggle_#{staff.id}" %>>
   <%= form_with model: staff, url: staff_staff_update_activation_path(staff), method: "post" do |form| %>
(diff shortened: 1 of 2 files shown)


kasugaijin (approved):
Ah good find and fix! @jmilljr24

Issue #701, opened 2024-05-11
Page Text Images: Investigate the best way to maintain aspect ratios on About Us images


kasugaijin:
We have the Page Text resource that allows org staff to upload a hero image and two About Us images that are displayed on the org's root page. Currently, if you upload two About Us images of different aspect ratios, they will not render in the same dimensions on the org root page. For example, below. 
<img width="1387" alt="image" src="https://github.com/rubyforgood/pet-rescue/assets/95949082/feaa7f9d-a3e4-40d8-8873-92021140dd99">

If both images are the same aspect ratio, they both show as the same size: 
<img width="1388" alt="image" src="https://github.com/rubyforgood/pet-rescue/assets/95949082/1ffe47cc-053c-4a42-8683-07b7733a8529">

Investigate an appropriate way to manage the aspect ratios of uploaded images. If we don't have to use a gem, great. But we can use a gem if there are no other options. 

Acceptance criteria:
- staff can upload images without worrying about aspect ratios and both images appear the same size on the org root page. 
- No need for tests

Hassaan68:
@kasugaijin , what are the steps to reproduce this one? In local environment the About page is quite different

kasugaijin:
@Hassaan68 so this is on the Org's home page, there is an about us section. You can ignore `/about_us` page. 

Log in as a staff, go to the dashboard, go to Page Texts, make some changes and upload images, save, then go back to that Org's home page and scroll the About Us section. 

Let me know if you have more questions!

jmilljr24:
I think the simplest solution would be to use bootstrap [ratio](https://getbootstrap.com/docs/5.3/helpers/ratio/#about) and style: "object-fit: cover;" so the image fills the space. 

I pushed a branch up if you want to play around with it. Not sure if this what you had it mind.
https://github.com/rubyforgood/pet-rescue/tree/jmilljr24/investigate-image-aspect-ratios

![Screenshot from 2024-05-17 18-13-25](https://github.com/rubyforgood/pet-rescue/assets/16829344/6c06f8d9-09f7-43fb-a2dc-486c84306e12)

![Screenshot from 2024-05-17 18-13-05](https://github.com/rubyforgood/pet-rescue/assets/16829344/0ff29d47-2bd8-40c1-86de-170d4db4a659)

kasugaijin:
Awesome thanks @jmilljr24 i will test it out. I think CSS should be all we need!

kasugaijin:
[...]

PR #731, opened 2024-05-18
Temp fix for About Us images with varying size/aspect ratios

jmilljr24:
# 🔗 Issue

#701 

# ✍️ Description

This is a temporary fix to ensure a consistent layout on the org home page. Org staff can upload two images without any constraints on their dimensions.

This PR uses a ratio of 1:1. The ratio can be changed via bootstrap if there is a aspect ratio that better fits the layout and would least affect the most common upload dimensions. 

Compromise:
- Landscape photos will be cropped on the left and right
- Portrait photos will be cropped top and bottom

# 📷 Screenshots/Demos

Screenshot from 2024-05-17 18-13-25


kasugaijin (approved):
LGTM!

Issue #645, opened 2024-04-17
Foster index and creation form


mononoken:
This work likely needs design tweaks but the implementation can certainly be built right now.

## Description
We want to have an index page for showing a list of all created fosters for the active organization. This page will be meant for staff use. The page should be able to also create new fosters by inputting into a form. In the future, it should be filterable/sortable.

### Data Vocab (refer if you have any confusion)
- `Foster` - caretaking for a pet for a duration of time (contrasted with adoption where there is no end time)
- `Fosterer` - person who fosters
- `Adopter` - person who adopts a pet (note: in the app, it is possible for a user to be both an `adopter` and a `fosterer`)
- `Match` - adoption or foster

## For Design
Design work completed: 
https://www.figma.com/file/0jVgYASUJy0KiX3BVc3dFM/Tasks?type=design&node-id=645-2&mode=dev

> The Pet Task index page can be used as a reference (Task V2 Completed in Figma):
> <img width="722" alt="image" src="https://github.com/rubyforgood/pet-rescue/assets/81536479/37154f49-0084-4942-bbe4-52460e39117f">
> - We do not need the checkboxes, descriptions, or "Show Completed" toggle
> - We do need: (for each entry)
>   - Fosterer name
>   - Pet name
>   - Start date of foster
>   - End date of foster
>   - Edit button
>   - Delete button
>   - Thought: Maybe the delete button should only appear when activating the Edit button to conserve space for mobile
> - We also need a button at the top, like tasks, to open a form for making new fosters
> - We want the mobile view to work as well.

## Acceptance Criteria
- [x] Show all `Match` records with match_type of `:foster`
  - [x] Matches will be grouped by Pet in the index per design direction
  - [x] `Match` records should be scoped for the organization (this most likely is already handled by ActsAsTenant)
  - [x] Default sort? Created at recency? Maybe touch Pet via match updates and sort by last updated pets.
- [x] Index should include the following info:
  - [x] Fosterer name
  - [x] Pet name
  - [x] Start date
  - [x] End date
  - [x] Link for editing individual fosters (should utilize Turbo)
  - [x] Button for destroying the individual foster
~~- [ ] Utilize Turbo to instantiate new `Foster` form~~ (We decided in meeting that this won't be necessary for now)

kasugaijin:
[...]

PR #732, opened 2024-05-18
645 foster index

mononoken:
# 🔗 Issue
#645 

# ✍️ Description
This PR adds the ability for staff to create fosters and manage existing fosters. Most of the action happens on the fosters index that has been added to the sidebar. There is a decent amount of other things touched by this, such as Pets now being able to have multiple Matches. 

I made comments for spots that deserve some attention from reviewers. Let me know if you have any questions!

One to do related to this:
We need to add new emails for fosters and logic to send the correct email.

I think we should remove the callback from the model and put the respective email sends in the controller actions they belong to. The model callback was fairly annoying while testing on this PR.

# 📷 Screenshots/Demos
https://vimeo.com/949462228?share=copy


Gemfile
```diff
@@ -83,6 +83,8 @@ gem "rails-i18n"
 
 # Provides helper methods to easily add 'active' tag on links
 gem "active_link_to"
+
+# Provides named content areas to Action View through `partial` method
```
mononoken, line 87: I was v confused with the _pages partial component logic, since I did not see ViewComponents in our gem file. I am hoping this comment will help others in a similar situation. I think it may be good enough to help others, but we may consider noting this gem somewhere else too.

app/controllers/organizations/staff/manage_fosters_controller.rb
```diff
@@ -1,8 +1,89 @@
...
-    authorize! User, context: {organization: Current.organization},
-      with: Organizations::ManageFostersPolicy
+    @q = authorized_scope(Match.fosters).joins(:pet).ransack(params[:q])
+    @pagy, paginated_fosters = pagy(
+      @q.result
+        .includes(:pet, :user)
+        .order("pets.updated_at DESC"),
+      items: 10
+    )
+    @foster_pets = paginated_fosters.group_by(&:pet)
```
mononoken, line 36: Using pagy and ransack together was painless but getting the sorting to work as expected with the grouping required some finessing. If anyone wants to update this in the future, be careful and read the docs.

app/models/concerns/authorizable.rb
```diff
@@ -30,7 +30,6 @@
     invite_fosterers
     purge_attachments
     manage_default_pet_tasks
-    manage_fosters
```
mononoken, line 33: I removed the ManageFosterPolicy. Since fosters are actually just Match records and currently managing different types of Matches does not require different levels of permission, I think this is simpler for now.

If we want to differentiate them, I think we could build that into the MatchPolicy instead so that it checks permissions based on the `@match.match_type`.

app/models/match.rb
```diff
@@ -25,16 +25,59 @@
...
+  def self.ransackable_associations(auth_object = nil)
+    %w[pet user]
+  end
+
+  def fosterer_name(format = :default)
+    user.full_name(format)
+  end
+
+  def status
+    return :not_applicable if start_date.nil? || end_date.nil?
```
mononoken, line 68: Anyone have thoughts on this method? Currently, it's only really applicable for fosters but could see a future where it could be useful for adoptions too.

app/models/pet.rb
```diff
@@ -32,7 +32,7 @@ class Pet < ApplicationRecord
 
   has_many :adopter_applications, dependent: :destroy
   has_many :tasks, dependent: :destroy
-  has_one :match, dependent: :destroy
+  has_many :matches, dependent: :destroy
```
mononoken, line 35: ## Key change!!
Updated matches so that pets can have multiple matches, since they can be fostered multiple times as well as be adopted after being fostered.

All the tests pass, but I am worried there may be logic somewhere that is muddled by this. Is there anywhere that I missed that expected pets to only have one Match and would be broken now that they can have multiple?

app/models/user.rb
```diff
@@ -81,4 +89,15 @@ def active_for_authentication?
...
+  def full_name(format = :default)
+    case format
+    when :default
+      "#{first_name} #{last_name}"
+    when :last_first
+      "#{last_name}, #{first_name}"
+    else
+      raise ArgumentError, "Unsupported format: #{format}"
+    end
+  end
```
mononoken, line 102: I was a little torn if this belonged here or in a helper. I think it makes sense in the model.

app/views/organizations/staff/manage_fosters/edit.html.erb
```diff
@@ -1,58 +1,11 @@
...
-    <div class="row">
-      <div class="col-md-6 mx-auto card p-5 ">
-        <div class='d-flex flex-row-reverse'>
-          <%= link_to "#{@application.adopter_foster_account.user.first_name}'s Profile",
-              staff_profile_review_path(@application.adopter_foster_account.adopter_foster_profile.id),
-              class: 'text-decoration-none bigger' %>
-        </div>
+<%= render "components/dashboard/page" do |p| %>
+  <% p.header_title do %>
+    Edit Foster
```
mononoken, line 5: Currently, editing is done with the turbo frames. However, I included this edit page for cases where JavaScript is disabled.

app/views/organizations/staff/manage_fosters/_foster_cards.html.erb
```diff
@@ -0,0 +1,73 @@
+<% foster_pets.each do |pet, fosters| %>
+  <%# Grid cell %>
+  <div class="col-lg-4 col-md-6 p-0">
```
mononoken, line 3: I think we have multiple card partials at this point. It may be a good task for the mobile work for the conference to make a shared card partial/component.

kasugaijin (approved):
This looks great @mononoken 
Just a couple small comments on the pet scopes.
Also, could we add some seeds for foster matches? This can be done in another task...also. This PR is already large!

I also note some new issues we can make:
- Prevent someone editing a pet to be only adoptable if a match with foster type is active
- Staff Pet Show - Add a new fosters tab to show all fosters on that pet
- Staff Pet Show - When a pet has a match, we remove some details from the overview table, these should not be removed if the match types are fosters.

app/models/pet.rb
```diff
@@ -70,9 +70,13 @@ class Pet < ApplicationRecord
     WEIGHT_UNIT_KG
   ]
 
-  scope :adopted, -> { Pet.includes(:match).where.not(match: {id: nil}) }
-  scope :unadopted, -> { Pet.includes(:match).where(match: {id: nil}) }
+  scope :adopted, -> { Pet.includes(:matches).where.not(matches: {id: nil}) }
```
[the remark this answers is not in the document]
mononoken: 28eeeb70a44f34875df6f7379f89f7a880e195a3 fixes

app/models/pet.rb
```diff
@@ -70,9 +70,13 @@ class Pet < ApplicationRecord
...
+  scope :unadopted, -> { Pet.includes(:matches).where(matches: {id: nil}) }
```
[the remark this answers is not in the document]
mononoken: I ended up renaming this to `unmatch`. It's only used in one spot, and I think its intention is for showing pets without any match. Please confirm though. 

1bf3051dfafc82ad4376607a51e01055edcaea9d

mononoken:
> This looks great @mononoken Just a couple small comments on the pet scopes. Also, could we add some seeds for foster matches? This can be done in another task...also. This PR is already large!
> 
> I also note some new issues we can make:
> 
> * Prevent someone editing a pet to be only adoptable if a match with foster type is active
> * Staff Pet Show - Add a new fosters tab to show all fosters on that pet
> * Staff Pet Show - When a pet has a match, we remove some details from the overview table, these should not be removed if the match types are fosters.

Thank you for the review, @kasugaijin ! I made the fix from your comments and have added three fosters, one for each status, for each seed. I will get this merged.

mononoken:
@all-contributors

please add @kiryth for design.

Issue #718, opened 2024-05-15
PageTexts: Enable staff to customize the text shown to the right of an adoptable dog


kasugaijin:
On the adoptable dog show page we have this box of info to the right, this is important details that an org wants to get across to someone considering adopting. We should enable staff to customize this. 

![image](https://github.com/rubyforgood/pet-rescue/assets/95949082/f4930a74-e477-4148-9fbd-f3a287396d83)

We want to add a new column on the page texts table called `adoptable_pet_info`, `text`, and add this as a text area input to the page texts form under a new H3 heading Adoptable Pet Information.

In the UI for the adoptable pet show page, we want to keep the sentence referring to the FAQ at all times. If there is a `pagetext.adoptable_pet_info` then we want to show a heading called Important Information, and render the text provided. We can delete the rest of the current text in the view for this page.

MooseCowBear:
I'd like to work on this

PR #733, opened 2024-05-18
718 allow staff customize text adoptable pet show

MooseCowBear:
# 🔗 Issue

#718 

# ✍️ Description

Adds the new column for adoptable pet info to page texts table and updates the relevant form and the adoptable pets show view to display info if present. 

The form looked a little weird with the label and header being the same text so I tried to come up with something that would convey the same meaning but not be identical. Happy to change it if you want. 

# 📷 Screenshots/Demos

The form with a placeholder: 

Screenshot 2024-05-18 at 5 02 42 PM

Adoptable pets view with information: 

Screenshot 2024-05-18 at 4 59 45 PM

### DIFF
--- a/app/views/organizations/adoptable_pets/show.html.erb
+++ b/app/views/organizations/adoptable_pets/show.html.erb
@@ -75,7 +75,7 @@
               <p class="fs-3  mb-5"><%= t('.please_read_faq_html', faq_link: link_to('FAQ', faq_index_path)) %></p>
             </div>
 
-            <% if @adoptable_pet_info %>
+            <% if @adoptable_pet_info.present? %>
               <div class="row align-items-center">
                 <div class="mb-6">
                   <h3 class="mb-3 fw-bold"><%= t('.important_info') %></h3>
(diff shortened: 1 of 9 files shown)


app/views/organizations/adoptable_pets/show.html.erb
```diff
@@ -73,24 +73,18 @@
...
               <!-- para -->
               <p class="fs-3  mb-5"><%= t('.please_read_faq_html', faq_link: link_to('FAQ', faq_index_path)) %></p>
-
             </div>
 
-            <div class=" row align-items-center ">
-              <div class="mb-6">
-                <h3 class="mb-3 fw-bold"><%= t('.please_know_that') %></h3>
-                <p class="mb-0 fs-4 ">
-                <ul><%= t('.please_know_that_items_html') %></ul>
```
jmilljr24, line 83: I don't see this  i18n used anywhere else after this change. Think we should remove it from en.yml?

jmilljr24 (commented):
I took a quick glance and this looks really good. I'll do a full review at a later time unless someone can do it sooner.

The only minor change would be to remove some i18n lines that wouldn't be used after this pr(unless I'm mistaken).

Thanks @MooseCowBear!!!

jmilljr24 (approved):
lgtm! Thanks @MooseCowBear

PR #734, opened 2024-05-19
Removes `/about` and `/partners` from tenanted organizations 

### LINKED ISSUE #726: Static pages: remove the partners and about us pages
Delete the routes, controller actions and views (and any applicable tests) for About Us and Partners pages as they are not required.

LindseyB:
# 🔗 Issue
Closes #726 

* #726 

# ✍️ Description

This removes the static pages for about and partners from tenanted orgs. 

eg: 

✅  `/about` works
✅  `/partners` works
❌  `/alta/about` 404s
❌  `/alta/partners` 404s

Also it looks like my editor decided to remove some trailing whitespace from a file I was poking in while I was checking if the view rendered for the tenanted org was the same as the root one. Harmless so I kept it in. 

Mostly this adds a new `before_action` to the `StaticPagesController` it could be moved up into the `ApplicationController` if we need this sort of behavior elsewhere, but for now I kept it there. 

```ruby
def require_no_tenant
  unless Current.organization.blank?
    render "errors/not_found", status: :not_found and return
  end
end
```

note that in this file earlier I see a `if !current_tenant` but in my testing `current_tenant` was always nil, so I decided to use `Current.organization` instead which appears to be set by the middleware and appears to have the right data I need to check if I'm in a tenanted org or not. 

# 📷 Screenshots/Demos

Not much of a demo to share here, but the tests should give a fair demonstration of the behavior.


kasugaijin (approved):
Looks good! Thank you @LindseyB

PR #738, opened 2024-05-23
728 Adopter dashboard adoption applications

### LINKED ISSUE #728: Adopter Applications: As an adopter I can see my list of adoption applications in my dashboard
Log in as an adopter and go to your dashboard. When you click My Applications you will just see the dashboard index redisplayed. Change this link text to be 'Adoption Applications' and when you click it, render the index/list of the user's adoption applications. See the AdopterApplicationController index action for context. 

The applications should be displayed in a table - create a new partial for this. It should display the pet image, pet name, created (show created_at date), status, and a red button to withdraw the application (hits the update action on the controller to set the status to withdrawn).

image

jmilljr24:
# 🔗 Issue

Resolves #728 

# ✍️ Description

This adds an Adoption Application list to the Adopters dashboard. The applications are listed in table for larger screens and cards for smaller screens. Application statuses are updated via turbo.

I also added a method to allow for enum's to be translated.
```
#example of AdopterApplication enum of :status

app = AdopterApplication.first
app.human_enum_name(:status)
=> "Under Review"
```

# 📷 Screenshots/Demos

[Screencast from 05-23-2024 10:20:06 AM.webm](https://github.com/rubyforgood/pet-rescue/assets/16829344/fb5bd670-0c43-4913-bef2-efa5e1829e40)


app/views/layouts/adopter_foster_dashboard.html.erb
```diff
@@ -1,118 +1,117 @@
...
+                      <!-- Nav item -->
+                      <li class="nav-item">
+                        <%= active_link_to adopter_fosterer_dashboard_index_path, class: "nav-link" do %>
+                          <i class="fe fe-home nav-icon"></i> My Dashboard
+                        <% end %>
+                      </li>
+                      <!-- Nav item -->
+                      <li class="nav-item">
+                        <%= active_link_to adopter_fosterer_adopter_applications_path, class: "nav-link" do %>
+                          <i class="fe fe-book nav-icon"></i> <%= t("dashboard.applications.header_title")%>
```
jmilljr24, line 61: Sorry the diff is showing everything. This is the only real change for this file.

kasugaijin (approved):
Looks good!

I note we have a controller test for adopter applications. Do we currently need the like integration test as well? It's not really doing much.

We have a `profile_show` attribute on AdopterApplication. I used this in the old app to allow adopters to delete an application (only if the pet was adopted to someone else, or the adopter withdrew) so it no longer showed in their dashboard...but the record still exists and available on the staff end. If this makes sense, we could implement something similar here. Otherwise someone could end up with a load of withdrawn apps in their view. This could be in a new issue.

jmilljr24:
> Looks good!
> 
> I note we have a controller test for adopter applications. Do we currently need the like integration test as well? It's not really doing much.
> 
> We have a `profile_show` attribute on AdopterApplication. I used this in the old app to allow adopters to delete an application (only if the pet was adopted to someone else, or the adopter withdrew) so it no longer showed in their dashboard...but the record still exists and available on the staff end. If this makes sense, we could implement something similar here. Otherwise someone could end up with a load of withdrawn apps in their view. This could be in a new issue.

I'll take a glance and get back to you. I honestly only briefly looked at the tests since they were already implemented.

jmilljr24:
Can you give more detail on this? I'm not sure what test/file you are referring to.
> I note we have a controller test for adopter applications. Do we currently need the like integration test as well? It's not really doing much.

<br>
<br>

Yes I think that is a good idea based on the conditions mentioned. Sounds liked a good new issue. I will not have time before RFG.
> We have a `profile_show` attribute on AdopterApplication. I used this in the old app to allow adopters to delete an application (only if the pet was adopted to someone else, or the adopter withdrew) so it no longer showed in their dashboard...but the record still exists and available on the staff end. If this makes sense, we could implement something similar here. Otherwise someone could end up with a load of withdrawn apps in their view. This could be in a new issue.

kasugaijin:
@jmilljr24 this test file `test/integration/adopter_application_test.rb` I think we can delete it. It's just dead weight at the moment, and seems fitting on the PR?

kasugaijin:
Thanks. Ready for merge I say.

kasugaijin (approved):
LGTM!

PR #742, opened 2024-05-23
Liked pets feature

kasugaijin:
# 🔗 Issue
NA

# ✍️ Description
Adds a liked pets feature so adopter users can like a pet and see them from their dashboard later. See comments below for more discussion.

https://www.loom.com/share/1d6210a54ebd40bd83c55cac955fb42a?sid=6e5bb0c8-f7e9-447c-8f47-e7bb15796a76

# 📷 Screenshots/Demos


app/controllers/organizations/adoptable_pets_controller.rb
```diff
@@ -3,7 +3,7 @@ class Organizations::AdoptablePetsController < Organizations::BaseController
 
   def index
     @pets = authorized_scope(
-      Pet.includes(:adopter_applications, images_attachments: :blob),
+      Pet.includes(:adopter_applications, :liked_pets, images_attachments: :blob),
```
kasugaijin, line 6: eager load a pet's liked pets for the index view. We need to tell a user if they have liked a pet or not, so for each pet card we will query this data to find out if their adopter_foster_account id is in the collection of the pet's liked_pets

app/models/concerns/authorizable.rb
```diff
@@ -22,6 +23,7 @@
       create_adopter_applications
       create_adopter_foster_profiles
       manage_adopter_foster_profiles
+      manage_liked_pets
```
kasugaijin, line 26: assuming there is no need for staff to like pets.

app/policies/organizations/liked_pet_policy.rb
```diff
@@ -0,0 +1,27 @@
...
+  private
+
+  def verify_pet_likable!
+    deny! if pet.organization_id != user.organization_id
+  end
+
+  def verify_adopter_foster_account!
+    deny! unless user.adopter_foster_account.present?
+  end
+end
```
kasugaijin, line 27: I have left this pretty open ended such that an adopter or fosterer, even without a profile, can like and unlike any pet. I say we take this approach to start and let feedback inform us if we need to make changes. My assumption here is that people would like to be able to track liked pets before investing the time to fill out a profile. I see it as a way for users to shortlist pets before making the investment.

app/views/partials/_like_button.html.erb
```diff
@@ -0,0 +1,17 @@
+<% if liked %>
+  <%= button_to adopter_fosterer_liked_pet_path(id: LikedPet.find_by(adopter_foster_account_id: current_user.adopter_foster_account.id, pet_id: pet.id)&.id),
```
kasugaijin, line 2: Not usually a fan of doing queries in views. Not sure where this would best live, otherwise, either. If we move this method somewhere it's just going to be more code to achieve the same thing as we'd still need to pass the args to the method that then does the query...

app/views/partials/_pet_card.html.erb
```diff
@@ -1,26 +1,34 @@
 <div class="col-lg-4 col-md-6 p-0">
   <%= link_to adoptable_pet_path(pet), class: 'text-decoration-none' do %>
     <div class="card card-hover m-4" style='overflow: hidden'>
```
kasugaijin, line 3: I moved the 'Adoption Pending' badge to the top right of the image so there was more room for the like functionality. I also prefer this placement of this badge to before. Thoughts?

I also make the image and the pet name clickable to go to the pet show page, instead of the whole card. This allows the user to click the like/unlike independently in the pets index view, and is consistent with how other websites link their cards. Usually details are not clickable, only names and images.

app/views/partials/_pet_card.html.erb
```diff
@@ -1,26 +1,34 @@
...
-          <div id="adoption-status">
-            <%= tag.span("Adoption Pending", class: "badge bg-info") if pet.has_adoption_pending? %>
-          </div>
+        <li class="list-group-item d-flex justify-content-between align-items-center flex-wrap">
+          <%= link_to adoptable_pet_path(pet), class: 'text-decoration-none' do %>
+            <h3 class="card-title text-secondary mb-0">
+              <%= pet.name %>
+            </h3>
+          <% end %>
+          <% if current_user&.adopter_foster_account %>
```
kasugaijin, line 21: I am still using this because I find it easier to read...thoughts?

test/policies/adopter_application_policy_test.rb
```diff
@@ -102,8 +102,8 @@ class AdopterApplicationPolicyTest < ActiveSupport::TestCase
         @user = create(:fosterer, :with_profile)
       end
```
kasugaijin, line 103: I added the adopter role to the fosterer factory seeing as we do this now.

test/integration/adoptable_pets_index_test.rb
```diff
@@ -17,9 +17,9 @@ class AdoptablePetsIndexTest < ActionDispatch::IntegrationTest
 
     assert_select "h1", "Up for adoption"
     assert_response :success
-    assert_select "a[href$='adoptable_pets/#{@available_pet.id}']", 1
+    assert_select "a[href$='adoptable_pets/#{@available_pet.id}']", 2
```
kasugaijin, line 20: There are now two a tags because I have made the image and pet name clickable.

app/controllers/organizations/adoptable_pets_controller.rb
jmilljr24: `    @likes = LikedPet.where(adopter_foster_account: current_user&.adopter_foster_account)
`

app/views/partials/_pet_card.html.erb
```diff
@@ -1,26 +1,34 @@
...
+            <div class='text-end' id="like_button_pet_<%= pet.id %>">
+              <%= render partial: 'partials/like_button', locals: { pet: pet, liked: pet.liked_pets.pluck(:adopter_foster_account_id).include?(current_user.adopter_foster_account.id) } %>
```
jmilljr24, line 23:
```suggestion
              <%= render partial: 'partials/like_button', locals: { pet: pet, liked: @likes&.find { |like| like.pet_id == pet.id}  } %>
```

app/views/partials/_like_button.html.erb
jmilljr24:
```suggestion
  <%= button_to adopter_fosterer_liked_pet_path(id: liked&.id),
```

app/views/organizations/adopter_fosterer/liked_pets/create.turbo_stream.erb
```diff
@@ -0,0 +1,7 @@
+<%= turbo_stream.replace "like_button_pet_#{@pet.id}" do %>
+  <div id="like_button_pet_<%= @pet.id %>" class="d-flex flex-column align-items-center">
+    <%= render partial: 'partials/like_button', locals: { pet: @pet, liked: true } %>
```
jmilljr24, line 3:
```suggestion
    <%= render partial: 'partials/like_button', locals: { pet: @pet, liked:  @liked_pet } %>
```

jmilljr24 (commented):
With regards to the like_button query, I think I have an alternate solution but if I'm missing the mark feel free to disregard. For clarity I'll be refer to the liked pet object from the LikedPet model not a Pet that has been 'liked' 

Starting from the adoptable_pets_controller, we could create an instance variable of all the LikedPet objects associated with the adopter.

Instead of true/false being passed to the like button, the pet_card would pass in the LikedPet object if the instances variable contains it. 

The like button would simply use the LikedPet object id for the button path instead of a query.

app/views/organizations/adopter_fosterer/liked_pets/create.turbo_stream.erb
kasugaijin: Yes this will be better @jmilljr24, I will make changes.

kasugaijin:
One thing I will also do (which I don't think is a blocker for review and approval?) is move text from views and controllers to the translations file before merging.

app/controllers/organizations/adopter_fosterer/liked_pets_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+        format.html { redirect_to request.referrer, alert: "Error, pet was not removed from your liked pets." }
+        format.turbo_stream { flash.now[:alert] = "Error, pet was not removed from your liked pets." }
+      end
+    end
+  end
+
+  private
+
+  def liked_pet_params
+    params.permit(:pet_id, :id, :_method, :authenticity_token)
```
kasugaijin, line 49: I note that if I do not have these in the permitted params I get this error
```
Unpermitted parameters: :_method, :authenticity_token. Context: { controller: Organizations::AdopterFosterer::LikedPetsController, action: destroy, request: #<ActionDispatch::Request:0x000000011a825990>, params: {"_method"=>"delete", "authenticity_token"=>"[FILTERED]", "controller"=>"organizations/adopter_fosterer/liked_pets", "action"=>"destroy", "id"=>"32"} }
```
The destroy or create still works, but I added them to the permitted params list to remove the erorr. Any thoughts?

kasugaijin: Yes this is the more rails way for sure.

package.json
mononoken, line 1: I think this was unintentionally generated.

test/factories/users.rb
```diff
@@ -46,6 +46,7 @@
 
       after(:build) do |user, _context|
         user.add_role(:fosterer, user.organization)
+        user.add_role(:adopter, user.organization)
```
mononoken, line 49: Good catch 👍

app/views/partials/_pet_card.html.erb
mononoken: Looks and sounds good to me! I also made the fosters cards with only the image clickable so this is consistent there.

mononoken (commented):
I have not looked at the tests yet, but here is my initial review. My main concern is in the comments about the naming of `LikedPet`. Would be curious if others agree.

app/controllers/organizations/adopter_fosterer/liked_pets_controller.rb
```diff
@@ -0,0 +1,59 @@
+class Organizations::AdopterFosterer::LikedPetsController < Organizations::BaseController
+  before_action :authenticate_user!
```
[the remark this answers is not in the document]
kasugaijin: Yes I agree the skip makes more sense as it occurs less often.

app/models/liked_pet.rb
```diff
@@ -0,0 +1,31 @@
...
+#  index_liked_pets_on_organization_id                       (organization_id)
+#  index_liked_pets_on_pet_id                                (pet_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (adopter_foster_account_id => adopter_foster_accounts.id)
+#  fk_rails_...  (organization_id => organizations.id)
+#  fk_rails_...  (pet_id => pets.id)
+#
+class LikedPet < ApplicationRecord
```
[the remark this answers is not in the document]
kasugaijin: 100% I agree. LikedPets is a bad name for it. Likes makes more sense.

app/models/adopter_foster_account.rb
```diff
@@ -23,6 +23,8 @@ class AdopterFosterAccount < ApplicationRecord
   has_one :adopter_foster_profile, dependent: :destroy
   has_many :adopter_applications, dependent: :destroy
   has_many :matches, dependent: :destroy
+  has_many :liked_pet_associations, class_name: "LikedPet", dependent: :destroy
+  has_many :liked_pets, through: :liked_pet_associations, source: :pet
```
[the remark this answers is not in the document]
kasugaijin: Exactly! Likes makes more sense. It even tripped me up once returning pets when I expected likes.

package.json
kasugaijin: Yes I tested out using awesome fonts. This is a remnant I can delete.

app/controllers/organizations/adopter_fosterer/liked_pets_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+  before_action :set_pet, only: [:create]
+  before_action :set_liked_pet, only: [:destroy]
+  layout "adopter_foster_dashboard", only: :index
+
+  def index
```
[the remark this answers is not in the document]
kasugaijin: My intention was to get this up and running then we can decide what to do here and make a future issue for it.

app/views/partials/_pet_card.html.erb
kasugaijin: Yes, I agree there is pros and cons to both approaches. Perhaps something to discuss broader so we can set a consistent approach through the app.

app/controllers/organizations/adopter_fosterer/liked_pets_controller.rb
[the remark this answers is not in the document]
kasugaijin: I think it's nicer for the user to see a pet they have liked has been adopted, rather than just have it disappear from their liked pets, which could be a bit confusing. Therefore, I propose we add the badge if a pet has been adopted. To that end, let's wait until your PR for fosters has been merged, because in order to show the badge we need to determine if the pet has been adopted, and that required the changes to the match, i.e., from singular to plural, on the Pet model.

app/controllers/organizations/adopter_fosterer/liked_pets_controller.rb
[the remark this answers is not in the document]
kasugaijin: https://github.com/rubyforgood/pet-rescue/issues/745

app/models/liked_pet.rb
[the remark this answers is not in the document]
kasugaijin: Done.

kasugaijin:
> I have not looked at the tests yet, but here is my initial review. My main concern is in the comments about the naming of `LikedPet`. Would be curious if others agree.

Yes, I totally agree, and have made the changes to rename as Like instead of LikedPet.

app/controllers/organizations/adopter_fosterer/likes_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+      else
+        format.html { redirect_to request.referrer, alert: t("organizations.adopter_fosterer.likes.create.failed", name: @pet.name) }
+        format.turbo_stream { flash.now[:alert] = t("organizations.adopter_fosterer.likes.create.failed", name: @pet.name) }
+      end
+    end
+  end
+
+  def destroy
+    @pet = @like.pet
+    authorize! context: {pet: @pet}, with: Organizations::LikePolicy
```
mononoken, line 33: I have some suggested refactoring here for the policy.

We can remove this `authorize!` line from `#destroy` and move it to `#set_like` but reword it as such:
```ruby
  def set_like
    @like = Like.find(params[:id])
    authorize! @like
  end
```

For actions that affect existing records (such as destroy, update, show), this is the pattern that should usually be used. We fetch the existing record, and then we authorize via that record. Action Policy can infer the correct policy based on the record's class. We don't need the pet context in this case. We should be performing checks directly on the record itself.

We can add this private method to the `LikePolicy`.
```ruby
  def pet
    @pet || record.pet
  end
```
This method will fetch the passed pet context if it exists (so for actions that do not have existing records). However, if the context is not given, it will then check if the passed record responds to `#pet`.

test/policies/organizations/like_policy_test.rb
```diff
@@ -0,0 +1,215 @@
...
+        @user = create(:staff)
+      end
+
+      should "return false" do
+        assert_equal @action.call, false
+      end
+    end
+  end
+
+  context "#create?" do
```
mononoken, line 98: This is a duplicate name of the above context. Should the tests in this block be moved into the one above?

test/policies/organizations/like_policy_test.rb
```diff
@@ -0,0 +1,215 @@
...
+        @user = create(:staff)
+      end
+
+      should "return false" do
+        assert_equal @action.call, false
+      end
+    end
+  end
+
+  context "#destroy?" do
```
mononoken, line 179: This is a duplicate name of the above context. Should the tests in this block be moved into the one above?

mononoken (approved):
Added a few more comments on the policies and tests. 

Looks good! I think this will be a nice feature for the adopters.

app/controllers/organizations/adopter_fosterer/liked_pets_controller.rb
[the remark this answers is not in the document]
mononoken: I agree that it would probably be confusing for adopters if liked pets just disappeared. Sounds good!

Issue #746, opened 2024-05-26
Refactor user full names by string concat to utilize new method instead


mononoken:
## Description
There are various places where we display a user's full name by calling something like:
```ruby
"#{user.first_name} #{user.last_name}"
```
We have introduced a new method, `User#full_name` so that we can refactor the above to:
```ruby
user.full_name
```

The method also supports a `:last_first` option if you find instances of that as well.

## Acceptance criteria
- [ ] All locations that concat `user.first_name` and `user.last_name` are replaced with `user.full_name`

mononoken:
Go for it @sparshalc ! You are assigned now.

PR #761, opened 2024-05-29
Refactor: Use 'full_name' attribute for user full names instead of st…

sparshalc:
🔗 Issue:  #746

### DIFF
--- a/app/helpers/avatar_helper.rb
+++ b/app/helpers/avatar_helper.rb
@@ -4,7 +4,7 @@ module AvatarHelper
 
     render partial_name do |partial|
       partial.alt "#{current_user.first_name.first.upcase} Avatar"
-      partial.initials "#{current_user.first_name.first}#{current_user.last_name.first}".upcase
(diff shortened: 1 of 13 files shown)


app/helpers/avatar_helper.rb
```diff
@@ -4,7 +4,7 @@ def render_avatar_partial(xl = false)
 
     render partial_name do |partial|
       partial.alt "#{current_user.first_name.first.upcase} Avatar"
-      partial.initials "#{current_user.first_name.first}#{current_user.last_name.first}".upcase
+      partial.initials "#{current_user.full_name.split.map { |part| part[0] }.join}".upcase
```
kasugaijin, line 7: Is it worth adding an initials method to the User model instead?

kasugaijin:
Looks good so far @sparshalc. See my comment above - and note a couple tests are broken and likely need updating. Run `rails standard:fix` to run the linter and pass that check.

kasugaijin (approved):
Great! Thank you.

Issue #741, opened 2024-05-23
Tests: Add adopter foster account model tests


kasugaijin:
We are currently missing the test file for this resource in test/models. Please add the file with tests for associations.

MooseCowBear:
I could do this. Do we also want tests for the scopes?

kasugaijin:
I think we could add for associations, validations, and the setter instance methods.

PR #762, opened 2024-05-29
add tests for AdopterFosterAccount model

MooseCowBear:
# 🔗 Issue

#741 

# ✍️ Description

Adds tests for the AdopterFosterAccount model

# 📷 Screenshots/Demos


test/models/adopter_foster_account_test.rb
```diff
@@ -0,0 +1,33 @@
+require "test_helper"
+
+class AdopterFosterAccountTest < ActiveSupport::TestCase
+  context "associations" do
```
kasugaijin, line 4: Sorry, I just merged a PR that adds a couple more associations on this model. 
```
has_many :likes, dependent: :destroy
  has_many :liked_pets, through: :likes, source: :pet
```

Can you also add these, please?

MooseCowBear:
@kasugaijin I added a factory for a fosterer that is not also an adopter so the `adopters` scope test would still test that is excludes the right things since the fosterer factory now includes the adopter role.

kasugaijin:
> @kasugaijin I added a factory for a fosterer that is not also an adopter so the `adopters` scope test would still test that is excludes the right things since the fosterer factory now includes the adopter role.

Hmm so we by default give fosters the adopter role also so I am not sure if this is necessary. It’s expected that fosterers have both roles and adopters only have the adopter role. So, I am thinking we don’t need that factory and the assertions should reflect these expectations. Does that make sense?

MooseCowBear:
@kasugaijin Yep, no problem. I wasn't sure if fosters would also always be adopters.

kasugaijin (approved):
LGTM! Thank you.

Issue #737, opened 2024-05-22
I18N - find places where we have not used I18N and add copy to en.yml


kasugaijin:
We started off well adding text displayed in the UI (view copy, alert messages, and active record error messages) in the I18n en.yml file, but then we got a bit forgetful. 

This task is to look through the application and find cases where we should move text from a view to en.yml, a controller (e.g., alert messages), and also add copy in en.yml for active record validation error messages. 

I am not specifying a set amount of work for this task as I don't know how many places this occurs. If you prefer, you can scope the work and report back, or just start making the changes and make a PR resource by resource.

wandergithub:
Can I work on this issue?

jmilljr24:
@wandergithub just an fyi, in #738 which will be merged soon I added a method which you can use for translating enum's if you get to those in this issue. Let me know if you have any questions about it.

wandergithub:
@kasugaijin Hi, after working and investigating this issue I've noticed that updating translations in big chunks has led to some translations being missing due to managing too much of them.

I don't know what you guys would like me to do with #791 or if you want me to continue with the views to manage to close this issue. But I've been thinking it would be better if the translations are added while working with code and changes related to them in isolated ways. Like a gradual work/progress on upcoming PRs. Now it should be easier with the rules while we reinforce the use of I18n in the reviews. I can help with that too.

kasugaijin:
@wandergithub I'd say do as much as you are comfortable doing in the PR, then we can try address the rest in related work. The only issue with leaving it to other work is that it will probably be forgotten about, and there may be areas that do not get any focus for a while, if at all. But, that said, I don't want this to become uncomfortably large for you either.

wandergithub:
I will wrap up my contributions to the translations with this last PR. Moving on to addressing other topics.

Meri-MG:
Hello, @kasugaijin, can I start working on this issue?

kasugaijin:
@Meri-MG hey this issue got split into two other issues that were picked up. I should close this issue. Sorry about that. Is there another ‘ready’ issue you are interested in?

PR #763, opened 2024-05-29
737 i18n update all missing translations initial PR

wandergithub:
# 🔗 Issue

https://github.com/rubyforgood/pet-rescue/issues/737
# ✍️ Description

These are general changes, now I'm planning to address each resource (Controller-Model-View Packs)  individually on separate PRs. (what do you think about this?)
- adds error translation to en.yml
- adds the error translation where it applies
- adds contactsController translation and updates it on test(This code is not currently being used on route)
- Important: I think we should Set config of i18n to raise errors while testing when missing a translation after I address all errors it raises. (what do you think about this?)
At:`config/enviroments/test.rb`
`  # Raises error for missing translations.`

 `config.i18n.raise_on_missing_translations = true`
# 📷 Screenshots/Demos


kasugaijin:
@wandergithub I agree with the approach to break it up like that. I also think it would be useful to add the missing translation exception in test and dev environments so we get quick feedback. We'll want to handle it more gracefully in production.
FYI two fixes required - test is broken, and run `rails standard:fix` to lint.

kasugaijin:
@wandergithub is it worth adding that configuration in this 'general' PR?

kasugaijin:
I see this test is failing in CI
```
Error:
  UsersTest#test_user_can_log_out:
  Ferrum::ProcessTimeoutError: Browser did not produce websocket url within 20 seconds, try to increase `:process_timeout`. See https://github.com/rubycdp/ferrum#customization
      test/system/users_test.rb:12:in `block in <class:UsersTest>'
  
  bin/rails test test/system/users_test.rb:11
```

Does it pass or fail locally for you? Run `rails test:system`

wandergithub:
> @wandergithub is it worth adding that configuration in this 'general' PR?

You are referring to I18n (if not clarify please).
`config.i18n.raise_on_missing_translations = true`
I would add it once I solve a few missing translations. If not then it's going to raise errors.

wandergithub:
> I see this test is failing in CI
> 
> ```
> Error:
>   UsersTest#test_user_can_log_out:
>   Ferrum::ProcessTimeoutError: Browser did not produce websocket url within 20 seconds, try to increase `:process_timeout`. See https://github.com/rubycdp/ferrum#customization
>       test/system/users_test.rb:12:in `block in <class:UsersTest>'
>   
>   bin/rails test test/system/users_test.rb:11
> ```
> 
> Does it pass or fail locally for you? Run `rails test:system`

Systems tests are falling for me locally I don't know why. But I though It wouldn't be a problem since I didn't change anything related. Seems I was wrong. Any Idea of what's happening here:
image

kasugaijin:
A brief look and it appears it could be a WSL gremlin. Try https://github.com/orgs/community/discussions/52722

wandergithub:
I just confirmed all test are passing locally. whats next step now? @kasugaijin 
image

wandergithub:
> > I see this test is failing in CI
> > ```
> > Error:
> >   UsersTest#test_user_can_log_out:
> >   Ferrum::ProcessTimeoutError: Browser did not produce websocket url within 20 seconds, try to increase `:process_timeout`. See https://github.com/rubycdp/ferrum#customization
> >       test/system/users_test.rb:12:in `block in <class:UsersTest>'
> >   
> >   bin/rails test test/system/users_test.rb:11
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Does it pass or fail locally for you? Run `rails test:system`
> 
> Systems tests are falling for me locally I don't know why. But I though It wouldn't be a problem since I didn't change anything related. Seems I was wrong. Any Idea of what's happening here: image

regarding this. The link you provided is a different error. I didn't have chromium. Installing it solves the issue. Can we do something like putting it as a requirement on the readme so new devs like me don't experience what I did? Give me the green light and I'll send the PR.

kasugaijin:
Ah ok good job. Yes you can make a PR to add to the readme to install chromium if that error is encountered.

for this PR, you could also resolve the existing errors then add the config to raise an error for missing translations. Or if you want to do that separately we can merge this one. Your call. Then continue on with the other PRs as you outlined above.

kasugaijin (approved):
LGTM!

PR #766, opened 2024-05-31
Add a docker-compose file to launch local service dependencies

cflipse:
This provides an _alternative_ to having to install a local postgres
service, and presents a mechanism for running ports isolated from any other
docker service clusters that may be running on the same machine.

# 🔗 Issue

# ✍️ Description

# 📷 Screenshots/Demos


maebeale (approved):
🙌 thanks, @cflipse !

Issue #752, opened 2024-05-28
Add validation for changing pet's placement type


mononoken:
## Description
Pets can have their attributes edited at `/staff/pets/:id/edit`

We want to add a validation to prevent the pet's `placement_type` from being changed to one that conflicts with their current foster `Match`. If a pet is currently being fostered (which means they have an active or upcoming `Match` with `match_type: "foster"`), we want a validation to prevent the `placement_type` from being changed to only `"Adoptable"`.

Note, the validation should not prevent a `Pet` that has a completed foster from having its `placement_type` changed. We only care about fosters that have `end_dates` after the current time. 

## Acceptance criteria
- [ ] Add described validation to `Pet` model
- [ ] Add unit test for validation

MooseCowBear:
I can work on this

jmilljr24:
@MooseCowBear Thank you!!

PR #783, opened 2024-05-31
752 validation pet placement type

MooseCowBear:
# 🔗 Issue

#752 
# ✍️ Description

Adds validation to make sure pets with unfinished foster matches cannot have placement type changed to adoptable + tests. 

Error message below. Should it be more explicit? Like, "must include Fosterable while pet is being fostered" or something like that?

# 📷 Screenshots/Demos

Screenshot 2024-05-31 at 4 47 27 PM


app/models/pet.rb
```diff
@@ -95,6 +97,12 @@ def append_images=(attachables)
     images.attach(attachables)
   end
 
+  def sensible_placement_type
+    if matches.where(end_date: DateTime.now..).exists? && placement_type == "Adoptable"
+      errors.add(:placement_type, "must include Fosterable")
```
kasugaijin, line 102: Yes I think we can make this error more explicit. Something like 'Cannot change placement type to adoptable on pet with active foster'

Also - please add this to the en.yml and use translation. I think you need to call the full i18n method to access in the model.

kasugaijin (approved):

Issue #707, opened 2024-05-14
Staff Dashboard Metrics: As staff I can see a tile for overdue tasks and click it to render a table of pets with overdue tasks


kasugaijin:
On the staff dashboard index page we can add value by showing a quick glance of metrics, and allowing staff to quickly access work that needs to be done. 

In this issue, we want to modify one of the existing tiles to show information on Pet Tasks. Specifically, show the total number of tasks that are not complete AND not overdue, and the number of tasks that are overdue, on all pets. The staff should be able to click on each stat in the tile, and using hotwire, we render a table below of Pets for which these tasks belong. For example, if I click 'Overdue', the table is populated with Pets with overdue tasks. Let's create a table that shows the Pet image, name, and a column for the number of tasks. The staff should be able to click this pet name and be redirected to that pet's show page.

This is a rough representation of the requirements (ignore the columns to the right of the pet name, these are irrelevant). 
![lol](https://github.com/rubyforgood/pet-rescue/assets/95949082/a5dc3a98-ab6b-4091-a93d-59cfd37332b0)

- On the index action we will want to generate the stats for overdue tasks and incomplete tasks (not overdue) to show in the tile
- Make two new actions on the staff dashboard controller to handle generating the responses for click the Overdue and Not overdue links in the tile, you will need to get the appropriate data for the table, including Pet image, name, and number of tasks (incomplete and not overdue, or overdue)
- Add tests to make sure the collections generated for each route contain what we expect
- The table should be paginated (I'd say 5 rows per page?)

There's a lot happening here, so let me know if you have questions!

sarvaiyanidhi:
Hi @kasugaijin I can take up this issue. Can you please assign it to me?

jmilljr24:
@sarvaiyanidhi all yours!

sarvaiyanidhi:
Just a quick update that I will work on this now from this week and will try push asap. sorry for the delay.

jmilljr24:
> Just a quick update that I will work on this now from this week and will try push asap. sorry for the delay.

No worries @sarvaiyanidhi !! Thank you for the update.

kasugaijin:
Absolutely no rush!

PR #787, opened 2024-05-31
707 Dashboard metric for Incomplete and Overdue tasks

sarvaiyanidhi:
# 🔗 Issue
#707 

# ✍️ Description
- Added new scopes in task model to get not_overdue and overdue tasks
- Updated Dashboard controller index method to calculate number of incomplete and overdue tasks in organization
- Added two new methods in dashboard controller to fetch for each pet number of incomplete and overdue tasks
- In dashboard index view, added metric to display pending and over tasks counts
- On click of each metric, we render pet table details using turbo frame

# 📷 Screenshots/Demos

Dashboard Page

Alta-Pet-Rescue-Pet-Rescue-dashboard

Dashboard Page with Incomplete Tasks

Alta-Pet-Rescue-Pet-Rescue-dashboard-pending-tasks

Dashboard Page with Overdue Tasks

Alta-Pet-Rescue-Pet-Rescue-dashboard-overdue-tasks

### DIFF
--- a/app/controllers/organizations/staff/dashboard_controller.rb
+++ b/app/controllers/organizations/staff/dashboard_controller.rb
@@ -12,7 +12,6 @@ class Organizations::Staff::DashboardController < Organizations::BaseController
   end
 
   def incomplete_tasks
-    @organization = Current.organization
     @pagy, @pets = pagy(
       Pet
       .left_joins(:tasks)
@@ -32,7 +31,6 @@ class Organizations::Staff::DashboardController < Organizations::BaseController
   end
 
   def overdue_tasks
-    @organization = Current.organization
     @pagy, @pets = pagy(
       Pet
       .left_joins(:tasks)
--- a/app/controllers/organizations/staff/dashboard_controller.rb
+++ b/app/controllers/organizations/staff/dashboard_controller.rb
@@ -5,6 +5,9 @@ class Organizations::Staff::DashboardController < Organizations::BaseController
     @user = current_user
     @organization = Current.organization
     @hide_footer = true
+    pet_ids = @organization.pets.pluck(:id)
+    @not_completed_not_overdue_tasks_count = Task.where(pet_id: pet_ids).is_not_completed.not_overdue.count
+    @not_completed_overdue_tasks_count = Task.where(pet_id: pet_ids).is_not_completed.overdue.count
 
     authorize! :dashboard, context: {organization: @organization}
   end
--- a/app/controllers/organizations/staff/dashboard_controller.rb
+++ b/app/controllers/organizations/staff/dashboard_controller.rb
@@ -17,18 +17,18 @@ class Organizations::Staff::DashboardController < Organizations::BaseController
     @pagy, @pets = pagy(
       @organization.pets
                    .left_joins(:tasks)
-                   .select('pets.*, COUNT(tasks.id) AS incomplete_tasks_count')
-                   .where(tasks: { completed: false })
-                   .where('tasks.due_date IS NULL OR tasks.due_date >= ?', Time.current)
-                   .group('pets.id'),
+                   .select("pets.*, COUNT(tasks.id) AS incomplete_tasks_count")
+                   .where(tasks: {completed: false})
+                   .where("tasks.due_date IS NULL OR tasks.due_date >= ?", Time.current)
+                   .group("pets.id"),
       items: 5
     )
     @column_name = "Incomplete Tasks"
     respond_to do |format|
       format.turbo_stream do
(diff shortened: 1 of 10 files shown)


app/controllers/organizations/staff/dashboard_controller.rb
```diff
@@ -1,11 +1,61 @@
 class Organizations::Staff::DashboardController < Organizations::BaseController
+  before_action :context_authorize!, only: %i[index incomplete_tasks overdue_tasks]
+  include Pagy::Backend
   layout "dashboard"
 
   def index
     @user = current_user
     @organization = Current.organization
     @hide_footer = true
+    pet_ids = @organization.pets.pluck(:id)
```
kasugaijin, line 10: Here we can use Pet.all.pluck(:id) also as queries should be scoped to the tenant already.

app/controllers/organizations/staff/dashboard_controller.rb
```diff
@@ -1,11 +1,61 @@
...
+    @not_completed_not_overdue_tasks_count = Task.where(pet_id: pet_ids).is_not_completed.not_overdue.count
```
kasugaijin, line 11: So for here, can't we call Task.is_not_completed.not_overdue.count directly? I think this should be querying the tenant scope as this controller inherits from Organizations::BaseController, which includes this module where we set the tenant https://github.com/rubyforgood/pet-rescue/blob/main/app/controllers/concerns/organization_scopable.rb

app/controllers/organizations/staff/dashboard_controller.rb
```diff
@@ -1,11 +1,61 @@
...
+    @not_completed_overdue_tasks_count = Task.where(pet_id: pet_ids).is_not_completed.overdue.count
+  end
+
+  def incomplete_tasks
+    @organization = Current.organization
+    @pagy, @pets = pagy(
+      @organization.pets
```
kasugaijin, line 18: Same here I think we can query Pet instead of @organization.pets

app/controllers/organizations/staff/dashboard_controller.rb
```diff
@@ -1,11 +1,61 @@
...
+        render turbo_stream: turbo_stream.replace("tasks-frame", partial: "organizations/staff/dashboard/tasks", locals: {column_name: @column_name})
+      end
+      format.html { render :tasks, locals: {column_name: @column_name} }
+    end
+  end
+
+  def overdue_tasks
+    @organization = Current.organization
+    @pagy, @pets = pagy(
+      @organization.pets
```
kasugaijin, line 38: And here

kasugaijin:
@sarvaiyanidhi looking great! I just have some comments on how we handle the queries. I think we should be able to call on the model directly as we should be executing within the tenant scope already.

sarvaiyanidhi:
> @sarvaiyanidhi looking great! I just have some comments on how we handle the queries. I think we should be able to call on the model directly as we should be executing within the tenant scope already.

Thank you for the feedback @kasugaijin .. Have pushed updated code by removing organization from query.

app/controllers/organizations/staff/dashboard_controller.rb
```diff
@@ -1,11 +1,60 @@
...
 
   def index
     @user = current_user
     @organization = Current.organization
     @hide_footer = true
+    @not_completed_not_overdue_tasks_count = Task.is_not_completed.not_overdue.count
+    @not_completed_overdue_tasks_count = Task.is_not_completed.overdue.count
+  end
+
+  def incomplete_tasks
```
kasugaijin, line 14: Does this instance variable get used? @organization

sarvaiyanidhi: My bad .. it was used earlier but not required now. Have removed and pushed code.

app/controllers/organizations/staff/dashboard_controller.rb
```diff
@@ -1,11 +1,58 @@
 class Organizations::Staff::DashboardController < Organizations::BaseController
+  before_action :context_authorize!, only: %i[index incomplete_tasks overdue_tasks]
+  include Pagy::Backend
   layout "dashboard"
 
   def index
     @user = current_user
     @organization = Current.organization
```
kasugaijin, line 8: And this @organization? :)

sarvaiyanidhi: This was already part of existing code so did not remove it and I see it is currently used in view https://github.com/rubyforgood/pet-rescue/blob/main/app/views/organizations/staff/dashboard/index.html.erb#L17

kasugaijin: Ah yes ok thank you.

kasugaijin (approved):
Thank you!

PR #791, opened 2024-06-01
737 all controllers i18n revision

wandergithub:
# 🔗 Issue

#737 
# ✍️ Description

1. Visited all controllers and set the lookup function with translations.
2. added translations.
3. solve mission translations on views
4. change hardcoded enum translation for new enum method previously added on Application Record #738 .

# 📷 Screenshots/Demos


app/controllers/organizations/adopter_fosterer/adopter_applications_controller.rb
```diff
@@ -26,7 +26,7 @@ def create
         .new_adoption_application.deliver_now
     else
       redirect_to adoptable_pet_path(@pet),
-        alert: "Error. Please try again."
+        alert: t(".error")
```
kasugaijin, line 29: So this finds the appropriately scoped translation? I have always called the full reference to it and did not know this would work!

kasugaijin (approved):
LGTM - Just the one question bout the references. If that's all good, then merge away.

app/controllers/organizations/adopter_fosterer/adopter_applications_controller.rb
wandergithub: Yes, it's called lazy lookup. Important to notice that it is not ALWAYS suitable. Sometimes we have to use the full path. In tests or module helpers requires `I18n.t("full/path"). 

ATE I18n rails documentation for this moment🤣

kasugaijin (approved):
Ah cool thank you for the info

wandergithub:
@kasugaijin Once this is merged and I know it's all good I'll work on views so we can close the related issue.

kasugaijin:
@wandergithub press that green button when you're ready :)

wandergithub:
@kasugaijin I don't have permission to merge.

Issue #709, opened 2024-05-14
Default Pet Tasks: tasks should be assigned to a general category, or a species


kasugaijin:
Currently default pet tasks are assigned to all new pets, regardless of species. This is not as practical as grouping default pet tasks by species. We can also have a general category, that applies to all species. 

- On the Pet model, add an enum at 0 index `all`
- In the form used to update/create a pet, do not show the `all` option in the select
- Add a new column to the DefaultPetTasks table, `species`, `string`, default `0` (the `all` enum)
- Add this new attribute to the Default Pet task form as a select showing all enums for pet species
- Update the default pet task index table in the UI to show species column
- Organize default pet tasks in the index alphabetically on species
- Update the default pet task service so that when a new pet is created, it's associated tasks will be the tasks with `all` AND the same species as the Pet. E.g., A new dog will get tasks with `all` and `dog` species. 
- Update tests

atbalaji:
Hi @kasugaijin, I have raised 2 PRs . After they are over and merged, i would like to work on this issue

kasugaijin:
@atbalaji once both PRs are merged, if this is still not assigned, for sure! But just in case someone else wants it before then we don't hold issues for people to give a fair chance to all to contribute.

atbalaji:
Hi @kasugaijin , I would like to take this issue up as now both of my PRs are merged

kasugaijin:
@atbalaji all yours

atbalaji:
Just for confirmation @kasugaijin Can you elaborate the first point. U mean to create an enum like named 'species' where 0 represents 'all' option ? That's what u mean ?

kasugaijin:
@atbalaji check the Pet model as there is already an enum. Just need to add an ‘all’ at the zero index of that enum.

atbalaji:
Hi @kasugaijin .I have made changes except for tests part locally. Some differences with the requirement, we shall discuss during PR. But one thing i noticed during the changes is since default pet tasks are not associated with the tasks they create. On deletion of default pet task the corresponding task of a pet is not deleted. Is this the intended behavior?

kasugaijin:
Thanks @atbalaji yes that’s fine. The default pet task is a blueprint for pet tasks but does not need to be associated.

kasugaijin:
Yes for sure you can make a PR without tests and add them later

wiliajc87:
[...]

PR #798, opened 2024-06-02
Added species to default pet tasks and handled related code

atbalaji:
# 🔗 Issue

https://github.com/rubyforgood/pet-rescue/issues/709

# ✍️ Description

 - Added species related filtering to default pet tasks

# 📷 Screenshots/Demos

Screenshot 2024-06-02 164554
Screenshot 2024-06-02 164609

### DIFF
--- a/app/models/default_pet_task.rb
+++ b/app/models/default_pet_task.rb
@@ -7,6 +7,7 @@
 #  due_in_days     :integer
 #  name            :string           not null
 #  recurring       :boolean          default(FALSE)
+#  species         :integer
 #  created_at      :datetime         not null
 #  updated_at      :datetime         not null
 #  organization_id :bigint           not null
@@ -24,4 +25,6 @@ class DefaultPetTask < ApplicationRecord
 
   validates :name, presence: true
   validates_numericality_of :due_in_days, only_integer: true, greater_than_or_equal_to: 0, allow_nil: true
+
+  enum species: {"All" => nil}.merge(Pet.species)
 end
--- a/app/views/organizations/staff/default_pet_tasks/index.html.erb
+++ b/app/views/organizations/staff/default_pet_tasks/index.html.erb
@@ -16,6 +16,7 @@
               <tr>
                 <th scope="col">Name</th>
                 <th scope="col">Description</th>
+                <th scope="col">Species</th>
                 <th scope="col">Due After</th>
                 <th scope="col">Recurring</th>
                 <th scope="col"></th>
@@ -29,6 +30,9 @@
                   </td>
                   <td>
                     <%= task.description %>
+                  </td>
+                  <td>
+                    <%= task.species %>
                   </td>
                   <td>
                     <%= task.due_in_days ? pluralize(task.due_in_days, "day") : "" %>
--- a/app/controllers/organizations/staff/default_pet_tasks_controller.rb
+++ b/app/controllers/organizations/staff/default_pet_tasks_controller.rb
@@ -5,7 +5,7 @@ class Organizations::Staff::DefaultPetTasksController < Organizations::BaseContr
   layout "dashboard"
 
   def index
-    @default_pet_tasks = authorized_scope(DefaultPetTask.all)
+    @default_pet_tasks = authorized_scope(DefaultPetTask.all).sort_by { |task| task.species }
   end
 
   def new
@@ -43,7 +43,7 @@ class Organizations::Staff::DefaultPetTasksController < Organizations::BaseContr
   private
 
   def task_params
-    params.require(:default_pet_task).permit(:name, :description, :due_in_days, :recurring)
+    params.require(:default_pet_task).permit(:name, :description, :due_in_days, :species, :recurring)
   end
 
   def set_task
--- a/app/services/organizations/default_pet_task_service.rb
+++ b/app/services/organizations/default_pet_task_service.rb
@@ -2,7 +2,7 @@
 class Organizations::DefaultPetTaskService
   def initialize(pet)
     @pet = pet
-    @default_pet_tasks = DefaultPetTask.all
+    @default_pet_tasks = DefaultPetTask.where(species: DefaultPetTask.species.values_at("All", pet.species))
   end
 
   def default_tasks_array
--- a/app/views/organizations/staff/default_pet_tasks/_form.html.erb
+++ b/app/views/organizations/staff/default_pet_tasks/_form.html.erb
@@ -8,6 +8,12 @@
         <%= f.text_area :description, class: 'form-control', rows: 3 %>
       </div>
       <div class="form-group">
+        <%= f.select :species,
+                     options_for_select(DefaultPetTask.species.keys, task.species),
+                     { prompt: t('general.please_select') },
+                     { class: 'form-control' } %>
+      </div>
+      <div class="form-group">
         <%= f.number_field :due_in_days, class: 'form-control', help: 'The task will be due this number of days after the pet is created.' %>
       </div>
       <div class="form-group">
--- a/db/migrate/20240529060957_add_species_to_default_pet_tasks.rb
+++ b/db/migrate/20240529060957_add_species_to_default_pet_tasks.rb
@@ -1,0 +1,5 @@
+class AddSpeciesToDefaultPetTasks < ActiveRecord::Migration[7.1]
+  def change
+    add_column :default_pet_tasks, :species, :integer
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_05_19_224035) do
+ActiveRecord::Schema[7.1].define(version: 2024_05_29_060957) do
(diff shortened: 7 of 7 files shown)


app/models/default_pet_task.rb
```diff
@@ -24,4 +25,6 @@ class DefaultPetTask < ApplicationRecord
 
   validates :name, presence: true
   validates_numericality_of :due_in_days, only_integer: true, greater_than_or_equal_to: 0, allow_nil: true
+
+  enum species: {"All" => nil}.merge(Pet.species)
```
jmilljr24, line 29:
```suggestion
    enum :species, ["All"].push(Pet.species.keys).flatten, default: 0
```
I think using`"All" => 0` as mention in the issue would more make sense. Using nil seems to go against the implicit value's and I'm not sure what we gain from that.

app/views/organizations/staff/default_pet_tasks/index.html.erb
```diff
@@ -16,6 +16,7 @@
               <tr>
                 <th scope="col">Name</th>
                 <th scope="col">Description</th>
+                <th scope="col">Species</th>
                 <th scope="col">Due After</th>
                 <th scope="col">Recurring</th>
```
jmilljr24, line 21: Could you add translations for these?

app/views/organizations/staff/default_pet_tasks/index.html.erb
```diff
@@ -30,6 +31,9 @@
                   <td>
                     <%= task.description %>
                   </td>
+                  <td>
+                    <%= task.species %>
```
jmilljr24, line 35:
```suggestion
                    <%= task.human_enum_name(:species) %>
```
We have a method to translate enums. Update `en.yml` :
```
activerecord:
  attributes:
    default_pet_tasks:
      species:
        All: "All"
        .......
```

app/views/organizations/staff/default_pet_tasks/index.html.erb
```diff
@@ -30,6 +31,9 @@
...
+                  </td>
                   <td>
                     <%= task.due_in_days ? pluralize(task.due_in_days, "day") : "" %>
```
jmilljr24, line 38: Translation

jmilljr24 (changes requested):
@atbalaji Thanks for working on this!! Just a couple of things. Let me know your thoughts/concerns.

app/models/default_pet_task.rb
atbalaji: The reason for using nil here is to maintain the consistency between the "species" enum in Pet model where Dogs have 0 and Cats have 1. Also I did not opt for modifying the enum "species" in Pet model and went with a new one in DefaultPetTask is because adding All as part of pet enum would not be consistent when viewed as a potential value in a single pet instance. So I went with DefaultPetTask. So keeping "All" as nil felt essential in my thought process. Love to hear your take.

app/views/organizations/staff/default_pet_tasks/index.html.erb
atbalaji: OK

app/views/organizations/staff/default_pet_tasks/index.html.erb
atbalaji: OK

app/models/default_pet_task.rb
jmilljr24: I see where you are coming from. I didn't think about the value matching on `Pets` and `DefaultPetTasks`. Although the value should rarely if ever be called, would this cause any confusion for future development? Maybe @kasugaijin or @mononoken might have some thoughts to add.

Current
```
Pet.species: {"Dog"=>0, "Cat"=>1} #implicit
DefaultPetTask.species: {"All"=>nil, "Dog"=>0, "Cat"=>1} #explicit
```

Suggestion (original issue criteria)
```
Pet.species: {"Dog"=>0, "Cat"=>1} #implicit
DefaultPetTask.species: {"All"=>0, "Dog"=>1, "Cat"=>2} #implicit
```

Third option
```
Pet.species: {"Dog"=>1, "Cat"=>2} #Explicit to have the same values as DefaultPetTask without relying on a nil value
DefaultPetTask.species: {"All"=>0, "Dog"=>1, "Cat"=>2} #implicit
```
Not part of the issue, but one thought came to mind was the potential need in the future to have a  `none` option. Maybe an admin does not want to delete a `DefaultPetTask` but they do not want it being added to a category temporarily. I think using the nil option in that scenario would make more sense.

kasugaijin: Ah, this is a good question and harps back on a discussion we had a while back about abstracting the concept of pet species into its own entity to prevent issues like this. We decided on keeping the species on Pet as the source of truth, but this PR identifies the issue with doing that...it causes coupling of models where it ought not to exist. I think the easiest way to fix this for now is to match the enums between the two models, per Justin's third option. The fact we have to change another model to facilitate this underlines the issue...but it should be a quick fix (maybe check the seed files too to update them if they use the integers to set species). Do you agree?

app/views/organizations/staff/default_pet_tasks/index.html.erb
atbalaji: Done

jmilljr24:
@atbalaji just checking in. Not sure if you saw the feedback from @kasugaijin .

kasugaijin:
@atbalaji how's this going?

atbalaji:
Hi ,addressed the comments. Had some conflicts after pulling. Will work on those and tests and push within tmr

kasugaijin:
Sounds good @atbalaji

atbalaji:
Current update: Updated failed tests. But facing issue with a page which worked before addressing comments correctly. New error. Investigating it

app/views/organizations/staff/default_pet_tasks/index.html.erb
atbalaji: I18n::InvalidPluralizationData at /staff/default_pet_tasks/13/edit
translation data {:All=>"All", :Dog=>"Dog", :Cat=>"Cat"} can not be used with :count => 1. key 'one' is missing. Due to the above change I'm getting this error in the bootstrap_form in default_pet_tasks form partial :species select.  
`  Model.human_attribute_name(:species) `  on enums returns same error which i think bootstrap_form_with internally works using this.

atbalaji: > I18n::InvalidPluralizationData at /staff/default_pet_tasks/13/edit translation data {:All=>"All", :Dog=>"Dog", :Cat=>"Cat"} can not be used with :count => 1. key 'one' is missing. Due to the above change I'm getting this error in the bootstrap_form in default_pet_tasks form partial :species select. ` Model.human_attribute_name(:species)` on enums returns same error which i think bootstrap_form_with internally works using this.

.

atbalaji:
I have updated the issue reason as reply to the earlier comments. Pls check it out.

app/views/organizations/staff/default_pet_tasks/index.html.erb
atbalaji: `<%= f.select :species,
                     options_for_select(DefaultPetTask.species.keys, task.species),
                     { prompt: t('general.please_select'), label: "Species" },
                     { class: 'form-control' } %>` . We can have some fix like this by passing label option which i feel is not satisfactory. But still Model.human_attribute_name is a normally used rails method which is failing for enum attributes which have been added as part of human_enum_name in YAML files. We can check for existing keys in console. This is my dilemma. Would like to hear your thoughts.

jmilljr24: In `en.yml` under species add `one: "Species"`

I think we should also add a new method that returns all the translations for a specific model enum. It would simplify the options for select.
```
<%= f.select :species,
        options_for_select(DefaultPetTask.human_enum_names(:species), task.human_enum_name(:species)), 
       { prompt: t('general.please_select') }, { class: 'form-control' } %>
```

In `application_record.rb` we could do something like
```
  def self.human_enum_names(enum)
    enum_i18n_key = enum.to_s
    names = []
    send(enum).each_key do |key|
      names << I18n.t("activerecord.attributes.#{model_name.i18n_key}.#{enum_i18n_key}.#{key}")
    end
    names
  end
```
Feel free to adjust it.

kasugaijin:
@atbalaji FYI you might want to rebase and force push to get rid of all of the past commit history. This makes reviewing really hard!

jmilljr24:
@atbalaji should be good to go for adding tests as well. Thanks for the changes!!

atbalaji:
Sorry for the trouble @kasugaijin and @jmilljr24 . During rebase some mishap happened due to a single commit unexpectedly which caused this. Also I have risen a new PR with a different branch name using merging for the clarity of reviewing.  I am traveling for next 3 days. So i won't be available for next 3 days.  I have updated failed tests. Regarding addition of new tests, pls assign it to someone in case you need it earlier. Sorry for the inconvenience.

kasugaijin:
@atbalaji thank you…no sweat…it happens! No rush to get this across the line either. Don’t want you to feel pressured to get stuff in.

jmilljr24:
Opened in new PR #879

PR #824, opened 2024-06-09
Dockerize Pet Rescue

kasugaijin:
Dockerizing the app to help lower the barrier to getting up and running for contributors. I opted to just use a dockerfile with docker-compose and rely on docker commands to get up and running - not using a devcontainer for now (I do not use VSCode).

With these changes so far, the app works as expected in Docker. However:
- System tests are failing still. I added Chromium to the base image, but now getting a ferrum timeout error. 
- I was having asset delivery issues, so I run the command to precompile assets, though I know this can slow boot times down a bit
- What would be ideal to include in a dockerignore file? 

# 🔗 Issue

# ✍️ Description

# 📷 Screenshots/Demos

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,9 +9,25 @@ The Pet Rescue app is derived from the [Baja Pet Rescue Dog Adoption Application
 
 # 🚀 Getting Started
 
-Let's get your machine setup to startup the application!
+Let's get your machine setup to start up the application!
 
-## Prerequisites
+## Docker
+
+Install a containerization app, such as Docker Desktop.
+
+Clone the repo to your local machine and navigate to the directory and run:
+
+`docker-compose build` to build the base image.
+
+`docker-compose up`  to start the containers. You can add the `-d` flag to run silently without logging.
+
+`docker-compose run --rm app bin/setup` to set up and seed the database.
+
+If you need to run migrations at all: `docker-compose run --rm app bin/rails db:migrate`
+
+Visit `localhost:3000` in your browser.
+
+## Local Installation
 
 ⚠️  We assume you already have ruby installed with your preferred version manager. This codebase supports [rbenv](
 https://github.com/rbenv/rbenv) and [asdf](https://github.com/asdf-vm/asdf-ruby).
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,20 +1,27 @@
----
+version: '3.8'
 services:
-  postgres:
-    image: postgres:14
-    environment:
-      POSTGRES_USER: ${DATABASE_USERNAME}
-      POSTGRES_PASSWORD: ${DATABASE_PASSWORD}
-      POSTGRES_DB: pet_rescue_development
-
-    ports:
-      - 5432
-
-    healthcheck:
-      test: ["CMD-SHELL", "pg_isready", "-U", "bwoa", "-d", "pet_rescue_development"]
-      interval: 10s
+  db:
+    image: postgres:13
     volumes:
-      - db:/var/lib/postgresql/data
+      - postgres-data:/var/lib/postgresql/data
+    environment:
+      POSTGRES_DB: myapp_development
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: password
+
+  app:
+    build: .
+    command: bash -c "rm -f tmp/pids/server.pid && bundle exec rails s -b '0.0.0.0'"
+    volumes:
+      - .:/myapp
+    ports:
+      - "3000:3000"
+    depends_on:
+      - db
+    environment:
+      DATABASE_URL: postgres://postgres:password@db:5432/myapp_development
+      RAILS_ENV: development
 
 volumes:
-  db:
+  postgres-data:
+  bundle:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,6 @@
 version: '3.8'
 services:
-  db:
+  postgres:
     image: postgres:14
     volumes:
       - postgres-data:/var/lib/postgresql/data
@@ -22,10 +22,10 @@ services:
     ports:
       - "3000:3000"
     depends_on:
-      db:
+      postgres:
         condition: service_healthy
     environment:
-      DATABASE_URL: postgres://${DATABASE_USERNAME:-postgres}:${DATABASE_PASSWORD:-password}@db:5432/petrescue_development
+      DATABASE_URL: postgres://${DATABASE_USERNAME:-postgres}:${DATABASE_PASSWORD:-password}@postgres:5432/petrescue_development
       RAILS_ENV: development
 
   sass:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,8 +6,13 @@ services:
       - postgres-data:/var/lib/postgresql/data
     environment:
       POSTGRES_DB: myapp_development
-      POSTGRES_USER: postgres
-      POSTGRES_PASSWORD: password
+      POSTGRES_USER: ${DATABASE_USERNAME:-postgres}
+      POSTGRES_PASSWORD: ${DATABASE_PASSWORD:-password}
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U ${DATABASE_USERNAME:-postgres} -d myapp_development"]
+      interval: 10s
+      timeout: 5s
+      retries: 5
 
   app:
     build: .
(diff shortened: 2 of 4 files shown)


docker-compose.yml
```diff
@@ -1,20 +1,27 @@
...
     environment:
-      POSTGRES_USER: ${DATABASE_USERNAME}
-      POSTGRES_PASSWORD: ${DATABASE_PASSWORD}
-      POSTGRES_DB: pet_rescue_development
-
-    ports:
-      - 5432
+      POSTGRES_DB: myapp_development
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: password
```
cflipse, line 10: it's good to go ahead and set a default value in here, but we still will want to allow overriding these values by setting environment variables.

There's a not-uncommon pattern of running support services, like postgres and redis, via a dockerfile, and even running the local dev server within docker, but still running tests in the non-docker context.  Respecting set envvars keeps friction lower.

```
POSTGRES_USER: ${DATABASE_USERNAME:-postgres}
```
ought to work

docker-compose.yml
```diff
@@ -1,20 +1,27 @@
...
 
-    healthcheck:
-      test: ["CMD-SHELL", "pg_isready", "-U", "bwoa", "-d", "pet_rescue_development"]
-      interval: 10s
```
cflipse, line 15: please don't nuke the healthchecks.

docker-compose.yml
```diff
@@ -1,20 +1,27 @@
...
+    depends_on:
+      - db
+    environment:
+      DATABASE_URL: postgres://postgres:password@db:5432/myapp_development
+      RAILS_ENV: development
 
 volumes:
-  db:
+  postgres-data:
+  bundle:
```
cflipse, line 27: you're not acutally using the bundle volume

dockerfile
```diff
@@ -0,0 +1,36 @@
+# Use the official Ruby 3.3.0 image as a base
+FROM ruby:3.3.0-slim
+
+# Install dependencies
+RUN apt-get update -qq && apt-get install -y \
+  nodejs \
+  postgresql-client \
+  build-essential \
+  libpq-dev \
+  chromium-driver
```
cflipse, line 10: So, this ends up getting into various deployement philosophies, and a lot depends on _how_ things are ultimately deployed, but ...

if deploying via a docker container (such as, with `kamal` ) then you really don't want to be shipping the image with heavyweight development dependencies like chromium.  There are a couple of ways that this can be dodged around, with a multistage docker build, or a separate dockerfile for dev/test vs production, but the smaller the image that gets deployed the better.

Happy to run through setting up a multistage dockerfile here, if that's all greek.

entrypoint.sh
```diff
@@ -0,0 +1,14 @@
...
+set -e
+
+# Remove a potentially pre-existing server.pid for Rails.
+rm -f /myapp/tmp/pids/server.pid
+
+# Ensure all gems are installed.
+bundle check || bundle install
+
+# Precompile assets (optional for development, but useful for production)
+bundle exec rails assets:precompile
```
cflipse, line 11: This is better done within the dockerfile.

cflipse: And the thing that you _aren't_ doing in the entrypoint, that ends up being important for a production deploy, is database migrations.

it's probably worth checking out the default dockerfile that rails 7.1+ generates now.  There are some tricks at play.

cflipse:
Also, you really probably want to also have the docker container launch the javascript / assets process as well.  The line in the Procfile that looks like:  `css: bin/rails dartsass:watch`  

basically, if you're running a full docker-compose container, you can fully avoid using the Procfile structure (`foreman` is quite a bit older than docker, and can be supplanted, it turns out)

That may also address what you were seeing and trying to address with the asset precompile, but there may be some port-mapping things to wrangle....

docker-compose.yml
kasugaijin: Ah I like this.

docker-compose.yml
kasugaijin: Ah ok.

dockerfile
kasugaijin: Right that's a good point. I was thinking for now this is only for dev/test. I am awaiting more details on deployment to product. All I know is we are going to use cloud66. I don't think this image size is as big of an issue if we are focusing on dev/test for now?

entrypoint.sh
kasugaijin: Ah yes same for here - I am not sure we are using this for production deployments right now. I was thinking this is for dev/test to get a developer up and running only. I need to read up more on deploying with docker - I am not sure why we would need to do this in the first place!

kasugaijin:
> css: bin/rails dartsass:watch

Ah right thank you for catching that. I have added a SASS container.

dockerfile
kasugaijin: I think there is more discussion required with the team on deployment preferences. Docker seems like more work, but for a benefit. The key thing is that we have capacity to create and maintain such infrastructure.

Do you forsee an issue with thie docker environment being used purely for dev/test for now, and keeping the deployment process to be determined?

docker-compose.yml
```diff
@@ -1,20 +1,41 @@
...
-  postgres:
+  db:
     image: postgres:14
+    volumes:
+      - postgres-data:/var/lib/postgresql/data
     environment:
-      POSTGRES_USER: ${DATABASE_USERNAME}
-      POSTGRES_PASSWORD: ${DATABASE_PASSWORD}
-      POSTGRES_DB: pet_rescue_development
+      POSTGRES_DB: myapp_development
```
cflipse, line 8:
```suggestion
      POSTGRES_DB: petrescue_development
```

`myapp` screams demoware.  :laughing:

docker-compose.yml
```diff
@@ -1,20 +1,41 @@
...
+      POSTGRES_USER: ${DATABASE_USERNAME:-postgres}
+      POSTGRES_PASSWORD: ${DATABASE_PASSWORD:-password}
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U ${DATABASE_USERNAME:-postgres} -d myapp_development"]
```
cflipse, line 12:
```suggestion
      test: ["CMD-SHELL", "pg_isready -U ${DATABASE_USERNAME:-postgres} -d petrescue_development"]
```

docker-compose.yml
```diff
@@ -1,20 +1,41 @@
...
+      timeout: 5s
+      retries: 5
 
+  app:
+    build: .
+    command: bash -c "rm -f tmp/pids/server.pid && bundle exec rails s -b '0.0.0.0'"
+    volumes:
+      - .:/myapp
     ports:
-      - 5432
```
cflipse, line 11: Ohh, this needs to stick with the database container -- again, exposure for local console development.

dockerfile
cflipse: No, no particular issue with holding off for the moment.

I'll note that dockerized production deployment _does_ save an extrordinary amount of headache when trying to keep a server updated.  Sure, it can be a little bit more complex at times, but on the other hand, you'll never OOM crash your production server while trying to compile a new version of ruby for deployment.... 

as for capacity, there are enough of us ops type folks hanging around r4g that we can cover most of the basic questions, and this is becoming much, much more common than it's been in years past.

entrypoint.sh
cflipse: If you're deploying the image -- the precompile _should_ be done during the docker build, not during startup.

It's less of an issue if you're just using importmaps, I _think_  (I'm not overly familiar with how they work), but generally the pattern is to use a build container to precompile your assets, and then copy the assets over to the production image.  Rationale is that the production image then _only_ needs to ship with ruby/database libraries, and doens't need to cart along a node installation and 400+ `node_modules` packages ...

which isn't _just_ deployment image size, it's also attack area.

dockerfile
kasugaijin: Ok good to note. I think I will lean towards dockerized prod deployment - it seems to be liked by those who use it from my perusals. I will need to read a bit more to make an informed decision.

entrypoint.sh
kasugaijin: Ok I 'think' that we are currently doing this by calling asset precompile command in the dockerfile, no?

docker-compose.yml
```diff
@@ -1,20 +1,41 @@
----
+version: '3.8'
 services:
-  postgres:
+  db:
```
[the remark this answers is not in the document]
kasugaijin: Ah yes. Now using `postgres` instead of `db`

entrypoint.sh
cflipse: yep

kasugaijin:
@cflipse So I think this is good to go for now until we need to consider configration for deployments, right? I added some instructions for setting up local dev environment to the readme.

kasugaijin:
@cflipse bump

README.md
```diff
@@ -9,9 +9,25 @@ The Pet Rescue app is derived from the [Baja Pet Rescue Dog Adoption Application
...
+Let's get your machine setup to start up the application!
 
-## Prerequisites
+## Docker
+
+Install a containerization app, such as Docker Desktop.
+
+Clone the repo to your local machine and navigate to the directory and run:
+
+`docker-compose build` to build the base image.
```
cflipse, line 20: these days, the command is `docker compose`  (no dash).  It's a newer version of the subcommand

cflipse (approved):

PR #825, opened 2024-06-09
Generate a person model

cflipse:
# ✍️ Description

A person represents a flesh-and-blood human in the system, and should be
the core data model representing _most_ actors in the system.

Eventually, we will want to attach the filled form results to a person
record, and Accounts should belong to a person to grant the ability to
log into the system.

Note that this table explicitly does _not_ include any authentication
related material.  tokens generated for things like magic login, or
authorization credentials for regular users, should exist in their own
dedicated classes.  I'll follow up with an email+token based auth flow
implementation later.

### DIFF
--- a/app/models/concerns/avatarable.rb
+++ b/app/models/concerns/avatarable.rb
@@ -4,6 +4,7 @@ module Avatarable
   included do
     has_one_attached :avatar
 
+    # TODO:  move these validation strings to a locale file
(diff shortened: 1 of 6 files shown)


kasugaijin (approved):
Looks good @cflipse

kasugaijin:
@all-contributors
Please add @cflipse for code.

kasugaijin:
@all-contributors
Please add @cflipse for code.

Issue #823, opened 2024-06-08
Rename AdopterApplication to CustomForm::Submission


kasugaijin:
AdoptionApplication should be renamed to CustomForm::Submission. This means renaming all areas in the app (table (submissions), model, controller, policies, views, tests etc.). It should be a relatively straightforward find and replace.

Note the table should be renamed to `submissions_table`. The CustomForm scope helps to organize this resource along with others that are similar.

Any questions - comment below!

MooseCowBear:
I can do this

MooseCowBear:
@kasugaijin Should the methods with 'application' in their names also be renamed? What about all the variables? And the AdoptionApplicationReviewsController?

kasugaijin:
@MooseCowBear yes ideally we replace all for consistency. I think search and replace should make this pretty straightforward!

kasugaijin:
@MooseCowBear please nest the classes instead of inline: https://thoughtbot.com/blog/why-you-should-nest-modules-in-ruby we are trying to move to the nested format throughout.

PR #837, opened 2024-06-13
823 rename adopter application

MooseCowBear:
# 🔗 Issue
#823 

# ✍️ Description

Renames adopter applications table to submissions and updates references to application throughout. 

After I switched to nested modules for the Organizations::AdopterFosterer::CustomForm::SubmissionsController, I realized the comment only mentioned nesting the model. But I assume we will also want to use nested modules for the controllers, is that right? If not, I can change it. 

Also right now, the reviews path/controller is just submissions_reviews but not namespaced under custom form. I can change that too. Let me know.

# 📷 Screenshots/Demos


app/controllers/organizations/adopter_fosterer/custom_form/submissions_controller.rb
```diff
@@ -0,0 +1,63 @@
+module Organizations
```
kasugaijin, line 1: Ahhh nested modules...so much better

kasugaijin (approved):
Phew that was a lot of work to change...but a good one to get done. Thank you for the attention to detail.

kasugaijin:
> # 🔗 Issue
> #823
> 
> # ✍️ Description
> Renames adopter applications table to submissions and updates references to application throughout.
> 
> After I switched to nested modules for the Organizations::AdopterFosterer::CustomForm::SubmissionsController, I realized the comment only mentioned nesting the model. But I assume we will also want to use nested modules for the controllers, is that right? If not, I can change it.
> 
> Also right now, the reviews path/controller is just submissions_reviews but not namespaced under custom form. I can change that too. Let me know.
> 
> # 📷 Screenshots/Demos

All good!

Issue #832, opened 2024-06-12
Add ViewComponents and replace nice_partials implementation


mononoken:
## Description

We have discussed how nice_partials is niche and ViewComponents is fairly regarded and used. Plus there are nice benefits to ViewComponents such as it allowing us to test our view models in isolation.

This issue aims to get our nice_partials implementation replaced with ViewComponents and set up the infrastructure to use ViewComponents for the rest of the app.

## Acceptance Criteria

- [x] Remove nice_partials gem
- [x] Add ViewComponents gem
- [x] Replace nice_partials templates in `views/components/` with components
- [x] Replace shared partials in `views/shared/` with components where applicable

PR #842, opened 2024-06-18
832 view components

mononoken:
# 🔗 Issue
#832 

# ✍️ Description

This PR adds the ViewComponent gem and removes nice_partials. ViewComponent gives us all the utility we were getting from nice_partials, namely being able to yield blocks from an html to a template file, but it gives us much more as well and is more widely used.

One other addition from ViewComponent is that it allows us to test our views. This is especially useful for more complex views such as those with conditionals.

This PR does not apply VC to our whole project. I only aimed to replace the content from `views/components/` and `views/shared`.

Some of the content from those previous directories I thought did not warrant being a partial or a component, so I moved them to their local files. I also refactored a few things along the way too. That said, there are two main types of components in here:

Some like `LikeButtonComponent` I simply copied completely from the partial. On the other spectrum, `CardComponent` is what I _think_ is a more ideal component. It is fairly abstract but provides structure to help unify snippets of reused code throughout our view layer. I was tempted to abstract things like `LikeButtonComponent` into something more abstract like a `ToggleButtonComponent`, but I am not sure how necessary/useful that would be. Instead, I think we can have these more specific components like `LikeButtonComponent` and let abstract ones emerge as more similar button components emerge.

I think a good next step would be to apply the CardComponent throughout the views layer to get all our cards using it and see how much more customization it may require.

# Additional Notes

I wrote this `filter_attribute` method for AvatarComponent, but I could see it potentially being useful for other components or even other parts of the app. I left it in this component for now, but I think it could potentially be nice in a BaseComponent or somewhere shared in the future.
```
# app/components/avatar_component.rb:13
  def filter_attribute(value, allowed_values, default: nil)
    return default unless value
    return value if allowed_values.include?(value)

    default
  end
```

# 📷 Screenshots/Demos

This video gives a brief intro to ViewComponent, specifically for this project.
_Note, I think I said at one point that the avatar is wrapped in a link, but that is actually in a different component. In this case, the template is rendering either an image or initials if no image url is provided._
https://vimeo.com/967458618?share=copy

[ViewComponent docs](https://viewcomponent.org/)


mononoken:
@cflipse it sounded like you were familiar with ViewComponent, so if you have the time, it would be great to get your input. If anyone else also has VC experience, would love feedback. Thank you!

app/views/organizations/adoptable_pets/index.html.erb
```diff
@@ -13,7 +13,39 @@
...
+              <ul class="list-group list-group-flush">
+                <li class="list-group-item d-flex justify-content-between align-items-center flex-wrap">
+                  <%= link_to adoptable_pet_path(pet), class: 'text-decoration-none' do %>
+                    <h3 class="card-title text-secondary mb-0">
+                      <%= pet.name %>
+                    </h3>
+                  <% end %>
+                  <% if current_user&.adopter_foster_account %>
+                    <div class='text-end' id="like_button_pet_<%= pet.id %>">
+                      <%= render LikeButtonComponent.new(pet: pet, like: @likes&.find { |lp| lp.pet_id == pet.id }) %>
```
cflipse, line 36: this `find` is an insidious little n*m query.  You're looping through (on average) half the likes for every single pet.  Instead, when the likes get gathered, can do this:
```
@likes_by_id = @likes.index_by(&:pet_id)
```
and then you'll have a map of <pet_id => like> that can be used as `like: @likes_by_id[pet.id]`  , transforming these repeated scans into a direct lookup.

app/views/organizations/adoptable_pets/show.html.erb
```diff
@@ -60,7 +60,49 @@
...
+                      <h5 class="mb-0 text-body"><%= pet_info[0] %></h5>
+                    </div>
+                  </div>
+                  <div>
+                    <div>
+                      <p class="text-dark mb-0 fw-semibold"><%= pet_info[1] %></p>
+                    </div>
+                  </div>
+                </div>
+              </li>
```
cflipse, line 101: this loop is a pretty good candiate for pulling out into a component as well.  Can have a renders many, and boil down to something like:
```erb
<% render PetInfos.new do |pi| %>
  <%= pi.item "Age", time_ago_in_words(@pet.birth_date).titleize %>
  <%= pi.item "Sex", @pet.sex %>
  <%= pi.item "Breed", @pet.breed %>
  <%= pi.item "Weight", "#{@pet.weight_from} to #{@pet.weight_to} #{@pet.weight_unit}"
<% end %>
```

app/components/image_attachment_table_component.html.erb
```diff
@@ -0,0 +1,48 @@
...
+      <thead>
+        <tr>
+          <th scope="col">Image</th>
+          <th scope="col">Filename</th>
+          <th scope="col"></th>
+        </tr>
+      </thead>
+      <tbody>
+        <% if images.any?  %>
+          <% images.each do |image| %>
```
cflipse, line 14: could be `Array(images).each do |image| `   and save the outer conditional.

app/components/image_attachment_table_component.html.erb
```diff
@@ -0,0 +1,48 @@
...
+                </td>
+              </tr>
+            <% end %>
+          <% end %>
+        <% else %>
+          <tr>
+            <td colspan="3">
+              No images attached.
+            </td>
+          </tr>
```
cflipse, line 43: One thing I've done is to have a general component that wraps collection lists like this.

That component takes a "blank" block, and it's render looks something like:
```ruby
def render
  collection.empty?  ? blank : content
end
```

and then the main block content can handle the rendering of everything as though it's Completely Normal, without running conditionals everywhere.

cflipse (approved):
Overall, looks good to me.  I have to run, so can't finish looking over everything specifically.

Just some general refinements suggested, and non-blocking comments.  :+1:

app/components/image_attachment_table_component.html.erb
mononoken: I like the premise of this abstraction, but I think I am missing the implementation details.

I could do something like this:
```ruby
# app/components/collection_component.rb

class CollectionComponent < ApplicationComponent
  param :collection

  renders_one :blank

  def call
    collection.empty? ? blank : content
  end
end
```

```erb
<%# app/components/image_attachment_table_component.html.erb %>

  <%= render CollectionComponent.new(images) do |c| %>
    <% images.each do |image| %>
      <%# render image content %>
    <% end %>
    <%= c.with_blank do %>
      <%# blank content %>
    <% end %>
  <% end %>
```

But this feels like it's just changing syntax from an if...else block to passing two blocks to an abstract component. It also isn't actually using the collection param in the template. Can you help me clean this up?

app/components/image_attachment_table_component.html.erb
mononoken: As the code currently stands, I think the outer conditional needs to stay since it has an else. I appreciate you pointing out this technique though as I have not seen it before.

app/views/organizations/adoptable_pets/show.html.erb
mononoken: I like that idea, but I think I will save that exercise for a future PR as this one is mostly refactoring existing partials. I will make a draft for this idea on the projects tab.

app/components/image_attachment_table_component.html.erb
mononoken: I am gonna go ahead and get this merged and make a note for this abstraction on the projects tab, so that we can consider refactoring this later.

app/components/avatar_component.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+# Renders a User's avatar as image or user's initials
+class AvatarComponent < ApplicationComponent
```
kasugaijin, line 4: Ah so much better having this logic live in one place.

PR #852, opened 2024-06-26
Bug Fix

kasugaijin:
# 🔗 Issue
N/A

# ✍️ Description
Playing in the dev environment I noted that we were getting a translation error after submitting an application on a pet. Calling send on the status enum value was returning an error because it is not a method. We need to pass the status symbol rather than value to this method for it to work. 

To reproduce, on main branch, log in as an adopter and apply to adopt a pet. You should see: 
image

### DIFF
--- a/app/views/organizations/adoptable_pets/show.html.erb
+++ b/app/views/organizations/adoptable_pets/show.html.erb
@@ -102,8 +102,8 @@
                 <% else %>
                   <% if AdopterApplication.adoption_exists?(current_user.adopter_foster_account&.id, @pet.id) %>
                     <div class='d-flex align-items-center mt-3'>
-                      <h4 class="me-2">
-                        <%= "#{@adoption_application.human_enum_name(@adoption_application.status) || t('.status.default')}" %>
+                      <h4 class="me-2 mb-0">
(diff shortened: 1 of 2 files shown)


app/views/organizations/adoptable_pets/show.html.erb
```diff
@@ -102,8 +102,8 @@
                 <% else %>
                   <% if AdopterApplication.adoption_exists?(current_user.adopter_foster_account&.id, @pet.id) %>
                     <div class='d-flex align-items-center mt-3'>
-                      <h4 class="me-2">
-                        <%= "#{@adoption_application.human_enum_name(@adoption_application.status) || t('.status.default')}" %>
+                      <h4 class="me-2 mb-0">
+                        <%= "Application Status: #{@adoption_application.human_enum_name(:status) || t('.status.default')}" %>
```
jmilljr24, line 106: Looks like you prepended "Application Status:" for clarity. Do we have a translation you could use there as well?

jmilljr24 (approved):
Good fix!

app/views/organizations/adoptable_pets/show.html.erb
kasugaijin: Haha classic I use a string in a view while working on translation fix. I will fix that.

Issue #819, opened 2024-06-07
PageText: Rename this resource to CustomPage


kasugaijin:
The PageText name was outgrown very quickly and not that intuitive. So, let's rename this resource CustomPage. This means renaming all areas in the app (table, model, controller, policies, views, tests etc.). It should be a relatively straightforward find and replace.

Any questions - comment below!

Acceptance Criteria:
- renaming of the PageText resource to CustomPage
- As staff, when looking at the dashboard, the navbar link says Custom Pages instead of Page Text

Jayesh-69:
Hie, let me try this

Trevor-Robinson:
I would like to take this issue.

kasugaijin:
How's this going @Trevor-Robinson? I will unassign unless you are still planning to work on it!

Trevor-Robinson:
Sorry for the delay. Hoping to have this done today.

PR #854, opened 2024-06-28
819 replace page text with custom page

Trevor-Robinson:
# 🔗 Issue
https://github.com/rubyforgood/pet-rescue/issues/819

# ✍️ Description
Changed all occurrences of page text to custom page. Created migration to update table name as well. All tests are passing with new name.


jmilljr24:
@all-contributors
please add @Trevor-Robinson  for code.

jmilljr24:
@Trevor-Robinson Thanks for working on this! Can you take a look that conflicts when you get a chance?

app/controllers/organizations/staff/custom_pages_controller.rb
```diff
@@ -0,0 +1,25 @@
+class Organizations::Staff::CustomPagesController < Organizations::BaseController
```
kasugaijin, line 1: While you are back in this for the conflicts, can you please nest these modules/class instead of inline. We are trying to move to nested slowly but surely. Not a biggie if not, but sweet if you can.

kasugaijin:
hey @Trevor-Robinson all you have to run now is `rails standard:fix` to get the linter to pass :)

Trevor-Robinson:
Should be all good now.

kasugaijin (approved):
Thank you! @Trevor-Robinson

kasugaijin:
@Trevor-Robinson Just a heads up - I missed it in the review, but there were some issues with the naming you used on migrations. First, I am not sure why you changed the class names of the two existing migrations from PageText to CustomPage. You should never change an existing migration (the only exception might be if it's the latest one). But this can cause a lot of unintended consequences. You are better of making changes through new migrations.

Also, the naming of the latest migration was wrong - i.e. the filename should always match the class name so Rails knows where to find the class when you call it. See below.

These are not major issues and easy to fix, but I wanted to provide feedback. Also, I am not sure how the schema got update with these errors...the migrations should have failed. So, did you update the file manually by chance? That's also a red flag if you had to do that.

image

Issue #851, opened 2024-06-26
Create FormSubmission model


kasugaijin:
This is a model that binds a set of submitted answers together from a form with a Person. Create the model and form_submissions table with person_id and organization_id foreign keys.

Model associations:
- Belongs_to Person
- Belongs_to Organization
- Has_many adopter_applications
- Has_many submitted_answers

Please add model tests.

Gabe-Torres:
I'll take this one!

Gabe-Torres:
Hey @kasugaijin!

I ran into a couple of issues when creating the `form_submission `model. I know there are some incoming big changes to the codebase, so maybe these will be resolved within the next couple of weeks.

I successfully created the `form_submission` model and test. In the process of updating the `adopter_applications` model test to account for the changes, I made an associations context block. I also had to create a quick factory so `form_submission` can be referenced within the test. After a couple of tweaks, the `form_submission`, `adopter_applications`, and `submitted_answers` model tests are all passing. However, the database will no longer seed properly. My guess is that the seeds need to be updated accordingly to account for the new changes. There are also quite a few errors after running the test suites.

I wanted to make sure I am heading in the right direction with this issue/PR.

Should I update the seeds and errors where needed? Was there no associations block within `adopter_applicaitons` yet for this reason?
Is creating a factory for `form_submission` and implementing it within `adopter_applications` something you wanted done in this PR? Or should I just create the new `form_submission` model and leave the other changes for a different issue ticket?

Thank you!

kasugaijin:
@Gabe-Torres i should have time to get a proper look in tomorrow! But yes it’s pretty common to need to update the seed data if an association has changed that the seeds rely upon.

Gabe-Torres:
Ah okay that does make sense! Thank you. Apologies for the 20 questions too 😂

PR #856, opened 2024-06-28
851 create form submission model

Gabe-Torres:
# 🔗 Issue
(https://github.com/rubyforgood/pet-rescue/issues/851)

# ✍️ Description
Create form_submission model 
- create test for form_submission model
- create the foreign key for has many relationship between form_submission and submitted_answer
- create the foreign key for has many relationship between form_submission and adopter_application
- update submitted_answer model test to include new association with form_submission
- update adopter_applicaiton model test to include new association with form_submission
   - Added an associations context block
   - this also involved creating a form_submission factory that can be called within tests

# 📷 Screenshots/Demos


app/models/form_submission.rb
```diff
@@ -0,0 +1,27 @@
...
+#
+#  fk_rails_...  (organization_id => organizations.id)
+#  fk_rails_...  (person_id => people.id)
+#
+class FormSubmission < ApplicationRecord
+belongs_to :person
+belongs_to :organization
+
+has_many :adopter_applications
+has_many :submitted_answers, class_name: "CustomForm::SubmittedAnswer"
```
kasugaijin, line 26: Does it need the module CustomForm specified here?

kasugaijin:
@Gabe-Torres This is looking good so far! Like you said, the pipeline is not passing because the seeds are not loading due to an error - my immediate guess is that AdopterApplication and SubmittedAnswer do not have a FormSubmission to belong to. 

To your questions about adding in the additional components to get the tests to pass - ideally, yes. PRs should be complete and always pass the pipeline. This will likely mean updating the seeds to make sure associated objects exist, and to update any test setups that also need the object to exist. Now, if this is starting to look like a huge chunk of work, we can always help you out! Let me know what you think! Ask as many questions as you need.

test/models/adopter_application_test.rb
```diff
@@ -2,17 +2,24 @@
 
 class AdopterApplicationTest < ActiveSupport::TestCase
   setup do
-    @application = create(:adopter_application)
+    @form_submission = create(:form_submission)
+    @application = create(:adopter_application, form_submission: @form_submission)
+  end
+
+  context "associations" do
```
kasugaijin, line 9: Ah I guess this was missing before! Thanks for adding.

Gabe-Torres:
> context "associations" do

> @Gabe-Torres This is looking good so far! Like you said, the pipeline is not passing because the seeds are not loading due to an error - my immediate guess is that AdopterApplication and SubmittedAnswer do not have a FormSubmission to belong to. So it will be a matter of creating that object and passing it to its associations in both Baja and Alta seeds files. I usually like to just focus on one seed file first and get that working, then it's more or less a copy paste into the other file as they should be almost identical.
> 
> To your questions about adding in the additional components to get the tests to pass - ideally, yes. PRs should be complete and always pass the pipeline. This will likely mean updating the seeds to make sure associated objects exist, and to update any test setups that also need the object to exist. Now, if this is starting to look like a huge chunk of work, we can always help you out! Let me know what you think! Ask as many questions as you need.
> 
> I usually try to break things up more into smaller chunks. But, given this is addition of a model, we need to add it and make sure it's working as expected. So, it's a bit more work...but should not be too much more as it's just two associated models. Famous last words.

Thank you for the feedback @kasugaijin ! I can get sidetracked sometimes, so it's good to know I was on the right track. I updated tests where needed and made some tweaks to the `adopter_applications` factory to account for the new association with `form_submissions`.

Regarding seeds – How are we planning to handle a `form_submission` instance? In my current version, I created one `form_submission` object that has an association with the `organization` at the top of the file, along with a singular `person`. The seeds create 10 `adopter applications`. Do you think it would make sense to also create 10 `form_submissions`, or would that suffice for our needs?

kasugaijin:
> > context "associations" do
> 
> > @Gabe-Torres This is looking good so far! Like you said, the pipeline is not passing because the seeds are not loading due to an error - my immediate guess is that AdopterApplication and SubmittedAnswer do not have a FormSubmission to belong to. So it will be a matter of creating that object and passing it to its associations in both Baja and Alta seeds files. I usually like to just focus on one seed file first and get that working, then it's more or less a copy paste into the other file as they should be almost identical.
> > To your questions about adding in the additional components to get the tests to pass - ideally, yes. PRs should be complete and always pass the pipeline. This will likely mean updating the seeds to make sure associated objects exist, and to update any test setups that also need the object to exist. Now, if this is starting to look like a huge chunk of work, we can always help you out! Let me know what you think! Ask as many questions as you need.
> > I usually try to break things up more into smaller chunks. But, given this is addition of a model, we need to add it and make sure it's working as expected. So, it's a bit more work...but should not be too much more as it's just two associated models. Famous last words.
> 
> Thank you for the feedback @kasugaijin ! I can get sidetracked sometimes, so it's good to know I was on the right track. I updated tests where needed and made some tweaks to the `adopter_applications` factory to account for the new association with `form_submissions`.
> 
> Regarding seeds – How are we planning to handle a `form_submission` instance? In my current version, I created one `form_submission` object that has an association with the `organization` at the top of the file, along with a singular `person`. The seeds create 10 `adopter applications`. Do you think it would make sense to also create 10 `form_submissions`, or would that suffice for our needs?

So the benefit of the FormSubmission is that we should be able to associated multiple adopter applications with a single FormSubmission, so let's go with that in the seeds for now.

app/models/form_submission.rb
```diff
@@ -0,0 +1,27 @@
...
+#  index_form_submissions_on_person_id        (person_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (organization_id => organizations.id)
+#  fk_rails_...  (person_id => people.id)
+#
+class FormSubmission < ApplicationRecord
+belongs_to :person
+belongs_to :organization
```
kasugaijin, line 23: Check out other resources that belong to an organization (most of them) and you'll see we use the acts as tenant helper here. This is important for setting the tenancy, or scope, for a given record i.e., which org it belongs to, and is used by the acts as tenant scoping logic elsewhere.

app/models/form_submission.rb
```diff
@@ -0,0 +1,27 @@
...
+#
+#  index_form_submissions_on_organization_id  (organization_id)
+#  index_form_submissions_on_person_id        (person_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (organization_id => organizations.id)
+#  fk_rails_...  (person_id => people.id)
+#
+class FormSubmission < ApplicationRecord
```
kasugaijin, line 21: I am also a fan of documenting in the app where we can. Can we add a line or two of comments above the class definition to explain what this class does? I think it's quite an important piece of the architecture managing adopter data. 

Something like...
```
The FormSubmission model binds together a set of submitted answers from an adopter into a single submission. This represents a snapshot in time. If an adopter submits answers either through an in-app form or a third party form that is uploaded to the app, we will have a set of answers belonging to a form submission. This can be valid for a period of time to allow the user to apply to adopt multiple pets. If this time expires, or the person's contet changes and they need to update the form, we now have a new set of answers and a new form submission.
```
```

kasugaijin:
@Gabe-Torres Looking great! Just a few comments.

Gabe-Torres:
Awesome! I changed the belongs_to association to acts_as_tenant.  Thank you @kasugaijin. I think everything is updated and ready to go!

app/models/form_submission.rb
kasugaijin: @Gabe-Torres can you add the above comments above the class definition?

kasugaijin:
@Gabe-Torres approved, but please add the comments to the class per my comment above.

kasugaijin (approved):

Gabe-Torres:
Added! Apologies, I had added the comments before merging the most recent main and must of not added them while sorting conflicts.

kasugaijin (approved):

Issue #848, opened 2024-06-26
Merge OrganizationProfile into Organization


kasugaijin:
Merge `OrganizationProfile` into `Organization`. This will require a refactor of models, controllers, views, policies, tests and tables to update the names and method names used. 

We also want to delete the Organization Profiles table, and add the attributes to the Organizations table (donation_url, email,         facebook_url, instagram_url, phone_number) and the applicable model logic for each.

In doing so, we also need to change the association with the Location model; an Organization should have_many locations, and a Location should belong_to an Organization (foreign key).

Ask any questions below!

sarvaiyanidhi:
Hi @kasugaijin I can take up this issue. Can you please assign it to me? Thanks

kasugaijin:
@sarvaiyanidhi all yours

PR #870, opened 2024-07-10
848 - Merge OrganizationProfile with Organization

sarvaiyanidhi:
# 🔗 Issue
#848 

# ✍️ Description
- Migrated OrganizationProfile fields into Organization table
- Updated Model, Controller, Policy, Routes with relevant changes
- Updated test cases
- Removed OrganizationProfile reference across the application

# 📷 Screenshots/Demos
NA


kasugaijin:
@sarvaiyanidhi thank you! I am away this weekend so unless someone beats me to it, I won't review this until next week.

sarvaiyanidhi:
> @sarvaiyanidhi thank you! I am away this weekend so unless someone beats me to it, I won't review this until next week.

Sure @kasugaijin No worries. Will await for review from you or someone else. Thanks

db/migrate/20240709185643_merge_organization_profiles_to_organizations.rb
```diff
@@ -0,0 +1,36 @@
+class MergeOrganizationProfilesToOrganizations < ActiveRecord::Migration[7.1]
+  def up
```
mononoken, line 2: I think this migration can be shortened using the `change` method. Something like this:
```ruby
  def change
    change_table :organizations do |t|
      t.string :email
      t.string :phone_number
      t.text :donation_url
      t.text :facebook_url
      t.text :instagram_url
      t.references :location, foreign_key: true
    end

    execute <<-SQL
      UPDATE organizations o
      SET 
        location_id = op.location_id,
        email = op.email,
        phone_number = op.phone_number,
        facebook_url = op.facebook_url,
        instagram_url = op.instagram_url,
        donation_url = op.donation_url
      FROM organization_profiles op
      WHERE op.organization_id = o.id
    SQL

    change_column_null :organizations, :location_id, false
  end
```
Are there any advantages to the `up/down` that I should be considering?

db/migrate/20240709185643_merge_organization_profiles_to_organizations.rb
```diff
@@ -0,0 +1,36 @@
...
+    add_column :organizations, :email, :string
+    add_column :organizations, :phone_number, :string
+    add_column :organizations, :donation_url, :text
+    add_column :organizations, :facebook_url, :text
+    add_column :organizations, :instagram_url, :text
+    add_reference :organizations, :location, foreign_key: true
+
+    # Migrate existing data
```
mononoken, line 10: I don't think we need these comments. You've written this code self-explanatory 🙂

db/migrate/20240709185643_merge_organization_profiles_to_organizations.rb
```diff
@@ -0,0 +1,36 @@
...
+        email = op.email,
+        phone_number = op.phone_number,
+        facebook_url = op.facebook_url,
+        instagram_url = op.instagram_url,
+        donation_url = op.donation_url
+      FROM organization_profiles op
+      WHERE op.organization_id = o.id
+    SQL
+
+    # Add NOT NULL constraint
```
mononoken, line 24: I don't think we need these comments. You've written this code self-explanatory 🙂

mononoken (changes requested):
Thank you for the thorough work @sarvaiyanidhi ! I appreciate the attention to detail on these changes. I have a few comments that need to be addressed and then this should be good to merge. If you need clarification or help feel free to ping me on slack.

sarvaiyanidhi:
Thank you @mononoken for the feedback. I have gone through all comments and will make requested changes. Will push the revised code soon.

db/migrate/20240709185643_merge_organization_profiles_to_organizations.rb
sarvaiyanidhi: I think as we are using raw SQL in this migration so it is recommended to use up/down to avoid reversible issue.

If we use change and try to rollback this migration we encounter below issues

This migration uses execute, which is not automatically reversible.
To make the migration reversible you can either:
1. Define #up and #down methods in place of the #change method.
2. Use the #reversible method to define reversible behavior.

app/models/location.rb
```diff
@@ -12,7 +12,7 @@
 #
 class Location < ApplicationRecord
   has_one :adopter_foster_profile
-  has_one :organization_profile
+  has_one :organization
```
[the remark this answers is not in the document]
sarvaiyanidhi: Got it. Sorry, somehow missed the point in the issue description. Will do the required changes.

mononoken:
Hi, @sarvaiyanidhi ! I saw you made some updates, which skimming over, look good. However, some tests are not passing. Are you still working on this? Just comment here whenever it's ready for review or click to re-request review. Let us know if you need help with any parts 😊

sarvaiyanidhi:
> Hi, @sarvaiyanidhi ! I saw you made some updates, which skimming over, look good. However, some tests are not passing. Are you still working on this? Just comment here whenever it's ready for review or click to re-request review. Let us know if you need help with any parts 😊

Hi @mononoken, I have implemented all changes related to location association changes and also made necessary changes as suggested for Avatar deletion to work with organization.

I am facing issue with test for attachment_policy.rb and attachment_controller.rb files after making those changes and trying to fix it. I did try to make related changes in test files to change from profile to organization but still not able to fix it. If you can provide pointers on 4 failing test cases related to it it will help me to proceed further.

mononoken:
@sarvaiyanidhi The problem from the tests seems to stem from how `AttachmentPolicy#organization` is working. The policy is trying to verify the organization associated with the attachment matches the user's org.

I had suggested you update the `#organization` method because of how `Organization` has been refactored. Since OrganizationProfile is gone, Organization itself now has attachments. However, that suggestion was incomplete because orgs are not the only things with attachments. Pets also have attachments for example. So the fix I suggested fixed Org attachments but broke all the other ones. 

```ruby
  # app/policies/active_storage/attachment_policy.rb
  def organization
    @organization || record_organization
  end

  def record_organization
    if record.record.is_a?(Organization)
      record.record
    else
      record.record.organization
    end
  end

```

This code adapts the policy to both cases. If the attachment's record is itself an Org, it uses the record. If it is not, it checks that the attachment's record has an associated org and uses that.

If you could take a few minutes to double check that I am correct and see if you can come up with a better approach that would be great! I don't like using `is_a?` but I think it's probably fine and the simplest solution I can think of in this case.

sarvaiyanidhi:
Thank you @mononoken .. It does solves the problem.

How about this alternate code suggestion?

```
  def organization
    @organization || record_organization
  end

  def record_organization
    (record.record.respond_to?(:organization) ? record.record.organization : record.record)
  end
```

Do you recommend this or the one you suggested?

mononoken:
I prefer my suggested approach. The only time we want `organization === record.record` is if it is an `Organization`, so having `record.record` in the else of that ternary feels a little wrong to me. I think the `is_a?` makes the intentions clearer too. However, both will work for this case, so I will leave it to you to decide which you prefer to push 🙂

You could also change my if to a ternary if you choose that path. However, if you use a ternary, I would lose the outer parentheses. They're not necessary for this statement.

mononoken (approved):
Looks good! Thank you for picking up those comments.

Just to let you know, one bug has surfaced from these changes. The locations are associated with both AdopterFosterProfiles and Orgs, but this PR only associates the locations with the Organizations. It also forces this association using the `acts_as_tenant` method.

I will be writing a new issue to address this concern. If you want to work on that one next, I would be happy to assign it to you. No pressure though if you want to work on something else. I'll try to get that one typed up today. 

Thanks again for your work on this one!

sarvaiyanidhi:
> Just to let you know, one bug has surfaced from these changes. The locations are associated with both AdopterFosterProfiles and Orgs, but this PR only associates the locations with the Organizations. It also forces this association using the `acts_as_tenant` method.
 
> I will be writing a new issue to address this concern. If you want to work on that one next, I would be happy to assign it to you. No pressure though if you want to work on something else. I'll try to get that one typed up today.

Hi @mononoken ,

Sure, I can work on this related issue and get it done. Can you please assign new issue to me once you add it. Thanks!

sarvaiyanidhi:
Hey @mononoken ,

As per earlier comment, I think we don't need new issue now for AdopterFosterProfiles and location association issue as I saw recently issue on removing AdopterFosterProfiles. Let me know if you are still looking for any other changes related to it, I can work on it.

Thanks

mononoken:
> Hey @mononoken ,
> 
> As per earlier comment, I think we don't need new issue now for AdopterFosterProfiles and location association issue as I saw recently issue on removing AdopterFosterProfiles. Let me know if you are still looking for any other changes related to it, I can work on it.
> 
> Thanks

Hi, @sarvaiyanidhi ! So sorry I did not get back to you on this. I've been MIA for a few weeks. You are right in your observation that we are removing AdopterFosterProfile now, so we won't need to make that change after all. Sorry again for any confusion this may have caused you!

Issue #849, opened 2024-06-26
Rename roles


kasugaijin:
Rename the roles we use for authorization. This will require updating of policies and wherever else we use roles in the logic.

`admin` should become `superadmin`
`staff` should become `admin`

Aaryanpal:
hi @kasugaijin, I will take this issue.

PR #871, opened 2024-07-14
849 rename admin roles

mononoken:
# 🔗 Issue

#849 

# ✍️ Description

This PR renames the roles:
- :admin to :super_admin
- :staff to :admin

I also included a migration to move data over to these new names. Figured it would be good to get in this habit before prod.

There were a few other things I cleaned up:
- Permissions constants were not defined inside the authorizable module... woops! I considered making them private too, but they feel like part of the public interface to me, so I left them as such.
- Fixed some link conditionals to stop using account models for authz logic.

# 📷 Screenshots/Demos
image


kasugaijin (approved):
LGTM thanks @mononoken !

Issue #869, opened 2024-07-10
CSV upload: write a CSV import service


kasugaijin:
We need to be able to import a CSV of third party form data containing adopter responses to questions. We will need to be able to match a Person in the app with a row in the CSV using an email address, and then save a new FormAnswer record for each question and answer - we will save the question and the answer on the record. All of the FormAnswers will belong to a single FormSubmission record, which acts as a binder for all of the FormAnswers, representing a submission of data at a given time. 

Note that we need to be able to determine if a given row has already been imported. If it has, we skip it. If it has not, we import it if there is a matching Person. To be able to determine if we have already imported, we need to investigate a little into what data is provided in form CSV outputs by common services like Google Forms. We might be able to use a date/timestamp and save that on the FormSubmission...if a person has a FormSubmission with a given timestamp, we skip that row. If not, we treat it as a new submission and import it. 

Ask away with questions! Note that the FormAnswer model still needs to be renamed (issue https://github.com/rubyforgood/pet-rescue/issues/867)

jmilljr24:
I took a peak at this and a couple of questions came to mind. Looking at a sample Google Form csv, I think checking the db for a person and datetime (timestamp from row) will be the easiest solution. 

1. Will FormAnswer belong to User or Person? FormSubmission is indexed on Person but I don't see that change listed for FormAnswer in #867 .
2. Person query based on email could have issues based off of how the email was entered/saved in the csv (case sensitivity). I know it can be dealt with in the query but I'm not sure if that's the best solution. Here is a [quick read](https://www.mikecoutermarsh.com/storing-email-in-postgres-rails-use-citext/) on some options.

kasugaijin:
@jmilljr24 yes FormAnswer will belong to Person. I think that issue to update FormAnswer is currently in the hopper and was blocked by the creation of form submission because it also belongs to that. That should now be unblocked so I will make the issue. This Csv service will be blocked until that is done…but certainly there’s a lot that can be done before then too.

I didn’t think about email case in the db so I appreciate you bringing that up. We should select an appropriate solution on our call!

PR #872, opened 2024-07-16
869 CSV Import service

jmilljr24:
#869

# ✍️ Description

Import Service for CSV files uploaded by organizations. Currently implemented base off Google Form's CSV files.

# 📷 Screenshots/Demos


app/services/organizations/csv_import_service.rb
```diff
@@ -0,0 +1,37 @@
...
+        previous = FormSubmission.where(person: person, csv_timestamp: timestamp)
+        next unless previous.empty?
+
+        ActiveRecord::Base.transaction do
+          form_submission = FormSubmission.create(person: person, csv_timestamp: timestamp)
+          row.each do |col|
+            # skip Email and Timestamp col as they are saved on Form Submission
+            next if col[0] == "Email" || col[0] == "Timestamp"
+
+            FormAnswer.create(person: person, form_submission: form_submission,
```
kasugaijin, line 30: do we need to save the person on the FormAnswer? FormAnswer belongs_to FormSubmission, which belongs_to Person, and we can set FormAnswer belongs_to Person through FormSubmission. I think i most contexts we will be calling FormAnswer from FormSubmission rather than Person, because we will want to always be considering the FormAnswer context i.e. only give me those answers from the same submission.

kasugaijin:
@jmilljr24 I think this is a good start!

app/services/organizations/csv_import_service.rb
jmilljr24: Agreed. The old submitted answer had a user column so I did it as a placeholder without thinking through the associations. I'll adjust when the FormAnswer PR is complete so I get everything correct.

kasugaijin: Ah yes. I’ve also asked in the PR to name FormAnswer to remove user from that model association

jmilljr24:
@kasugaijin This is updated to work with FormAnswer. 

Waiting on the email format that will be used in the DB and if we decided on a timestamp column.

test/services/organizations/csv_import_service_test.rb
```diff
@@ -0,0 +1,74 @@
+require "test_helper"
+require "csv"
+
+class Organizations::CsvImportServiceTest < ActiveSupport::TestCase
```
kasugaijin, line 4: can we nest modules and classes from now on please :)

kasugaijin:
Looking good @jmilljr24 
For the email, we are only collecting it on one model, User, and currently only on one form. What I suggest we do is have a before save hook. It will serve MVP fine and will do for a long time. I don't see us adding any new models that collect email any time soon.

```
class User < ApplicationRecord
  before_save :downcase_email

  private

  def downcase_email
    self.email = email.downcase if email.present?
  end
end
```

With respect to the CSV timestamp, I like your proposed approach to store this on the FormSubmission as `csv_timestamp`. 

Did you want to include both of those changes in this PR?

jmilljr24:
> Looking good @jmilljr24 For the email, we are only collecting it on one model, User, and currently only on one form - adopter sign up. What I suggest we do is have a before save hook. It will serve MVP fine and will do for a long time. I don't see us adding any new models that collect email any time soon.
> 
> ```
> class User < ApplicationRecord
>   before_save :downcase_email
> 
>   private
> 
>   def downcase_email
>     self.email = email.downcase if email.present?
>   end
> end
> ```
> 
> With respect to the CSV timestamp, I like your proposed approach to store this on the FormSubmission as `csv_timestamp`.
> 
> Did you want to include both of those changes in this PR?

Yes. I can include the changes. Thanks!

kasugaijin (approved):
LGTM thanks @jmilljr24 
I was thinking...once we start actually using this in the wild, if there are nuances per third party service, we might want to write a parser specific to each service to handle its nuance. We could then include a dropdown select next to the form upload that specifies where the data came from. Then we can be sure to handle things neatly and be able to clearly support X services. It will avoid issues where a user uploads something from some obscure service, or worse, their own CSV, and wonders why it doesn't work as expected. Thoughts? Not something we have to do now, but I bet we'll run into this sooner than later.

jmilljr24:
> LGTM thanks @jmilljr24 I was thinking...once we start actually using this in the wild, if there are nuances per third party service, we might want to write a parser specific to each service to handle its nuance. We could then include a dropdown select next to the form upload that specifies where the data came from. Then we can be sure to handle things neatly and be able to clearly support X services. It will avoid issues where a user uploads something from some obscure service, or worse, their own CSV, and wonders why it doesn't work as expected. Thoughts? Not something we have to do now, but I bet we'll run into this sooner than later.

Yes I had a similar thought about the dropdown. For the time being, we should definitely make a note on the upload page that lets the admin know this is only setup up for google forms right now. I also left the comment regarding the headers being specific to Google Forms for the next dev.

If there is any way we could test out one of the stakeholders csv files before MVP I would feel a bit better about this PR. There can be a lot a edge cases with csv's but I didn't want to add complexity without an actual need. I'm just not sure what we will come across with their files.

kasugaijin:
Yes I agree it would be good to add a note re google forms to the upload form and also do some testing. I think once we are at MVP we can test with real data. I just hesitate to ask for all that private data right now. I think the next best thing we can do is make our own google form and do some testing. We can make an issue for this once the form input route, controller and view are in place.

Issue #867, opened 2024-07-10
Refactor CustomForm::SubmittedAnswer to FormAnswer


kasugaijin:
** You can start this but one of the associations cannot be made until this PR is merged (which is close) https://github.com/rubyforgood/pet-rescue/pull/856/files**

After some small architectural decisions requiring changes, we need the SubmittedAnswer model to be removed from the CustomForm scope and renamed to FormAnswer. We also need the submitted_answers table to be named form_answers. Also, please update all references to the class name and methods throughout the code. Basically a find and replace exercise for part of this work.

We should also update the associations on the model:
- has_one :organization, through: :submission
- Get rid of  belongs_to :question
- Get rid of  has_one :form, through: :question
- Belongs_to FormSubmission (blocked until above PR merged)

Gabe-Torres:
Can I take this one?

kasugaijin:
@Gabe-Torres all yours!

kasugaijin:
@Gabe-Torres quick update - I think we should stick to using `acts_as_tenant(:organization)` here (like we do on all other records) instead of making a belongs_to org through submission.

PR #873, opened 2024-07-16
867 refactor customform  submittedanswer to formanswer

Gabe-Torres:
# 🔗 Issue

https://github.com/rubyforgood/pet-rescue/issues/867

# ✍️ Description

- Removes `SubmittedAnswer` from the `CustomForm` scope and renames it to `FormAnswer`

- Removes `belongs_to :question` and`has_one :form, through: :question` from `FormAnswer`

- Renames `submitted_answer` table to `form_answer ` table in apps database

- Updates all references to the new class name

- Adds `acts_as_tenant(:organization)` to `FormAnswer` with Foreign Key association to `Organization`

# 📷 Screenshots/Demos

https://jam.dev/c/e590b25e-5bc0-4971-ae7e-a60c5686c7ba


app/models/form_answer.rb
```diff
@@ -1,35 +1,33 @@
...
-    belongs_to :question
-    belongs_to :user
-    belongs_to :form_submission
-
-    has_one :form, class_name: "CustomForm::Form", through: :question
-    has_one :organization, through: :form
-  end
+class FormAnswer < ApplicationRecord
+  acts_as_tenant(:organization)
+  belongs_to :user
```
kasugaijin, line 31: We are reserving the User model for authentication and authorization, so can we make this belong to Person instead?

db/migrate/20240716194544_add_organization_to_form_answers.rb
```diff
@@ -0,0 +1,5 @@
+class AddOrganizationToFormAnswers < ActiveRecord::Migration[7.1]
+  def change
+    add_reference :form_answers, :organization, null: false, foreign_key: true
```
kasugaijin, line 3: Shouldn't we have a reference to FormSubmission here? Given the belongs_to relationship

test/models/form_answer_test.rb
```diff
@@ -0,0 +1,8 @@
+require "test_helper"
+
+class FormAnswerTest < ActiveSupport::TestCase
+  context "associations" do
+    should belong_to(:user)
```
kasugaijin, line 5: Per above, please update.

app/models/form_answer.rb
Gabe-Torres: Makes sense! I'll update

db/migrate/20240716194544_add_organization_to_form_answers.rb
Gabe-Torres: I believe there is a reference to `FormSubmission` already and this migration file only adds the `Organization` reference since we added the `acts_as_tenant(:organization)`

```ruby
create_table "form_answers", force: :cascade do |t|
    t.json "value", null: false
    t.json "question_snapshot", null: false
    t.bigint "question_id", null: false
    t.bigint "user_id", null: false
    t.datetime "created_at", null: false
    t.datetime "updated_at", null: false
    t.bigint "form_submission_id", null: false
    t.bigint "organization_id", null: false
    t.index ["form_submission_id"], name: "index_form_answers_on_form_submission_id"
    t.index ["organization_id"], name: "index_form_answers_on_organization_id"
    t.index ["question_id"], name: "index_form_answers_on_question_id"
    t.index ["user_id"], name: "index_form_answers_on_user_id"
    ```

Comment (Gabe-Torres) [reply db/migrate/20240716194544_add_organization_to_form_answers.rb:3]: I also just noticed the question_id and question_snapshot reference. It would make sense to remove those two, right?

File: test/models/form_answer_test.rb (hunk 3 above)
Comment (Gabe-Torres) [reply test/models/form_answer_test.rb:5]: Will do!
We still want to keep the `user_id` reference within the `form_answers` table though yeah?

File: db/migrate/20240716194544_add_organization_to_form_answers.rb (hunk 2 above)
Comment (kasugaijin) [reply db/migrate/20240716194544_add_organization_to_form_answers.rb:3]: We are actually keeping question snapshot because we want to store the question with every answer so we have a record of what the question was. Let’s keep question Id for now but remove the null false constraint so a FormAnswer can have a parent question, but doesn’t have to.

Comment (kasugaijin) [reply db/migrate/20240716194544_add_organization_to_form_answers.rb:3]: We can fully remove user id and it’s index

File: test/models/form_answer_test.rb (hunk 3 above)
Comment (kasugaijin) [reply test/models/form_answer_test.rb:5]: We now have a person model for things like this, keeping User just for authentication and authorization. We can fully remove all refs to User on FormAnswer because FormAnswer belongs to FormSubmission which belongs to Person. So we can add a has through association on Person and FormAnswer (through FormSubmission) but remove the table column and index for user as it’s not required. Does that make sense?

Comment (kasugaijin) [reply test/models/form_answer_test.rb:5]: If you need more clarification let me know and I can make a look. Typing on my phone sucks haha

Comment (Gabe-Torres) [reply test/models/form_answer_test.rb:5]: Makes sense! Thank you 💯

Comment (Gabe-Torres) [reply test/models/form_answer_test.rb]: I added the desired changes in my most recent push! 
One issue though - I could not add a  `belongs_to :person, through :form_submission`. Turns out the `.through` option is only available with `has_many` or `has_one` associations. It doesn't work with `belongs_to` associations. Something I didn't know! 

I added a Foreign Key reference to `Person` instead to satisfy the requirement for the `belongs_to :person` association in `FormAnswer`. Is there another way we should go about this?

**Comment by Gabe-Torres** [comment]
I added the desired changes in my most recent push!
One issue though - I could not add a belongs_to :person, through :form_submission. Turns out the .through option is only available with has_many or has_one associations. It doesn't work with belongs_to associations. Something I didn't know!

I added a Foreign Key reference to Person instead to satisfy the requirement for the belongs_to :person association in FormAnswer. Is there another way we should go about this?

**Comment by kasugaijin** [comment]
> I added the desired changes in my most recent push! One issue though - I could not add a belongs_to :person, through :form_submission. Turns out the .through option is only available with has_many or has_one associations. It doesn't work with belongs_to associations. Something I didn't know!
> 
> I added a Foreign Key reference to Person instead to satisfy the requirement for the belongs_to :person association in FormAnswer. Is there another way we should go about this?

Ah right yes that’s right. I have not checked the code as I am on my cell but did you add has many FormAnswers through FormSubmission on Person? We can still do that. Also we don’t need any reference to Person on the FormAnswer model.

File: app/models/organization.rb [hunk 4]
```diff
@@ -22,6 +22,7 @@ class Organization < ApplicationRecord
   has_many :default_pet_tasks
   has_many :forms, class_name: "CustomForm::Form", dependent: :destroy
   has_many :faqs
```
kasugaijin, line 24: Shouldn’t this be plural?

Gabe-Torres: Ah yes! Good catch. Thank you

Gabe-Torres:
> > I added the desired changes in my most recent push! One issue though - I could not add a belongs_to :person, through :form_submission. Turns out the .through option is only available with has_many or has_one associations. It doesn't work with belongs_to associations. Something I didn't know!
> > I added a Foreign Key reference to Person instead to satisfy the requirement for the belongs_to :person association in FormAnswer. Is there another way we should go about this?
> 
> Ah right yes that’s right. We can still use has many FormAnswers through FormSubmission on the Person model.
> 
> Also we don’t need any reference to Person on the FormAnswer model. We can get rid of the person ID column and the associations and the index for this.
> 
> The reason we don’t need FormAnswer belongs to Person is because we already have the through relationship for Person and FormAnswer so we can call person.form_answers and we will always be using form answers through the context of a FormSubmission because that is the bucket that groups a related set of answers together. I don’t see the need to call form_answer.person at the moment and if we don’t need it we shouldn’t add it :)

Gotcha! That makes sense and the vision is becoming more clear in my head. I just tested this out in rails console and it worked. So I can get rid of the `belongs_to :person` association within the `FormAnswer` model? So the only associations will be 
`acts_as_tenant(:organization)` & `belongs_to :form_submission` ?

kasugaijin:
> > > I added the desired changes in my most recent push! One issue though - I could not add a belongs_to :person, through :form_submission. Turns out the .through option is only available with has_many or has_one associations. It doesn't work with belongs_to associations. Something I didn't know!
> > > I added a Foreign Key reference to Person instead to satisfy the requirement for the belongs_to :person association in FormAnswer. Is there another way we should go about this?
> > 
> > 
> > Ah right yes that’s right. We can still use has many FormAnswers through FormSubmission on the Person model.
> > Also we don’t need any reference to Person on the FormAnswer model. We can get rid of the person ID column and the associations and the index for this.
> > The reason we don’t need FormAnswer belongs to Person is because we already have the through relationship for Person and FormAnswer so we can call person.form_answers and we will always be using form answers through the context of a FormSubmission because that is the bucket that groups a related set of answers together. I don’t see the need to call form_answer.person at the moment and if we don’t need it we shouldn’t add it :)
> 
> Gotcha! That makes sense and the vision is becoming more clear in my head. I just tested this out in rails console and it worked. So I can get rid of the `belongs_to :person` association within the `FormAnswer` model? So the only associations will be `acts_as_tenant(:organization)` & `belongs_to :form_submission` ?

That’s right yes!

Gabe-Torres:
Awesome! I applied those changes 👍🏽  
Should hopefully be ready to go

db/migrate/20240718195012_add_person_to_form_answers.rb
```diff
@@ -0,0 +1,5 @@
+class AddPersonToFormAnswers < ActiveRecord::Migration[7.1]
+  def change
+    add_reference :form_answers, :person, null: false, foreign_key: true
```
kasugaijin, line 3: Why did you add a reference to person? If we look at the associations for FormAnswer it's just Organization and FormSubmission.

db/migrate/20240719030053_remove_person_from_form_answers.rb
```diff
@@ -0,0 +1,6 @@
+class RemovePersonFromFormAnswers < ActiveRecord::Migration[7.1]
+  def change
+    remove_index :form_answers, :person_id
+    remove_column :form_answers, :person_id
+  end
+end
```
kasugaijin, line 6: So, if you remove `AddPersonToFormAnswers` migration you don't need this migration, right?

db/schema.rb
```diff
@@ -142,6 +142,18 @@
     t.index ["organization_id"], name: "index_faqs_on_organization_id"
   end
 
+  create_table "form_answers", force: :cascade do |t|
```
kasugaijin, line 145: This end result looks good. I just think we can clean up the migrations a bit.

db/schema.rb
```diff
@@ -373,6 +372,9 @@
   add_foreign_key "custom_pages", "organizations"
   add_foreign_key "default_pet_tasks", "organizations"
   add_foreign_key "faqs", "organizations"
+  add_foreign_key "form_answers", "form_submissions"
+  add_foreign_key "form_answers", "organizations"
+  add_foreign_key "form_answers", "questions"
```
kasugaijin, line 377: We don't need this foreign key for Question on FormAnswer

kasugaijin:
Getting close @Gabe-Torres just a couple of comments re migrations.

db/migrate/20240719030053_remove_person_from_form_answers.rb
Gabe-Torres: Right! I'm just hesitant to alter existing migration files manually. But in this case, they were unneeded

Gabe-Torres:
Cleaned up and removed the unnecessary migration files

db/migrate/20240718195301_update_form_answers_table.rb
```diff
@@ -0,0 +1,10 @@
+class UpdateFormAnswersTable < ActiveRecord::Migration[7.1]
```
kasugaijin, line 1: You should be able to combine AddOrganizationToFormAnswers and UpdateFormAnswersTable migrations. You created both of these on your branch, right? So you should be able to roll back the last two migrations and merge them together (you can keep the UpdateFormAnswersTable class name) and then run migrations again.

Gabe-Torres: Yeah, there is no need to split them up. They both update the FormAnswer table. I combined them and pushed the changes! 

I ended up using `rails db:migrate:down VERSION=20240716194544` (AddOrganizationToFormAnswers) to undo the changes of that migration and then added the Organization reference to the other.

I had some trouble using rollback. Something about NotNullViolation: ERROR. However, I think I could have done something similar to what you've done in some of your migration files with `def up` and `def down`. I haven't gone too deep into migrations and it's something I need more work on!

`def up` would be the changes made? and `def down` would be how to reverse those changes? So a command like rails rollback could work? I accomplished the same thing but in a roundabout way, I suppose. I need to look into that more!

Gabe-Torres:
Ah, new conflicts! Ill take care of those now.

db/migrate/20240718195301_update_form_answers_table.rb
kasugaijin: Ah yes I think you encountered that because you removed a null constraint, so undoing that would cause an error given there are records with null? That's just a guess. 

There's definitely a few ways to write migrations. We are soon adding the Strong Migrations gem that will ensure some steps are made in these files.

kasugaijin (approved):
LGTM! Thanks @Gabe-Torres

Gabe-Torres:
Of course! That was a fun one 👍

Issue #875, opened 2024-07-17
User belongs_to Person


kasugaijin:
We have a User model and a Person model. Person is a place to store information about an individual, and may or may not have a user account associated with it. Therefore, we want to set up the relationship so that User belongs to Person. You will need to update the Users table, and add a foreign key and index.

Update the model tests accordingly.

PR #882, opened 2024-07-19
User person

cflipse:
# 🔗 Issue

#875 

# ✍️ Description

This adds a `person_id` column to the users table in the database, and wires up the various associations.

There's also a little bit of "rails magic" in the form of a `before` hook on user creation, which ensures that a person will be connected to any newly created users:  first, it will attempt to match an existing user with the same email address, and failing that, a new person will be created with a basic set of attributes.

# 📷 Screenshots/Demos


kasugaijin (approved):
Looks good @cflipse

PR #888, opened 2024-07-22
Refactor donate page

kasugaijin:
…oster dashboard

# 🔗 Issue
https://github.com/rubyforgood/pet-rescue/issues/887

# ✍️ Description
PR refactors the donate page to use generic copy and render the org's donate link in the card. The view logic was moved to a partial that could be used in the adopter foster dashboard context as well, as I think it's nice to keep link clicks in the dashboard within the dashboard. Gets rid of no longer required donation routes and logic that was specifically for baja pet rescue's paypal page.

If there is a donation URL, that is displayed, otherwise the link to contact us is displayed.

These pages could do with some design love, but they are MVP ready for now. 

image

image

with URL
image

# 📷 Screenshots/Demos

### DIFF
--- a/app/controllers/organizations/adopter_fosterer/donations_controller.rb
+++ b/app/controllers/organizations/adopter_fosterer/donations_controller.rb
@@ -1,0 +1,10 @@
+module Organizations
+  module AdopterFosterer
+    class DonationsController < Organizations::BaseController
+      skip_verify_authorized only: [:index]
+      layout "adopter_foster_dashboard", only: :index
+
+      def index; end
+    end
+  end
+end
--- a/app/controllers/organizations/adopter_fosterer/donations_controller.rb
+++ b/app/controllers/organizations/adopter_fosterer/donations_controller.rb
@@ -4,7 +4,8 @@ module Organizations
       skip_verify_authorized only: [:index]
       layout "adopter_foster_dashboard", only: :index
 
-      def index; end
+      def index
+      end
     end
   end
 end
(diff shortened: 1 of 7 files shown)


app/controllers/organizations/adopter_fosterer/donations_controller.rb
```diff
@@ -0,0 +1,11 @@
+module Organizations
+  module AdopterFosterer
+    class DonationsController < Organizations::BaseController
+      skip_verify_authorized only: [:index]
```
kasugaijin, line 4: no need to authorize accessing this route as it's re-showing a route with public information, just inside the dashboard layout

jmilljr24 (commented):
On my machine the "contact us" is not displaying after an admin removes their donation link from the dashboard. Can you check if you have the same issue on your machine?

Steps to reproduce:
1. Save valid donate link to profile in dashboard settings.
2. Remove link and save profile.

Screenshot from 2024-07-26 07-46-44

kasugaijin:
> On my machine the "contact us" is not displaying after an admin removes their donation link from the dashboard. Can you check if you have the same issue on your machine?
> 
> Steps to reproduce:
> 
> 1. Save valid donate link to profile in dashboard settings.
> 2. Remove link and save profile.
> 
> Screenshot from 2024-07-26 07-46-44

Ah good catch I bet it’s an empty string and not nil.

jmilljr24 (approved):
lgtm!

Issue #886, opened 2024-07-21
Add Filter to Default Pet Tasks page


kasugaijin:
Default Pet Tasks are used to create a list of tasks that get automatically created on a new pet. These tasks can apply to all pets, or a specific species. We render these tasks in a table - sign in as staff, go to dashboard, go to Default Pet Tasks. 

It would be handy to have a filter so we can filter these tasks on Name, Species, Due After, and Recurring. Please use the same UI design as we do for the Fosters and Adoption Application Reviews page.

ErinClaudio:
@kasugaijin 
little help here, please.

is the page below of the tasks page you would like filtered?


<img width="1424" alt="Screenshot 2024-07-23 at 2 45 34 PM" src="https://github.com/user-attachments/assets/a6216672-f182-4f83-9adf-2e7fc4e001a0">

if so within the box shown below would you like to 

- show all those filters or 
- after clicking on the tasks you would like the list of dogs below to have a filter

<img width="293" alt="Screenshot 2024-07-23 at 2 47 11 PM" src="https://github.com/user-attachments/assets/0bd0484f-f782-4eba-b640-09287e6f7377">

kasugaijin:
Hey @ErinClaudio 

This is for the Default Pet Tasks page at `/staff/default_pet_tasks`. You can see this page if you log in as as staff and then go to the dashboard and click the Default Pet Tasks link in the navbar. 

The filter should look like the one we use for adoption application reviews page `/staff/adoption_application_reviews`

ErinClaudio:
It would help if I read the ticket correctly. Thank you

ErinClaudio:
Got it - I will take this one as long as that is ok.

ErinClaudio:
[start](https://gist.github.com/ErinClaudio/ff166c442dcb7fa0bb2e1dca2c8213b1)

kasugaijin:
@ErinClaudio haha no worries. Yes all yours. Please mimick the other filters that use the ransack gem (different from the details provided in your link)

PR #898, opened 2024-07-24
Add Filter to Default Pet Tasks page

ErinClaudio:
# 🔗 Issue
 [886](https://github.com/rubyforgood/pet-rescue/issues/886)

# ✍️ Description
This PR adds a filter for Default Pet Tasks in the `DefaultPetTasksController` index method. The filter includes options for Name, Species, Due After, and Recurring, using the same UI design as the Fosters and Adoption Application Reviews page.

### Changes:
- Added filtering logic to `DefaultPetTasksController`.
- Updated the index view to support the new filter options.

# 📷 Screenshots/Demos

https://github.com/user-attachments/assets/931abd26-e11a-4f94-ade8-f805e567a672


ErinClaudio:
https://github.com/rubyforgood/pet-rescue/issues/886

ErinClaudio:
**Notes for Erin**

- Are we only dealing with dogs and cats?

kasugaijin (commented):
Thanks @ErinClaudio 
I am looking at how we do this for Pets on the staff > pets controller and a lot of the logic is actually on the model to keep the controller thin. Ideally controllers only contain RESTful actions with some setter methods. Can we move the ransack logic out of the controller? This makes the code more reusable in another controller as well should we need to ransack this resource elsewhere in the future.

I would go a step further than we do on the Pet example, and put the logic into a model concern called something like DefaultPetTaskRansackable and then include that concern in the DefaultPetTask model so it has access to the methods. Then in the controller we can call these methods as needed and keep the logic in a reusable place, in its own file, and the controllers and models clean. Does this make sense?

ErinClaudio:
@kasugaijin Thank you! keeping that logic in the controller did not feel convention. Happy to make the change

app/views/organizations/staff/default_pet_tasks/_search_results.html.erb
```diff
@@ -0,0 +1,39 @@
+<div class="justify-content-md-between mb-4 mb-xl-0 gx-3">
```
kasugaijin, line 1: @ErinClaudio why do we have two partials to render the results? Can't we use the original table?

app/controllers/organizations/staff/default_pet_tasks_controller.rb
```diff
@@ -1,11 +1,20 @@
...
-    @default_pet_tasks = authorized_scope(DefaultPetTask.all).sort_by { |task| task.species }
+    ensure_due_in_days_in_q_params
+
+    tasks = authorized_scope(DefaultPetTask.all)
+
+    @q = DefaultPetTask.ransackable_tasks(tasks, params)
+    @default_pet_tasks = @q.result
+
+    @pagy, @default_pet_tasks = pagy(@default_pet_tasks, items: 10)
+    @partial_to_render = params[:q].present? ? "search_results" : "default_pet_tasks_table"
```
kasugaijin, line 17: See my comment re using one partial instead of two partials, below.

app/models/concerns/default_pet_task_ransackable.rb
```diff
@@ -0,0 +1,41 @@
...
+        elsif params[:q]["recurring"] == "false"
+          tasks = tasks.where(recurring: false)
+        end
+      end
+      tasks
+    end
+
+    def apply_due_in_days_filter(tasks, params)
+      if params[:q].present? && params[:q]["due_in_days"].present?
+        tasks = tasks.where("due_in_days >= ?", params[:q]["due_in_days"].to_i)
```
kasugaijin, line 36: Do we want to have an equality match here? So if I enter 5, it should return only tasks that have `due_in_days` value of 5? Right now it will return 5 and greater.

app/views/organizations/staff/default_pet_tasks/index.html.erb
```diff
@@ -4,67 +4,44 @@
...
+                <div class="form-group mb-3 col-md-4">
+                  <%= f.label :name_cont, t('.name') %>
+                  <%= f.text_field :name_cont, class: "form-control", placeholder: "Enter Pet Name" %>
+                </div>
+                <div class="form-group mb-3 col-md-4">
+                  <%= f.label :species_eq, t('.species') %>
+                  <%= f.select :species_eq, Pet.species.keys.map { |type| [type.titleize, type] }, { include_blank: 'Select Species' }, class: "form-select" %>
+                </div>
+                <div class="form-group mb-3 col-md-4">
+                  <%= f.label :due_in_days, t('.due_after') %>
```
kasugaijin, line 23: If we end up matching on equality, e.g. if I search `due_in_days` of 5 and only expect results with value of 5, then this should have `_eq` appended like `due_in_days_eq` ? Per https://activerecord-hackery.github.io/ransack/getting-started/search-matches/#search-matchers

app/views/organizations/staff/default_pet_tasks/index.html.erb
```diff
@@ -4,67 +4,44 @@
...
+                  <%= number_field_tag :due_in_days, params[:due_in_days], class: "form-control", placeholder: "Due After (Days)" %>
+                </div>
+                <div class="form-group mb-3 col-md-4">
+                  <%= f.label :recurring, t('.recurring') %>
```
kasugaijin, line 27: I think this should also be `recurring_eq` given we are looking for an equality match.

app/controllers/organizations/staff/default_pet_tasks_controller.rb
```diff
@@ -1,11 +1,20 @@
 class Organizations::Staff::DefaultPetTasksController < Organizations::BaseController
   before_action :context_authorize!, only: %i[index new create]
   before_action :set_task, only: %i[edit update destroy]
+  include ::Pagy::Backend
 
   layout "dashboard"
 
   def index
-    @default_pet_tasks = authorized_scope(DefaultPetTask.all).sort_by { |task| task.species }
+    ensure_due_in_days_in_q_params
```
kasugaijin, line 9: What is the `ensure_due_in_days_in_q_params` method for and why do we do this only for this parameter?

If we must have it, can we move `ensure_due_in_days_in_q_params` this to a before_action filter for index?

app/views/organizations/staff/default_pet_tasks/index.html.erb
```diff
@@ -4,67 +4,44 @@
...
+                  <%= f.label :name_cont, t('.name') %>
+                  <%= f.text_field :name_cont, class: "form-control", placeholder: "Enter Pet Name" %>
+                </div>
+                <div class="form-group mb-3 col-md-4">
+                  <%= f.label :species_eq, t('.species') %>
+                  <%= f.select :species_eq, Pet.species.keys.map { |type| [type.titleize, type] }, { include_blank: 'Select Species' }, class: "form-select" %>
+                </div>
+                <div class="form-group mb-3 col-md-4">
+                  <%= f.label :due_in_days, t('.due_after') %>
+                  <%= number_field_tag :due_in_days, params[:due_in_days], class: "form-control", placeholder: "Due After (Days)" %>
```
kasugaijin, line 24: I think we should use text `Due in (days)` for the placeholder

app/views/organizations/staff/default_pet_tasks/_search_results.html.erb
```diff
@@ -0,0 +1,39 @@
...
+  <div class="card">
+    <div class="table-responsive overflow-y-hidden">
+      <table class="table mb-0 text-nowrap table-hover table-centered">
+        <thead>
+          <tr>
+            <th scope="col"><%= t('.name') %></th>
+            <th scope="col"><%= t('.description') %></th>
+            <th scope="col"><%= t('.species') %></th>
+            <th scope="col"><%= t('.due_after') %></th>
```
kasugaijin, line 10: Let's change this to `Due in` instead of `Due after`

kasugaijin:
@ErinClaudio looks really good! Much better moving the logic to a concern like that. I have some questions and small comments.

ErinClaudio:
I will finish this up on monday

ErinClaudio:
@kasugaijin

app/views/organizations/staff/default_pet_tasks/_default_pet_tasks_table.html.erb
```diff
@@ -0,0 +1,49 @@
+<div class=" mb-4 mb-xl-0 gx-3">
```
kasugaijin, line 1: The table has lost its margin, or padding, from the left and right edges. Not a blocking issue, but worth seeing if you can fix so that the left and right margins of the table align with the left and right margins of the filter. 

image

kasugaijin (approved):
Looks great @ErinClaudio 
I will approve now. If you can fix the table width (see comment above) go for it, and then you can merge without anothe review. If you don't have time to do that, that's fine, then you can merge this right away.

app/views/organizations/staff/default_pet_tasks/_default_pet_tasks_table.html.erb
ErinClaudio: @kasugaijin let me know if the recent change is good to merge

kasugaijin: @ErinClaudio thank you

Issue #874, opened 2024-07-17
Add Strong Migrations gem


kasugaijin:
https://github.com/ankane/strong_migrations

PR #904, opened 2024-07-27
moving gem to global scope

nsiwnf:
# 🔗 Issue
Follow up of #874 

# ✍️ Description
Moving gem to be used in all environment


kasugaijin (approved):
LGTM

PR #906, opened 2024-07-27
Removes adopter foster profile

kasugaijin:
# 🔗 Issue
https://github.com/orgs/rubyforgood/projects/16/views/1?pane=issue&itemId=72013077

# ✍️ Description
Removes the AdopterFosterProfile. This work unblocks the replacement of the AdopterFosterAccount with the Person model.

Adopters cannot make adoptions currently because a FormSubmission is required (not an outcome of this PR, but a note as to what feature currently does not work). This will get wired up when we replace AdopterFosterAccount with Person.

# 📷 Screenshots/Demos


jmilljr24 (approved):
Looks good!

PR #907, opened 2024-07-27
Update donate page

kasugaijin:
# 🔗 Issue

# ✍️ Description
We were setting the wrong Org id in the controller, so this fixes that to set based on the current org scope. Also updates the view to look on the Org for the url instead of the org profile.

# 📷 Screenshots/Demos


jmilljr24 (approved):
Good stuff.

Issue #900, opened 2024-07-25
AdopterFoster: Add adopted pets to the dashboard


kasugaijin:
We need to enable adopters to go to their dashboard and view their adopted pets and the files associated with those pets. These files get uploaded by staff and are usually things like medical records, so should be shared with the adopter. The adopter should be able to download the files from the dashboard for a given pet.

We need:
- adopted_pets route scoped under org > adopter_fosterer
- adopted pets controller under the same scope
- Index action that displays the adopted pets by that user who is an adopter
- Policy and policy action to make sure the user is an authenticated adopter
- design the pet cards and table per the image below
- see the table in `app/views/organizations/staff/pets/_file_attachment_table.html.erb` except we do not need the buttons to attach or delete
- the icon used for Adopted Pets in the image below is a little dog. We should stick to what's available in our theme at https://geeksui.codescandy.com/ so take a look there and see if you can find something. I was thinking of the star from this dash nav https://geeksui.codescandy.com/geeks/pages/dashboard-instructor.html ... unless you can get a puppy to work without having to upload a new icon...then even better.

For this issue, just get the above working. We will then want to, later, update the table using turbo stream when you click on a pet to render that pet's files. If you really want to do that here, you can as well. I think we should use the show action on the same controller.

design
![Image](https://github.com/user-attachments/assets/76e30755-0692-47c1-b2f0-6ed2e7888d67)

sarvaiyanidhi:
Hi @kasugaijin ,

I can work on this issue.. Can you please assign it to me?

Thanks,
Nidhi

kasugaijin:
@sarvaiyanidhi all yours!

PR #910, opened 2024-08-02
900 - AdopterFoster: Add adopted pets

sarvaiyanidhi:
# 🔗 Issue
#900 

# ✍️ Description
Displaying Adopted Pets for Adopter account and listing its files on click of each pet.

# 📷 Screenshots/Demos

Alta-Pet-Rescue-Adopted-Pets-


kasugaijin:
@sarvaiyanidhi thanks for doing this, one of us will review this soon.

app/views/organizations/adopter_fosterer/adopted_pets/_pet_files.html.erb
```diff
@@ -0,0 +1,36 @@
+<div id="pet_files">
```
kasugaijin, line 1: This table is almost identical to the one we use on staff/pets to upload and show files, right? `app/views/organizations/staff/pets/_file_attachment_table.html.erb` The main difference is whether or not we want to render a delete button.

I am wondering if instead of duplicating a lof ot this logic here, if we can move this partial `app/views/organizations/staff/pets/_file_attachment_table.html.erb` to a shared folder e.g. at `aap/views/organizations/shared` and then conditionally render the delete button based on the User type, e.g. if the user role is staff we render the delete button. Does that make sense?

app/controllers/organizations/adopter_fosterer/adopted_pets_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+    class AdoptedPetsController < Organizations::BaseController
+      before_action :context_authorize!
+      layout "adopter_foster_dashboard"
+
+      def index
+        @adopted_pets = authorized_scope(AdopterApplication.joins(:pet)
+        .where(status: :adoption_made, profile_show: true), with: Organizations::AdopterFosterAdoptedPetPolicy)
+      end
+
+      def files
```
kasugaijin, line 12: We could create a new route and controller to handle this instead e.g., 
```
module Organizations
  module AdopterFosterer
    class AdoptedPetFilesController
```

And use the index action to get the list of files for a given pet. This will keep things RESTful so we don't have to use custom methods. What do you think?

app/helpers/organizations/adopter_fosterer/adopted_pets_helper.rb
kasugaijin, line 1: i think we can get rid of this? I assume this was create when using a Rails generator?

app/javascript/controllers/card_controller.js
kasugaijin, line 1: I like this, it can be re-used for other situations with multiple cards to select from.

app/controllers/organizations/adopter_fosterer/adopted_pets_controller.rb
```diff
@@ -0,0 +1,33 @@
+module Organizations
+  module AdopterFosterer
+    class AdoptedPetsController < Organizations::BaseController
+      before_action :context_authorize!
+      layout "adopter_foster_dashboard"
+
+      def index
```
kasugaijin, line 7: Could we instead query the `Match` model to find all matches with this `adopter_foster_account_id` where the `match_type: :adoption`?

app/policies/organizations/adopter_foster_adopted_pet_policy.rb
```diff
@@ -0,0 +1,15 @@
+class Organizations::AdopterFosterAdoptedPetPolicy < ApplicationPolicy
```
kasugaijin, line 1: Ah yes this is close to what I had in mind with this comment https://github.com/rubyforgood/pet-rescue/pull/910/files#r1720837255...I should have looked further down :) 

Still, we should be querying Match instead of AdopterApplication. 
Also, this name `Organizations::AdopterFosterAdoptedPetPolicy` is duplicating a common module name in this app `AdopterFosterer` so let's add that directory so we have `Organizations::AdopterFosterer::MatchPolicy` (I think the policy names ideally mirror the model names, right?)

config/routes.rb
```diff
@@ -61,6 +61,11 @@
       resources :dashboard, only: [:index]
       resources :likes, only: [:index, :create, :destroy]
       resources :adopter_applications, path: "applications", only: %i[index create update]
+      resources :adopted_pets, only: [:index] do
```
kasugaijin, line 64: https://github.com/rubyforgood/pet-rescue/pull/910/files#r1720837255 to follow up on this comment.

I think something like below is cleaner. 
```
resources :adopted_pets, only: [:index] do
  resources :files, only: [:index], module: :adopted_pets
end
```

test/integration/organizations/adopter_fosterer/adopted_pet_test.rb
kasugaijin, line 1: I think controller tests are more maintinable in the long run. If we start asserting dom elements it makes the tests very fragile when we come to modify the UI, which happens often enough to become annoying :)

I think we should keep the policy tests, and add controller tests to assert the actions return what we expect in the collections, and then get rid of integration tests asserting DOM elements and System tests.

kasugaijin (commented):
@sarvaiyanidhi great start on this! The functionality is all good. I have some comments about structure and approach in some cases. Let me know if you have any questions on these comments :)

sarvaiyanidhi:
> @sarvaiyanidhi great start on this! The functionality is all good. I have some comments about structure and approach in some cases. Let me know if you have any questions on these comments :)

Thanks @kasugaijin for the detailed feedback. Have gone through the comments and will push the required changes soon.

kasugaijin:
@sarvaiyanidhi how's this going?

sarvaiyanidhi:
> @sarvaiyanidhi how's this going?

Hey sorry for the delay... Will work and push changes in 1-2 days

kasugaijin:
@sarvaiyanidhi not at all! No rush. Just making sure you’re not blocked.

app/views/organizations/adopter_fosterer/adopted_pets/_pet_files.html.erb
sarvaiyanidhi: Thanks for the input. Common file does make sense and easy to maintain. Have updated changes for this.

app/controllers/organizations/adopter_fosterer/adopted_pets_controller.rb
sarvaiyanidhi: Nice.. I liked this approach to keep things more RESTful. Have made the required changes and created new files_controller.rb

sarvaiyanidhi:
Hi @kasugaijin,

Sorry took some more time than expected as was tied up with few things.. I have addressed all points and pushed code. Can you please check if changes made for query using Match model is as per requirement. Also, I am facing one issue with authorize test in controller test, not able to figure why it is failing if you can help with it.

Thanks

kasugaijin:
> Hi @kasugaijin,
> 
> Sorry took some more time than expected as was tied up with few things.. I have addressed all points and pushed code. Can you please check if changes made for query using Match model is as per requirement. Also, I am facing one issue with authorize test in controller test, not able to figure why it is failing if you can help with it.
> 
> Thanks

Thanks @sarvaiyanidhi. Can you post a picture of the error screenshot in the pet rescue slack channel? It's the fastest way to get unblocked.

kasugaijin (approved):
Awesome @sarvaiyanidhi !

Issue #899, opened 2024-07-25
Adopter Foster Dashboard - update the liked pets design


kasugaijin:
Sign in as an adopter and then like some pets on the adoptable pets page. Then go to the dashboard and look at the Liked Pets link and you will see the liked pets. This issue is just to update the UI to make it look per the image below. Please use the existing avatar component. Please copy over the heart from the existing card along with the turbo stream so it will make the request to unlike or like the pet when you click the heart

![Image](https://github.com/user-attachments/assets/279ea9a5-dcb2-4a49-942b-4a711419943d)

ErinClaudio:
@kasugaijin Happy to start this one once the other is approved.

ErinClaudio:
[Gist](https://gist.github.com/ErinClaudio/001e9862beafa1231578444c9b72a85b)

kasugaijin:
Go for it! I would say for this we don't need to test, actually. So don't worry about that. I am not a fan of tests that couple closely with UI elements as they are a pain to manage when changing the UI at all.

kasugaijin:
Hey @ErinClaudio let's just leave as it is. We can always add in the broadcast later to update the page, but that does add additional overhead (and I suspect some undesirable overhead because it requires we touch a model) and at this stage for MVP it's not necessary.

ErinClaudio:
@kasugaijin 
I believe this can be marked complete per
https://github.com/rubyforgood/pet-rescue/pull/912

PR #912, opened 2024-08-02
Adopter foster dashboard update the liked pets design

ErinClaudio:
# 🔗 Issue

(https://github.com/rubyforgood/pet-rescue/issues/899)

# ✍️ Description
This PR updates the UI for displaying liked pets on the adopter's dashboard. The changes involve using the existing AvatarComponent to render the pet's image and name. Additionally, the heart icon has been copied over from the existing card component, maintaining the Turbo Stream functionality, so that users can like or unlike pets directly from the dashboard by clicking the heart icon.

# 🔗 Issue

# 📷 Screenshots/Demos
https://github.com/user-attachments/assets/2c678763-339d-45bf-8776-fb7be140fe54


ErinClaudio:
[899](https://github.com/rubyforgood/pet-rescue/issues/899)

app/components/avatar_component.rb
```diff
@@ -2,13 +2,17 @@
 
 # Renders a User's avatar as image or user's initials
 class AvatarComponent < ApplicationComponent
```
kasugaijin, line 4: It feels weird to have User as a param and Pet as an option. Can they both be options instead for consistency?

app/components/avatar_component.rb
```diff
@@ -20,7 +24,11 @@ def image_url
   end
 
   def initials
-    "#{user.first_name[0]}#{user.last_name[0]}".upcase
+    if pet
```
kasugaijin, line 27: Maybe we can just use the first letter since the names are all singular (no second names)

kasugaijin:
@ErinClaudio looks good! I have thought about the implementation. It looks like you're modifying the AvatarComponent to accommodate Pets, and in this particular use case, it also needs to accept likes so you can add the LikeButtonComponent. However, in doing this, we are taking a concept that handled user avatars, and now expanded it to handle Pets and Likes. It feels like these should be handled separately, because user avatars have nothing to do with Likes, if you get me? 

I would instead create a new PetAvatarComponent that we can also accept a like argument on. These two concepts make sense to relate with each other. It will simplify the logic in the component as well, so we don't need to have conditionals to find the correct image paths. What do you think about that?

ErinClaudio:
@kasugaijin 
Thank you for the review. Just to clarify, I should remove the pet and like-related logic from avatar_component.rb, create a new PetAvatarComponent to handle that logic, and then update any downstream code accordingly?

kasugaijin:
> @kasugaijin Thank you for the review. Just to clarify, I should remove the pet and like-related logic from avatar_component.rb, create a new PetAvatarComponent to handle that logic, and then update any downstream code accordingly?

Yes I think that will be a bit cleaner

ErinClaudio:
@kasugaijin

app/components/pet_avatar_component.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+# Renders a User's avatar as image or user's initials
```
kasugaijin, line 3: This comment needs updating

app/components/pet_avatar_component.rb
```diff
@@ -0,0 +1,41 @@
...
+class PetAvatarComponent < ApplicationComponent
+  option :pet, Types::Instance(Pet), optional: true
```
kasugaijin, line 5: Is there a case we would use this Class without passing in a Pet? The benefit of requiring is it will cause an error if missing from the args.

app/components/pet_avatar_component.rb
```diff
@@ -0,0 +1,41 @@
...
+  option :like, optional: true
+
+  private
+
+  def pet_avatar
```
kasugaijin, line 10: And I think this would better to be
pseudocode:
```
if pet_image exists, then return that image
otherwise, create the initials
```
I think this woul not render the initials if the pet did not have an image right now?

kasugaijin (approved):
Thanks @ErinClaudio

Issue #913, opened 2024-08-07
Organization: add null constraints and update org create service


kasugaijin:
Please add null constraints on organizations table in the database for `name`, `slug`, `email` and then update then add email attribute to the Organizations::CreateService when creating an Organization.

Also on the Organization model, we need to validate:
`name` for presence
`slug` for presence and also do not allow any characters that are forbidden in URLs. Look into options for validating this slug.
`email` for format, presence.

Gabe-Torres:
Seems like a fun one! I'll take it

kasugaijin:
@Gabe-Torres all yours!

coalest:
Looks like this task was completed with #918. Can this issue be closed?

PR #918, opened 2024-08-16
913 add null constraints and update org create service

Gabe-Torres:
# 🔗 Issue
#913 

# ✍️ Description
This Pr adds null constraints to the Organization model name, email, and slug attributes. 
Initially, I tried to add the null constraints by simply adding the columns.
Rails didn't approve and gave the following error

"Dangerous operation detected #strong_migrations ===
Setting NOT NULL on an existing column blocks reads and writes while every row is checked"

It then guided me to add a check constraint and validate it in a separate migration.
Adding NOT NULL constraints on an existing column can be dangerous in a production environment and
can lock the table and block reads and writes, which is why strong_migrations recommended a safer approach.

This will also allow for easier rollbacks if needed.

Also adds: 
- email attribute on Organizations Create Service
- validation for presence and format for the following organization attributes:
name, slug, and email
- All corresponding test updates


app/models/organization.rb
```diff
@@ -42,6 +42,9 @@ class Organization < ApplicationRecord
   before_save :normalize_phone
 
   validates :phone_number, phone: {possible: true, allow_blank: true}
+  validates :name, presence: true
+  validates :slug, presence: true, format: {with: /\A[a-zA-Z0-9_-]+\z/, message: "only allows letters, numbers, hyphens, and underscores"}
+  validates :email, presence: true
   validates_format_of :email, with: /\A([^@\s]+)@((?:[-a-z0-9]+\.)+[a-z]{2,})\z/i, allow_blank: true
```
kasugaijin, line 48: I think we can get rid of the `allow_blank: true` here now that we are requiring an email

kasugaijin:
@Gabe-Torres Ah yes the strong migrations gem, rightfully so, will not allow you to apply these constraints. However, we don't have any production data t the moment and we don't have to adhere to these process the gem sets out. Have you tried writing the migration to add the null constraints on the three columns by wrapping it in a `safety_assured` block?

We had strong migrations installed early on in this project and ran a lot of migrations it moaned about...and we had to wrap each migration in the `safety_assured` block. See this PR that removed the gem and this block in the migrations https://github.com/rubyforgood/pet-rescue/pull/400/files 

We can take the same approach here to write a simpler migration that just adds the null constraints. You will probably need to drop your db and go from there to run the migrations and seed it for it to work properly. That is fine as we do not care about the data. 

Also, should we not also update the seed files to add the required attributes to the code creating the Organizations?

Let me know if you have any questions.

kasugaijin:
@Gabe-Torres how's this one going?

kasugaijin:
@Gabe-Torres bump

Gabe-Torres:
I'll get these changes submitted asap! @kasugaijin

Gabe-Torres:
> @Gabe-Torres Ah yes the strong migrations gem, rightfully so, will not allow you to apply these constraints. However, we don't have any production data t the moment and we don't have to adhere to these process the gem sets out. Have you tried writing the migration to add the null constraints on the three columns by wrapping it in a `safety_assured` block?
> 
> We had strong migrations installed early on in this project and ran a lot of migrations it moaned about...and we had to wrap each migration in the `safety_assured` block. See this PR that removed the gem and this block in the migrations https://github.com/rubyforgood/pet-rescue/pull/400/files
> 
> We can take the same approach here to write a simpler migration that just adds the null constraints. You will probably need to drop your db and go from there to run the migrations and seed it for it to work properly. That is fine as we do not care about the data.
> 
> Also, should we not also update the seed files to add the required attributes to the code creating the Organizations?
> 
> Let me know if you have any questions.
Updated! 
note - 
looks like both the alta and baja seed files organization instances have the required attributes! But maybe I'm missing something?
```ruby
@organization = Organization.create!(
  name: "Baja",
  slug: "baja",
  email: "baja@email.com",
  phone_number: "250 816 8212",
  custom_page: CustomPage.new(hero: "hero text", about: "about us text")
)
```


```ruby
@organization = Organization.create!(
  name: "Alta Pet Rescue",
  slug: "alta",
  email: "alta@email.com",
  phone_number: "250 816 8212",
  custom_page: CustomPage.new(hero: "Where every paw finds a home", about: "Alta was founded by an incredible group of ladies in April of 2020. Our initial goal was to have both a rescue and a spay/neuter clinic, however, we quickly realized that it would be more efficient to separate into two organizations.")
)
```

app/services/organizations/create_service.rb
```diff
@@ -24,7 +24,8 @@ def signal(args)
     ActiveRecord::Base.transaction do
       create_organization(
         args[:organization][:name],
-        args[:organization][:slug]
+        args[:organization][:slug],
+        args[:organization][:email]
       )
       create_location(
```
kasugaijin, line 30: I think we can get rid of the location, too right? It used to be required, but we are not validating for it.

kasugaijin (approved):
@Gabe-Torres looks great! 
Just one comment - not something you have missed, just may as well address it while we are working on this specific logic.

Gabe-Torres:
Gotcha! Made those changes as well 👍🏽

kasugaijin (approved):
Thank you @Gabe-Torres

Issue #915, opened 2024-08-11
Implement Person


mononoken:
For [Conference Discussion Refactoring](https://github.com/orgs/rubyforgood/projects/16/views/1#)

> AdopterFosterAccount remove and wire up Person (TBD)
> StaffAccount remove and wire up Person (TBD)

The account models are no longer necessary and can be replaced by the new `Person` model now.

mononoken:
#919 completes this for adopters and fosterers

PR #919, opened 2024-08-25
915 implement person model

mononoken:
# 🔗 Issue
#915 

# ✍️ Description

Basically, this removes AdopterFosterAccount and wires up Person to be used in its place wherever necessary.

If you have any questions, please let me know, or we can discuss more in the next meeting.

I fixed some unrelated bugs and such and cleaned up some things while making this refactoring as well. I also fixed some behavior that wasn't working on main where I could. For example, adopters should be able to submit adoptions now (if they have a form submission, which they can't currently create themselves but we can make in console).

# 📷 Screenshots/Demos

None. Everything at the top level should mostly behave as it did before.


mononoken:
Not sure what is going on with the failing tests. I'll try to take a look later this week.

mononoken:
Getting some flakey test behavior on my machine which think is reflected on main too. Not sure if I will have time to investigate it until the weekend.

kasugaijin:
@mononoken I should get to take a look this weekend! Thanks for this PR.

mononoken:
> @mononoken I should get to take a look this weekend! Thanks for this PR.

Let me know if you have any questions or find any mistakes!

app/controllers/organizations/adoptable_pets_controller.rb
```diff
@@ -21,14 +22,14 @@ def show
...
     authorize! @pet, with: Organizations::AdoptablePetPolicy
 
-    if current_user&.adopter_foster_account
+    if current_user
       @adoption_application =
         AdopterApplication.find_by(
-          pet_id: @pet.id,
-          adopter_foster_account_id: current_user.adopter_foster_account.id
+          pet: @pet,
+          form_submission: current_user.person.form_submission
```
kasugaijin, line 29: I was thinking we would need to call `person.form_submission.last` because we would allow for multiple form submissions/sets of answers. But as I think of it, perhaps it makes sense to just keep one set (the latest set) of answers as the 'truth'. There's no point in keeping record of historical answers, right? We currently declare `has_one :form_submission` on `Person` so no need to change that in this case. Just raising here for a thought.

I was originally thinking we would track all form submissions i.e., keep the answers every time a form is submitted for a given person, and just just the last one for Matches. But, I cannot think of a compelling reason to have multiple at this stage. Can you?

app/controllers/organizations/staff/matches_controller.rb
```diff
@@ -32,7 +32,7 @@ def destroy
   private
 
   def match_params
-    params.require(:match).permit(:pet_id, :adopter_foster_account_id, :match_type, :start_date, :end_date)
+    params.require(:match).permit(:pet_id, :match_type, :start_date, :end_date)
```
kasugaijin, line 35: Does this need a `:person_id` ?

app/mailers/adoption_mailer.rb
```diff
@@ -1,11 +1,8 @@
...
   def reminder(match)
-    @match = match
-
-    @adopter_foster_account = match.adopter_foster_account
-    @user = @adopter_foster_account.user
+    person = match.person
     @pet = match.pet
 
-    mail to: @user.email, subject: "#{@pet.name}'s Reminder"
+    mail to: person.email, subject: "#{@pet.name}'s Reminder"
```
kasugaijin, line 6: We don't need the person instance variable in the view template do we?

app/models/adopter_application.rb
kasugaijin, line 1: Do we want to add similar validations here to prevent a Person from applying on a pet if they have already done so?

kasugaijin: Ah I see there is a check in the policy as well. Might still be best to keep a model level check just in case?

kasugaijin:
@mononoken looking good. I have left a few questions and comments. 
One key thing we will need to do is create the Person's form submission when the account is created. You're working on this right, @MooseCowBear ?

MooseCowBear:
@kasugaijin @mononoken Taking care of this. As of now, I'm creating a form submission like: 

```
def create
    super do |resource|
      if resource.persisted?
        ::Organizations::FormSubmissionService.new(resource).create
      end
    end
  end
```
because the form submission belongs to person and person is being created in a callback for user. Not sure that's the best place for it. We could create it after person is created in the callback as well?

kasugaijin:
> @kasugaijin @mononoken Taking care of this. As of now, I'm creating a form submission like:
> 
> ```
> def create
>     super do |resource|
>       if resource.persisted?
>         resource.add_role(:adopter, Current.organization)
>         ::Organizations::FormSubmissionService.new(resource).create
>       end
>     end
>   end
> ```
> 
> because the form submission belongs to person and person is being created in a callback for user. Not sure that's the best place for it. We could create it after person is created in the callback as well?

Ah yes good call. I think we are at risk of an issue here if the Person does not get created before we try to create the FormSubmission. You are right. I think we should instead have an after_create callback on Person that will create their form submission.

app/controllers/organizations/staff/matches_controller.rb
mononoken: Added 85edebb507fd

mononoken: This commit also fixes creating adoptions

app/models/adopter_application.rb
mononoken: That sounds right but think it is not in the scope of this PR. Should be a fairly easy issue. Would just need to add a validation, and it would be good to create a unique index on the postgres server too.

mononoken:
@kasugaijin , again, thank you for reviewing this one! I think I addressed all your comments. I fixed the adoption creation from reviews as well as the sign up error too.

kasugaijin (approved):
LGTM!

Issue #862, opened 2024-07-03
Move pet image in the staff pets show pages


mononoken:
## Description

**tl;dr:** We want to remove the right side image from the staff pets show page and add a thumbnail in its place to the header of the page.

We have received negative feedback from developers in regards to the image found on the staff pets show page. Currently, this page has multiple tabs for staff to navigate through for a pet. There is an image of the pet to the right.

Many of these tabs could be reusing partials from other pages. However, those partials in other locations do not have this image, and so we end up having to redesign a partial to be used for multiple pages but with different width constrictions. The image does not serve any particular benefit either other than helping signal to staff which pet they are looking at currently.

We want to remove the image, add a thumbnail, and we also need to adjust some of the html/css here. With the image removed, we will want the content of the page to take up the space that the image used to occupy. That way, we can share partials from other pages and they will have the same space constraints.

We also may need to adjust the html to allow for the thumbnail to exist as shown in the screenshot.

And finally, make sure the layout still works well for mobile.

![Image](https://github.com/rubyforgood/pet-rescue/assets/81536479/18d893ae-702e-4a54-9afb-b196170c747d)

https://www.figma.com/design/0jVgYASUJy0KiX3BVc3dFM/Tasks?node-id=1094-70&t=WZjsRcM8EoT4DZdu-1
_Note: Ignore the content in the box; this PR is only concerned about the thumbnail_
![Image](https://github.com/rubyforgood/pet-rescue/assets/81536479/3652789c-4a73-4341-b0bf-840eb928e827)

## Acceptance Criteria
- [ ] Remove the image to the right side
- [ ] Add thumbnail image to the header space of the page layout per figma design
- [ ] Page content extends full page width in lieu of removed image

Aaryanpal:
Hi @mononoken I can take up this issue. Can you please assign it to me? Thanks

jmilljr24:
@Aaryanpal All yours! Let us know if you have any questions.

kasugaijin:
How's this doing @Aaryanpal ?

kasugaijin:
bump @Aaryanpal need any help on this?

kasugaijin:
@Aaryanpal thanks for the update! No rush at all. Take your time. I just like to make sure people are still interested otherwise I like to reassign.

Aaryanpal:
@kasugaijin Thanks for Understanding.

Aaryanpal:
@kasugaijin 
Is it Seems Right? the image size

[...]

PR #924, opened 2024-08-27
Move staff pets show image

Aaryanpal:
# 🔗 Issue
#862 

# ✍️ Description
- [x] Remove the image to the right side
- [x] Add thumbnail image to the header space of the page layout per figma design
- [x] Page content extends full page width in lieu of removed image

# 📷 Screenshots/Demos

https://github.com/user-attachments/assets/d2ab1990-5170-4a0f-98cf-ed8b6abcc1bd

### DIFF
--- a/app/views/organizations/staff/pets/show.html.erb
+++ b/app/views/organizations/staff/pets/show.html.erb
@@ -19,7 +19,9 @@
 
   <% c.with_body do %>
     <div class="row">
+      <div class = "col-12">
         <%= render partial: "organizations/staff/pets/tabs/#{@active_tab}", locals: { applications: @pet.adopter_applications, pet: @pet, fosters: @pet.matches.fosters } %>
+      </div>
     </div>
   <% end %>
 <% end %>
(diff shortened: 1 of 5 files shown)


app/views/organizations/staff/pets/show.html.erb
```diff
@@ -1,5 +1,13 @@
 <%= render DashboardPageComponent.new(crumb: :dashboard_pet, crumb_options: [@pet]) do |c| %>
-  <% c.with_header_title { @pet.name } %>
+  <% c.with_header_title do %>
```
kasugaijin, line 2: @Aaryanpal we just made a new pet avatar component, actually. Could you please use this here instead?

app/components/pet_avatar_component.rb

Aaryanpal: okay

Aaryanpal: [Screencast from 2024-08-28 20-06-53.webm](https://github.com/user-attachments/assets/672b1140-7df1-4606-ba5a-68363a7be735)

@kasugaijin, is it okay to show the initial letter of the pet's name when an image is not available? Also, should we update the pets index and fix the images to ensure they appear rounded?

image

kasugaijin (approved):
Looks good @Aaryanpal 
Thanks for addigin more options to the component. 
I think showing the pet initial is fine for now. We can update this later if we need to show something else. 

I do agree that we should use this avatar in the Pet index table - so if you want to work on that, please create a new issue and have a go!

Thanks

Issue #926, opened 2024-08-28
Staff Dashboard - make overdue tasks the default shown table


kasugaijin:
When you log in as staff and go to the dashboard you will see some tiles and empty space below. Let's change this so that when you visit the Dahboard page, you see the overdue tasks table loaded by default. Let's maintain the existing behaviour too, so that clicking the links for incomplete tasks and overdue tasks will load the respective tables. 
Let's also add a header so it is clear what table you are looking at, and change the column to say 'count'
There is no need to add any tests for this UI change. 

![Image](https://github.com/user-attachments/assets/3f5be1f6-406c-420a-8bed-ed20a605e4c0)

Aaryanpal:
@kasugaijin starting with this

PR #943, opened 2024-08-30
926 Staff Dashboard - make overdue tasks the default shown table

Aaryanpal:
# 🔗 Issue
#926

# ✍️ Description
- [x] Set the Overview table as the default
- [x] Added a header to the table to make it more informative
- [x] Used Pet Avatar Component for Image 

# 📷 Screenshots/Demos

[Screencast from 2024-08-30 20-01-55.webm](https://github.com/user-attachments/assets/34388952-9378-4b39-808b-d29de897fa2b)


kasugaijin:
@Aaryanpal just need an update to the test - see broken test in pipeline - assuming you changes some text it expects to see.

app/controllers/organizations/staff/dashboard_controller.rb
```diff
@@ -55,4 +47,18 @@ def context_authorize!
     authorize! :dashboard,
       context: {organization: Current.organization}
   end
+
+  def calculate_overdue_tasks
```
kasugaijin, line 51: I think we can move this to a controller concern to keep this controller thin. We can call it something like `PetTaskable` or whatever you thin works. I am thinking it will be a concern to make queries on Tasks in the scope of a pet. Then we can add this method there

app/views/organizations/staff/dashboard/index.html.erb
```diff
@@ -46,6 +46,7 @@
         <% end %>
       </div>
     </div>
+    <%= render 'task'%>
     <%= turbo_frame_tag "tasks-frame" %>
```
kasugaijin, line 50: Do we need this turbo frame tag still?

app/views/organizations/staff/dashboard/_task.html.erb
```diff
@@ -0,0 +1,43 @@
+<turbo-frame id="tasks-frame">
```
kasugaijin, line 1: Can we use `<%= turbo_frame_tag "tasks-frame" %>` instead?

app/controllers/organizations/staff/dashboard_controller.rb
```diff
@@ -22,30 +23,21 @@ def incomplete_tasks
       items: 5
     )
     @column_name = "Incomplete Tasks"
+    @header_title = "Incomplete Table"
```
kasugaijin, line 26: Could label this `Incomplete Pet Tasks`?

app/controllers/organizations/staff/dashboard_controller.rb
```diff
@@ -55,4 +47,18 @@ def context_authorize!
...
+      Pet
+      .left_joins(:tasks)
+      .select("pets.*, COUNT(tasks.id) AS incomplete_tasks_count")
+      .where(tasks: {completed: false})
+      .where("tasks.due_date < ?", Time.current)
+      .group("pets.id"),
+      items: 5
+    )
+    @column_name = "Count"
+    @header_title = "Overdue Table"
```
kasugaijin, line 62: And name this `Overdue Pet Tasks`?

kasugaijin:
@Aaryanpal looks good! Just a few comments, nothing major. 
I notice we are using Pagy and expect 5 records, but in your demo video the paginated collections are much larger than 5. IT would be good if we keep this table small, at 5 records per page. Do you know why there is the discrepancy?

app/controllers/organizations/staff/dashboard_controller.rb
Aaryanpal: Moved it into the `PetTaskable.rb` concern

app/views/organizations/staff/dashboard/index.html.erb
Aaryanpal: Yes, they are still in use, but I'm not very familiar with them. I need some time to understand the underlying functionality before integrating it with the current scenario. For now, it's working fine.

app/controllers/organizations/staff/dashboard_controller.rb
Aaryanpal: Done

app/controllers/organizations/staff/dashboard_controller.rb
Aaryanpal: Done

Aaryanpal:
- [x] Updated Screencast

[Screencast from 2024-08-30 23-42-49.webm](https://github.com/user-attachments/assets/bc4323ba-dc6f-41e5-9ced-77642494c849)

Aaryanpal:
> @Aaryanpal looks good! Just a few comments, nothing major. I notice we are using Pagy and expect 5 records, but in your demo video the paginated collections are much larger than 5. IT would be good if we keep this table small, at 5 records per page. Do you know why there is the discrepancy?

Wrong
```
  @pagy, paginated_adoptable_pets = pagy(
      @q.result,
      items: 9
    )
```
Right

```
  @pagy, paginated_adoptable_pets = pagy(
      @q.result,
      limit: 9
    )
```
We're using `items` to restrict the records, but we should be using `limit` instead to properly limit them. Could you create an issue so I can replace this throughout the entire codebase?

[Reference](https://ddnexus.github.io/pagy/docs/how-to/#control-the-items-per-page)

app/views/organizations/staff/dashboard/_task.html.erb
kasugaijin: It's just a rails helper that will create the same markup as you have there.

app/views/organizations/staff/dashboard/index.html.erb
kasugaijin: If you look at the markup in the dev console it should just be `<turbo-frame id="tasks-frame">` because `    <%= turbo_frame_tag "tasks-frame" %>` is just a helper from rails to create the markup.

kasugaijin:
> > @Aaryanpal looks good! Just a few comments, nothing major. I notice we are using Pagy and expect 5 records, but in your demo video the paginated collections are much larger than 5. IT would be good if we keep this table small, at 5 records per page. Do you know why there is the discrepancy?
> 
> Wrong
> 
> ```
>   @pagy, paginated_adoptable_pets = pagy(
>       @q.result,
>       items: 9
>     )
> ```
> 
> Right
> 
> ```
>   @pagy, paginated_adoptable_pets = pagy(
>       @q.result,
>       limit: 9
>     )
> ```
> 
> We're using `items` to restrict the records, but we should be using `limit` instead to properly limit them. Could you create an issue so I can replace this throughout the entire codebase?
> 
> [Reference](https://ddnexus.github.io/pagy/docs/how-to/#control-the-items-per-page)

Ah thank you, I made an issue here https://github.com/rubyforgood/pet-rescue/issues/944

app/views/organizations/staff/dashboard/index.html.erb
kasugaijin: I think you can delete the `<%= turbo_frame_tag "tasks-frame" %>` given you moved the same frame into the partial now?

app/controllers/organizations/staff/dashboard_controller.rb
```diff
@@ -55,4 +37,16 @@ def context_authorize!
     authorize! :dashboard,
       context: {organization: Current.organization}
   end
+
```
kasugaijin, line 40: These two methods can be private, right?

app/controllers/organizations/staff/dashboard_controller.rb
```diff
@@ -12,40 +14,20 @@ def index
   end
 
   def incomplete_tasks
```
kasugaijin, line 16: A thought here - we are returning a collection of pets. But we are conflating them with tasks. Would it be more accurate to change the method names to `pets_with_incomplete_tasks` and `pets_with_overdue_tasks` and change the partial name to `pets` instead of `tasks`?

app/controllers/organizations/staff/dashboard_controller.rb
Aaryanpal: Yes they are indeed

app/controllers/organizations/staff/dashboard_controller.rb
Aaryanpal: hmm.. in that case we need to change the routes as well.

kasugaijin (approved):
Looks good. Thank you! @Aaryanpal I think that new naming is a bit clearer for anyone looking at this for the first time. What do you think?

Issue #931, opened 2024-08-28
Adopter application review: When a pet is adopted disable the status dropdown and remove the adopt button


kasugaijin:
Log in as staff and go to the adopter applcation reviews page. The status dropdowns here update via turbostream. If you update one to successful applicant, the new adoption button is enabled, click that and you have made a new adoption. Go back to that adopter application and you will be able to update the status again. This is incorrect. The status dropdown should be disabled if an adoption is made. Also, the new adoption button should be removed if an adoption is made.
No need to add tests.

![Image](https://github.com/user-attachments/assets/5e13dda8-8402-42e7-9aac-997d3d301816)

Aaryanpal:
@kasugaijin please assign this so that i can start.

PR #957, opened 2024-09-05
931- Disable drop-down status button on adoption made and remove new adoption made button

Aaryanpal:
# 🔗 Issue
#931 

# ✍️ Description
- [x] Disable the select button when Adoption made is set
- [x] Remove the New Adoption Made when Adoption made is already set 

# 📷 Screenshots/Demos
- [x] ScreenCast Attatched

[Screencast from 2024-09-05 09-42-57.webm](https://github.com/user-attachments/assets/6eba0307-2981-484a-9940-e1cbe246cf5e)


app/views/organizations/staff/pets/tabs/partials/_application_status.html.erb
```diff
@@ -1,16 +1,18 @@
...
   </button>
-  <ul class="dropdown-menu">
-    <% AdopterApplication.statuses.keys.each do |status| %>
-      <%unless status == "adoption_made" %>
-        <li>
-          <%= button_to t("organizations.staff.pets.application_status.#{status}"),
-                        staff_adoption_application_review_path(app, params: { adopter_application: { status: status}}),
-                        :method => :patch, class: 'dropdown-item' %>
-        </li>
+  <% unless app.status == 'adoption_made' %>
```
kasugaijin, line 5: Do we need this check here? We already disable it above if the `app.status == 'adoption_made'` on page load, and when we create a new adoption [we redirect back](https://github.com/rubyforgood/pet-rescue/blob/34706588950d68476d5daac1e30d18464939a821/app/controllers/organizations/staff/matches_controller.rb#L5) which refreshes the page anyway...so that should be all we need?


Also, if you look at the [Create Adoption button](https://github.com/rubyforgood/pet-rescue/blob/34706588950d68476d5daac1e30d18464939a821/app/views/organizations/staff/pets/tabs/partials/_application_create_button.html.erb#L1), it is making a request with turbo stream. This is unnecessary I think. We can get away with an html request and it will redirect back. Do you agree?

Aaryanpal: > Do we need this check here? We already disable it above if the `app.status == 'adoption_made'` on page load, and when we create a new adoption [we redirect back](https://github.com/rubyforgood/pet-rescue/blob/34706588950d68476d5daac1e30d18464939a821/app/controllers/organizations/staff/matches_controller.rb#L5) which refreshes the page anyway...so that should be all we need?
> 

Implementing this optimizes DOM composition by eliminating extraneous HTML to get render

Aaryanpal: > Also, if you look at the [Create Adoption button](https://github.com/rubyforgood/pet-rescue/blob/34706588950d68476d5daac1e30d18464939a821/app/views/organizations/staff/pets/tabs/partials/_application_create_button.html.erb#L1), it is making a request with turbo stream. This is unnecessary I think. We can get away with an html request and it will redirect back. Do you agree?

Yes u are right, 
Done with the changes

kasugaijin (approved):
LGTM!

Issue #951, opened 2024-09-04
Add URL param on Staff Dashboard results


kasugaijin:
As staff, if you go to the dashboard and look at the overdue and incomplete pet task tables, they render a Pet. The pet name can be clicked to go to the pet show page, and the count can be clicked to take you to the tasks tab of the pets show page. Let's make the whole row clickable to take you to the tasks tab of the pets show page, instead.

sarvaiyanidhi:
Hi @kasugaijin,

With regards to this issue, I just wanted to highlight that tasks link is already rendered to URL http://localhost:3000/alta/staff/pets/1?active_tab=tasks when we click on count of tasks displayed in table. Just highlighting it as I had worked on this feature earlier. 

Right now in current implements, when we click on pet name it goes to show page and on click of count of overdue or incomplete tasks it goes to pets show page with param active_tab=tasks

Thanks

kasugaijin:
@sarvaiyanidhi  ah I see! Then it would be good to include the URL parameter on all clicks from that table so it always takes the user to the pet tasks tab

sarvaiyanidhi:
Sure.. Will make changes and push it.

PR #966, opened 2024-09-09
951 - Dashboard Task Row Clickable

sarvaiyanidhi:
# 🔗 Issue
#951 

# ✍️ Description
- Code changes to make entire row clickable for result displayed for overdue and incomplete tasks

# 📷 Screenshots/Demos
NA


app/views/organizations/staff/dashboard/_pets_with_incomplete_or_overdue_tasks.html.erb
```diff
@@ -11,13 +11,13 @@
       </thead>
       <tbody>
         <% @pets.each do |pet| %>
-          <tr>
+          <tr data-controller="click-row" data-click-row-url-value="<%= staff_pet_path(pet, active_tab: 'tasks') %>" class="cursor-pointer">
```
kasugaijin, line 14: Do we need to use JavaScript for this? Can't we just wrap the row in a `link_to` helper that has params for the `pet.id` and the `active_tab`?

sarvaiyanidhi: I don't think we can wrap link_to directly with tr tag so went for this approach. I still checked again and implemented it but it does not work. Let me know if I am missing something and if there is simple solution to this.

kasugaijin: @sarvaiyanidhi ah I did not know that! Today I learned...
In that case, can we wrap the pet name, sex and count in the same link_to tag? I think it's a much simpler solution than adding a new JS controller.

sarvaiyanidhi: Sure, adding link_to to each value of column will be straight forward.. Will make changes and push code.

kasugaijin (approved):
Thank you @sarvaiyanidhi

Issue #970, opened 2024-09-11
Staff Pets: Add badges so staff can tell if a pet has been adopter, is in foster, or open


kasugaijin:
We have a badge component that already tells people if an adoption is pending on a pet. 
![Image](https://github.com/user-attachments/assets/05e3d702-128a-4716-bd80-d9b25caec9b6)

Use this badge component to create three new badges for 
- adopted (`bg-light-success`)
- in foster (`text-white bg-dark-primary`)
- open (`bg-light-warning`)

Add this badge to pets index table/mobile card for staff - replace the sex column with a 'status' column and place the badge.
![Image](https://github.com/user-attachments/assets/42df6ed5-e7f1-47fe-a8bc-362da4704a5a)
 
Add this badge to the pets show page next to the pet name
![Image](https://github.com/user-attachments/assets/9216e59e-3dfc-4243-92d9-b60a4725d419)

Given that this view logic will be used in more than one place, it probably makes sense to use a Helper, something like `PetBadgeHelper` or the likes

italomatos:
Hi @kasugaijin ! can I do this one?

kasugaijin:
For sure! @italomatos

PR #975, opened 2024-09-14
add pet status badges

italomatos:
# 🔗 Issue
https://github.com/rubyforgood/pet-rescue/issues/970

# ✍️ Description
This PR aims to include a badge according to the pet’s status

# 📷 Screenshots/Demos
image

### DIFF
--- a/app/helpers/pets_helper.rb
+++ b/app/helpers/pets_helper.rb
@@ -4,8 +4,10 @@ module PetsHelper
       render BadgeComponent.new("Adoption Pending", class: "badge bg-info")
     elsif pet.is_adopted?
       render BadgeComponent.new("Adopted", class: "badge bg-warning")
+    elsif pet.in_foster?
+      render BadgeComponent.new("In Foster", class: "badge text-white bg-dark-primary")
     elsif pet.open?
-      render BadgeComponent.new("Open", class: "badge bg-light-warning")
+      render BadgeComponent.new("Open", class: "badge bg-light-warning text-dark")
     else
       ""
     end
--- a/app/helpers/pets_helper.rb
+++ b/app/helpers/pets_helper.rb
@@ -1,2 +1,13 @@
 module PetsHelper
+  def status_badge(pet)
+    if pet.has_adoption_pending?
+      render BadgeComponent.new("Adoption Pending", class: "badge bg-info")
+    elsif pet.is_adopted?
+      render BadgeComponent.new("Adopted", class: "badge bg-warning")
+    elsif pet.open?
+      render BadgeComponent.new("Open", class: "badge bg-light-warning")
+    else
+      ""
+    end
+  end
 end
--- a/app/models/match.rb
+++ b/app/models/match.rb
@@ -46,6 +46,8 @@ class Match < ApplicationRecord
   scope :adoptions, -> { where(match_type: :adoption) }
   scope :fosters, -> { where(match_type: :foster) }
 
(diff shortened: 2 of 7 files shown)


app/helpers/pets_helper.rb
```diff
@@ -1,2 +1,11 @@
 module PetsHelper
+  def status_badge(pet)
+    if pet.has_adoption_pending?
+      BadgeComponent.new('Adoption Pending', class: 'badge bg-info')
+    elsif pet.is_adopted?
+      BadgeComponent.new('Adopted', class: 'badge bg-warning')
+    else
+      BadgeComponent.new('Free', class: 'badge bg-light-warning')
```
italomatos, line 8: I think the bg-light-warning class for the open status isn’t very readable.

image

app/models/match.rb
```diff
@@ -46,6 +46,8 @@ class Match < ApplicationRecord
   scope :adoptions, -> { where(match_type: :adoption) }
   scope :fosters, -> { where(match_type: :foster) }
 
+  scope :in_foster, -> { fosters.where('? between start_date and end_date', Time.zone.now) }
```
italomatos, line 49: I still have some doubts about whether it’s correct to say that if a pet’s match is of the foster type and the current date is between the start date and end date, it is considered in_foster?

app/helpers/pets_helper.rb
kasugaijin: Yes you are right. This is not accessible. Perhaps we could have black text on the background? Does that look better?

app/views/organizations/adoptable_pets/index.html.erb
```diff
@@ -60,8 +60,7 @@
             image_options: {src: pet.images.first, url: adoptable_pet_path(pet)}
           ) do |c| %>
             <%= c.with_badge do %>
-              <%= render BadgeComponent.new("Adoption Pending", class: "badge bg-info", when: pet.has_adoption_pending?) %>
-              <%= render BadgeComponent.new("Adopted", class: "badge bg-warning", when: pet.is_adopted?) %>
+              <%= status_badge(pet) %>
```
kasugaijin, line 63: So, for this particular view, we only need to show the adoption pending badge for a pet. Otherwise we don't need to show a badge for in foster or open.

app/views/organizations/staff/pets/show.html.erb
```diff
@@ -1,7 +1,7 @@
 <%= render DashboardPageComponent.new(crumb: :dashboard_pet, crumb_options: [@pet]) do |c| %>
   <% c.with_header_title do %>
```
kasugaijin, line 2: Can you please align the new badge on the center?
image

app/models/match.rb
kasugaijin: I think `in_foster` makes sense. 
Are you waiting on feedback on this before adding this to the helper method? I currently note that pets currently in foster do not have a badge showing, yet.

kasugaijin:
@italomatos looks good so far! Let me know if you have more questions.

app/models/match.rb
italomatos: @kasugaijin Yes, I just made the adjustment :) 

image

italomatos:
> @italomatos looks good so far! Let me know if you have more questions.

Thank you for the reviewing!  @kasugaijin 🚀

kasugaijin (approved):
LGTM!

Issue #969, opened 2024-09-11
Staff Dashboard - Fix pet filter for status


kasugaijin:
Log in as staff and go to the dashboard > pets and you will see an index of pets. Add a filter e.g. `unadopted` and you will note that in the results there will still be adopted pets. This is incorrect. Please fix this and check the other filters work as expected.

khiga8:
Can I pick this up?

kasugaijin:
@khiga8 for sure

PR #982, opened 2024-09-18
Fix Pet filter bug in staff view

khiga8:
# 🔗 Issue

https://github.com/rubyforgood/pet-rescue/issues/969

# ✍️ Description

This PR addresses a bug in the search filter resulting in the `unadopted` filter to not work as expected.

Changes made:
* Defines a missing `unadopted` scope.
* Update the `ransack_adopted` method to use strings instead of booleans to get around a known bug. Kudos to @coalest for pointing me to https://github.com/activerecord-hackery/ransack/issues/1375#issuecomment-1674885632.
* Add test coverage for `ransack_adopted` to ensure the adopted/unadopted filters work.
* Add missing test coverage for other filters in the search.

### DIFF
--- a/app/models/pet.rb
+++ b/app/models/pet.rb
@@ -132,7 +132,7 @@ class Pet < ApplicationRecord
   end
 
   def self.ransack_adopted(boolean)
-    boolean ? adopted : unadopted
+    boolean ? adopted : Pet.where.not(id: Pet.adopted)
   end
 
   def self.ransack_birth_date(date)
--- a/app/models/pet.rb
+++ b/app/models/pet.rb
@@ -76,6 +76,7 @@ class Pet < ApplicationRecord
   ]
 
   scope :adopted, -> { joins(:matches).merge(Match.adoptions) }
+  scope :unadopted, -> { where.not(id: Pet.adopted) }
   scope :unmatched, -> { includes(:matches).where(matches: {id: nil}) }
   scope :published, -> { where(published: true) }
   scope :fostered, -> { joins(:matches).merge(Match.fosters) }
@@ -131,13 +132,9 @@ class Pet < ApplicationRecord
     [:ransack_adopted, :ransack_birth_date]
   end
 
-  # Using string values to get around a known ransack bug: https://github.com/activerecord-hackery/ransack/issues/1375
+  # Using string values to get around ransack bug: https://github.com/activerecord-hackery/ransack/issues/1375
   def self.ransack_adopted(adoption_state)
-    if adoption_state == "adopted"
-      adopted
-    else
-      Pet.where.not(id: Pet.adopted)
-    end
+    adoption_state == "adopted" ? adopted : unadopted
   end
 
   def self.ransack_birth_date(date)
--- a/app/models/pet.rb
+++ b/app/models/pet.rb
@@ -134,7 +134,7 @@ class Pet < ApplicationRecord
 
   # Using string values to get around ransack bug: https://github.com/activerecord-hackery/ransack/issues/1375
   def self.ransack_adopted(adoption_state)
-    adoption_state == "adopted" ? adopted : unadopted
+    (adoption_state == "adopted") ? adopted : unadopted
   end
 
   def self.ransack_birth_date(date)
--- a/app/models/pet.rb
+++ b/app/models/pet.rb
@@ -131,8 +131,13 @@ class Pet < ApplicationRecord
     [:ransack_adopted, :ransack_birth_date]
   end
 
-  def self.ransack_adopted(boolean)
-    boolean ? adopted : Pet.where.not(id: Pet.adopted)
+  # Using string values to get around a known ransack bug: https://github.com/activerecord-hackery/ransack/issues/1375
+  def self.ransack_adopted(adoption_state)
+    if adoption_state == "adopted"
+      adopted
+    else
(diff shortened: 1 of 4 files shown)


test/models/pet_test.rb
```diff
@@ -126,4 +126,38 @@ class PetTest < ActiveSupport::TestCase
...
+    should "should return adopted pets via Pet.ransack" do
+      adopted_pets = Pet.ransack({"ransack_adopted"=>"true"})
+
+      assert_equal Pet.ransack_adopted(true).to_sql, adopted_pets.result.to_sql
+      assert_equal 1, adopted_pets.result.count
+    end
+
+    # Fails. SQL isn't what we'd expect?
+    should "should return unadopted pets via Pet.ransack" do
+      unadopted_pets = Pet.ransack({"ransack_adopted"=>"false"})
```
khiga8, line 157: This isn't returning the SQL we'd expect. In other words, it's not calling the else condition of `def self.ransack_adopted(boolean)`. 

Perhaps class methods with conditional logic isn't supported by `ransackable_scopes` 🤔

kasugaijin: @khiga8 sorry, I did not get notified when you made this comment...only just saw it. 

It's currently a bit weird. The method returns a scope depending on the argument, which is OK. But, we have an `adopted` scope defined on `Pet`, but I do not see one defiend for `unadopted`. Can you try creating an `unadopted` scope and see if this filter behaves as expected?  For the unadopted scope, we will want to find Pets that do not have a Match with `type` `adoption`. All other pets should be included in the results (including those with a Match `type` of `foster`.
 
```
  def self.ransack_adopted(boolean)
    boolean ? adopted : unadopted # there is no unadopted scope on the Pet model
  end
```

kasugaijin:
@all-contributors
please add @khiga8 for code.

app/models/pet.rb
```diff
@@ -132,7 +132,7 @@ def self.ransackable_scopes(auth_object = nil)
   end
 
   def self.ransack_adopted(boolean)
-    boolean ? adopted : unadopted
+    boolean ? adopted : Pet.where.not(id: Pet.adopted)
```
kasugaijin, line 135: Sorry, I see you have already done this! Can we make this a scope called `unadopted` instead?

test/models/pet_test.rb
kasugaijin: Ok I get what you're saying now I am caught up. So, we have asserted the expected behaviour of this filter in this file, right? `test/controllers/organizations/staff/pets_controller_test.rb` 

And you have tested this works as expected in the browser?

If so, I am comfortable removing these tests. I don't think we need to assert the SQL. Thoughts?

app/models/pet.rb
khiga8: I can do that once I get the tests passing! Still having problems with this. 😅

test/models/pet_test.rb
kasugaijin: I can check out your branch this week to see but before I can get to that can you check the pet object you create in the setup to make sure it belongs to the expected organization

kasugaijin:
@khiga8 how's this going? Do you have any questions?

test/models/pet_test.rb
khiga8: Hey @kasugaijin. The Pet object within my test all belongs to the same organization.

I've updated the tests so it's clearer what cases are failing. 

The following scenarios work as expected (as shown by the passing tests):
* `Pet.ransack_adopted(true)`
* `Pet.ransack_adopted(false)`
* `Pet.ransack({ "ransack_adopted" => "true" })

However, the following scenario fails:
* `Pet.ransack({ "ransack_adopted" => "false" })

Curious if you might know why  `Pet.ransack_adopted(false)` returns expected results, but `Pet.ransack({ "ransack_adopted" => "false" })` doesn't.

coalest: @khiga8 I think you are running into an open issue/bug with the ransack gem. This workaround might help: https://github.com/activerecord-hackery/ransack/issues/1375#issuecomment-1674885632

kasugaijin:
@all-contributors
please add @coalest for review.

test/models/pet_test.rb
khiga8: Amazing @coalest, thank you for pointing me to that 😁

app/models/pet.rb
khiga8: Done :)

khiga8:
Hi! This is now ready for a review, thank you!

app/models/pet.rb
```diff
@@ -131,8 +132,9 @@ def self.ransackable_scopes(auth_object = nil)
     [:ransack_adopted, :ransack_birth_date]
   end
 
-  def self.ransack_adopted(boolean)
-    boolean ? adopted : unadopted
+  # Using string values to get around ransack bug: https://github.com/activerecord-hackery/ransack/issues/1375
```
kasugaijin, line 135: Nice, thank you for leaving a comment on this.

kasugaijin (approved):
LGTM. Thankyou for getting to the bottom of this one!

kasugaijin:
@khiga8 could just please run `rails standard:fix` to lint your changes - then it should pass the pipeline.

kasugaijin (approved):

Issue #925, opened 2024-08-28
Refactor adopter flow


kasugaijin:
- org's third party form in an i-frame
- new registration - create form submission with status awaiting_data
- point new adopters to the iframe form
- allow new adopters to make adopter applications (with or without filling out the form)

PR #997, opened 2024-09-22
925 refactor adopter flow

MooseCowBear:
# 🔗 Issue

#925

# ✍️ Description

Adds the external form flow for new adopter accounts

# 📷 Screenshots/Demos


app/controllers/organizations/adopter_fosterer/external_form_controller.rb
```diff
@@ -0,0 +1,31 @@
+module Organizations
+  module AdopterFosterer
+    class ExternalFormController < ApplicationController
+      skip_verify_authorized # MARK: what should the auth policy look like for this action?
```
kasugaijin, line 4: I think we just need to check the user is authenticated and an adopter?

app/controllers/organizations/adopter_fosterer/external_form_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      private
+
+      def form_layout
+        return "adopter_foster_dashboard" if params[:dashboard]
+
+        "application"
+      end
+
+      def form_instructions
```
kasugaijin, line 22: I think it might cleaner to set an instance variable here which you can use to show the correct text in the view, rather than define the text in the controller. The text is really view layer logic. So, in the controller you can do something like 

```
@dashboard = params[:dashboard].present?
```

and in the view
```
if @dashboard
this text
else
that text
end
```

app/controllers/registrations_controller.rb
```diff
@@ -11,10 +11,12 @@ def new
     respond_with resource
   end
 
+  # MARK: only adopters are created through this route. Adopters need both the adoper role and a form submission to attach their applications to
   def create
     super do |resource|
       if resource.persisted?
         resource.add_role(:adopter, Current.organization)
+        FormSubmission.create!(person_id: resource.person_id)
```
kasugaijin, line 19: Can we call the associations here?
e.g. `resouce.person.create_form_submission`

kasugaijin (commented):
Looking good! A few small suggestions

I also note that currently

- we still have `has_one :form_submission` on `Person` - I cannot remember if we discussed and said we would do this separately?
- if, as an adopter, I apply on a pet, the updated frame says my application is awaiting review. Is this not supposed to say awaiting data?

MooseCowBear:
@kasugaijin I don't remember what we decided about the move to from has_one to has_many, but I can add that to this.

I thought for the adopter shouldn't see the distinction between under review and awaiting data and that was going to be a distinction for the staff side since they are the ones that have to act when there are applications awaiting data. Does it help the adopter to know that their application is in the awaiting data stage?

kasugaijin:
> @kasugaijin I don't remember what we decided about the move to from has_one to has_many, but I can add that to this.
> 
> I thought for the adopter shouldn't see the distinction between under review and awaiting data and that was going to be a distinction for the staff side since they are the ones that have to act when there are applications awaiting data. Does it help the adopter to know that their application is in the awaiting data stage?

Yes please, may as well update the association here!

You’re right about the status for the adopter. Let’s keep as is and solve that problem if it actually becomes a problem in production.

test/system/adoption_fosterer_test.rb
```diff
@@ -3,10 +3,9 @@
 class AdoptionFostererTest < ApplicationSystemTestCase
   setup do
     @user = create(:adopter_fosterer)
-    @custom_page = create(:custom_page, :with_about_us_image)
 
     @pet = create(:pet)
-    create(:adopter_application, pet: @pet, form_submission: @user.person.form_submission || @user.person.create_form_submission)
+    create(:adopter_application, pet: @pet, form_submission: @user.person.form_submissions.create)
```
kasugaijin, line 8: I think rails gives us a dynamic method `create_form_submission`

app/controllers/registrations_controller.rb
```diff
@@ -16,7 +16,7 @@ def create
     super do |resource|
       if resource.persisted?
         resource.add_role(:adopter, Current.organization)
-        FormSubmission.create!(person_id: resource.person_id)
+        resource.person.form_submissions.create
```
kasugaijin, line 19: Same here https://github.com/rubyforgood/pet-rescue/pull/997/commits/d948471b7f2cafe885a10c0523b8932bb0e40dab#r1781695089

MooseCowBear: We can't use that with has_many though.

kasugaijin: Oh really? Ok I will have to look into that tomorrow when I am at my computer. I thought we could use that helper but I may be wrong.

kasugaijin: Ah yes, I tested this in the console and you're right.

kasugaijin:
@MooseCowBear hey Alisa, is this almost done now? I think it was just the one question I had around the create method?

MooseCowBear:
> @MooseCowBear hey Alisa, is this almost done now? I think it was just the one question I had around the create method? I do not want to rush you at all! Just trying to plan work and this is a key piece.

@kasugaijin I think I changed them all to use the association, so I think it should be ready? Let me know if there's anything else you see. I have time tomorrow evening I can make any changes then.

kasugaijin:
@MooseCowBear ok great I will take a look. Looks like there’s just some conflicts to resolve

db/migrate/20240902115121_remove_pet_adopter_foster_account_index_from_adopter_applications.rb
```diff
@@ -0,0 +1,5 @@
+class RemovePetAdopterFosterAccountIndexFromAdopterApplications < ActiveRecord::Migration[7.1]
```
kasugaijin, line 1: Do you think we should replace this index with form_submission_id instead of completely removing it?

db/migrate/20240902111455_change_adopter_foster_account_null_on_adopter_applications.rb
```diff
@@ -0,0 +1,6 @@
+class ChangeAdopterFosterAccountNullOnAdopterApplications < ActiveRecord::Migration[7.1]
```
kasugaijin, line 1: Do you think we should replace this index with form_submission_id instead of completely removing it?

kasugaijin (commented):
Looks great @MooseCowBear 
Just one question on the indexes.

test/factories/users.rb
```diff
@@ -17,6 +17,10 @@
       end
```
kasugaijin, line 17: Should we also update the `adopter` factory to create the associated `form_submission` after we add the role? See line 30 of this file.

test/factories/users.rb
```diff
@@ -17,6 +17,10 @@
...
     end
 
+    trait :with_person do
```
kasugaijin, line 20: And should we move this to an association in the main block of the User factory? I think we always want a Person when we create a User, right?

MooseCowBear: There's a test to check the creation of a person for a user that needs a user without a person, if we don't want that test then we can add person to the main block.

test/factories/users.rb
MooseCowBear: Yeah we can do that

db/migrate/20240902111455_change_adopter_foster_account_null_on_adopter_applications.rb
MooseCowBear: @kasugaijin Doesn't this index exist already? https://github.com/rubyforgood/pet-rescue/blob/a86d748b5462c6ebd8084e3a59390e3da05b4f8a/db/schema.rb#L54

test/factories/users.rb
kasugaijin: I think if we are always creating a User with a person then we should update the factories accordingly.

db/migrate/20240902111455_change_adopter_foster_account_null_on_adopter_applications.rb
kasugaijin: Ah you are right! Was added in Ken's PR. Thanks for that.

kasugaijin:
@MooseCowBear just a conflict to resolve!

kasugaijin (approved):
LGTM!

Issue #1024, opened 2024-10-02
AdopterApplication: Add unique constraint on model and database


kasugaijin:
We should prevent a given adopter from being able to make more than one adopter application on a given pet. 

So, please add a validation on `AdopterApplication` that will check for uniqueness on `form_submission_id` and `pet_id` and prevent creation of a record if this validation fails.

Please also add a migration to create an index that will also enforce this on the database. You can wrap the migration changes in the safety assured block so it won't upset strong migrations gem (see other recent migrations).

coalest:
Can I take this task?

kasugaijin:
@coalest all yours

PR #1030, opened 2024-10-04
#1024: Add unique constraint and model validation to AdopterApplication

coalest:
# 🔗 Issue
#1024

# ✍️ Description
- [x] Added unique constraint on `pet_id` and `form_submission_id` in the `adopter_applications` table.
- [x] Added the above as a uniqueness validation on the `AdopterApplication` model.
- [x] Added model tests for the new validation.

### DIFF
--- a/test/models/adopter_application_test.rb
+++ b/test/models/adopter_application_test.rb
@@ -11,12 +11,18 @@ class AdopterApplicationTest < ActiveSupport::TestCase
     should belong_to(:form_submission)
   end
 
+  context "validations" do
+    should validate_uniqueness_of(:pet_id).scoped_to(:form_submission_id)
+     .with_message("Only one application per pet per person is allowed")
+  end
+
   context "self.retire_applications" do
     context "when some applications match pet_id and some do not" do
       setup do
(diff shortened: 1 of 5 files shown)


test/models/adopter_application_test.rb
```diff
@@ -11,12 +11,18 @@ class AdopterApplicationTest < ActiveSupport::TestCase
...
   context "self.retire_applications" do
     context "when some applications match pet_id and some do not" do
       setup do
-        @selected_applications = Array.new(3) {
-          create(:adopter_application, pet_id: @application.pet_id, form_submission: @form_submission)
-        }
+        @form_submissions = create_list(:form_submission, 3)
+        @selected_applications = 3.times.map do |i|
+          create(:adopter_application, pet_id: @application.pet_id, form_submission: @form_submissions[i])
+        end
```
coalest, line 25: I had to do something here, because we were creating three adopter applications with the same `pet_id` and `form_submission_id`, so it was failing the newly created validation.

kasugaijin:
@coalest please run `rails standard:fix` to pass the linter

Issue #1009, opened 2024-09-25
App Root Page: when I click the CTA buttons I am directed to a new organization request form


kasugaijin:
The CTA buttons on the app root page should take the user to a form for organizations to fill out so they can request access to the app (see how we implemented the Feedback form available in the footer and on dashboards). When the user submits this form, it should send an email to the dev email account (TDB) with a title **New Organization Request** and provide the details from the form, below. 

This email will inform the devs that they need to vet the organization, discuss their preferred slug, and then run the service to create a new organization.

The form should collect this information:
- Organization Name
- Your name (the person filling out the form) (we don't need to separate first and last, it's just for conversational purposes)
- Organization Phone Number
- Organization email
- Country
- City/Town
- Province State

Please also change the CTA button text to **Request an account**
Please also update the Org footer link to point to this new form
![Image](https://github.com/user-attachments/assets/d8ecdd00-269f-4de3-81eb-705243914e0d)

App root page CTA buttons x 3
![Image](https://github.com/user-attachments/assets/9ff08720-ac9e-40a4-8d53-bdc88433e5d2)
![Image](https://github.com/user-attachments/assets/1cdb363a-10e0-4990-abdc-6f25cb0c9fd7)

DreamVenerable:
I'm interested ✋🏻

kasugaijin:
@DreamVenerable all yours

kasugaijin:
@DreamVenerable how's this going? Let me know if you have questions.

DreamVenerable:
@kasugaijin Honestly I've been a bit busy, hence the lack of progress. 

If I find myself stuck or unable to complete it I'll inform you.

DreamVenerable:
@kasugaijin Should I create a new model or handle it from the controller?

kasugaijin:
@DreamVenerable please add a new controller - see `app/controllers/contacts_controller.rb` for an example.  We don't need a model as the data will be sent to email. 

I think we can call it something like `app/controllers/organization_acount_requests_controller.rb` unless you can think of a better name. Please also add the route `organization_account_request`, and view for this.

I also note that this will be the third contact form we are adding to the `app/controller` scope. I think there is a good argument to add a module like `contact_forms` in the controller and views so we can keep these controllers and views together. We can move all three controllers/views to this module. Let me know if you want to tackle that also...or we can put in a new issue. 

```
scope module: 'admin' do
  resources :articles, :comments
end
```

[...]

PR #1031, opened 2024-10-04
Organization Request - Issue #1009

DreamVenerable:
# 🔗 Issue

#1009 

# ✍️ Description

Allow organizations to request access to the app.

# 📷 Screenshots/Demos


app/helpers/organization_requests_helper.rb
kasugaijin, line 1: We can get rid of these empty files :)

Issue #1035, opened 2024-10-06
Bug fix: adopter application reviews changing the application status


kasugaijin:
See gif for demo of the bug - we are using a `<div>` with an id to identify what frame we want to replace but we cannot place a div inside a table. So it is replacing outside of the table. 
![bug](https://github.com/user-attachments/assets/15be5b4e-bb76-479f-9662-fd45e7223bee)

PR #1036, opened 2024-10-06
Fix adopter application review status change bug

kasugaijin:
# 🔗 Issue
https://github.com/rubyforgood/pet-rescue/issues/1035

# ✍️ Description
Cannot use a div as a child inside a table. So, the div with the ID to replace sits outside the table and the turbostream updates it...and does not replace the old row...see gif.

bug
bug

Fix
fix


app/views/organizations/staff/pets/tabs/partials/_application_info_table_row.html.erb
kasugaijin, line 1: I renamed this to include 'row' in the name because it is just a row, and not a complete table

MooseCowBear (approved):
lgtm

Issue #1025, opened 2024-10-02
View Translations: look in the adopter fosterer scope for any remaining text not using translation file


kasugaijin:
Please review the view files in the `Organizations::AdopterFosterer` scope to see if any are using plain strings in the view, and refactor these to use a value from `en.yml`. You may be able to leverage existing entries, and if you have to make new ones, please follow the existing structure in `en.yml`

wuletawwonte:
I would like to work on this

PR #1037, opened 2024-10-06
1025 make sure translations used in the adopter fosterer scope

wuletawwonte:
# 🔗 Issue
#1025 
# ✍️ Description
- Plain strings were replaced in the adopter and fosterer scope with localized strings.
- Added new translation keys where necessary, ensuring consistency with the existing structure in the `en.yml` file.
# 📷 Screenshots/Demos


app/views/organizations/adopter_fosterer/adopted_pets/index.html.erb
```diff
@@ -1,5 +1,5 @@
 <%= render DashboardPageComponent.new do |c| %>
-  <% c.with_header_title { "Adopted Pets" } %>
+  <% c.with_header_title { t(".header_title") } %>
```
kasugaijin, line 2: Interesting…does Rails find the correctly scoped header title for this? I have not done it this way before.

app/views/organizations/adopter_fosterer/likes/index.html.erb
```diff
@@ -1,5 +1,5 @@
 <%= render DashboardPageComponent.new do |c| %>
-  <% c.with_header_title { t('organizations.adopter_fosterer.likes.index.liked_pets') } %>
+  <% c.with_header_title { t('.header_title') } %>
```
kasugaijin, line 2: Same question here

app/views/organizations/adopter_fosterer/adopted_pets/index.html.erb
wuletawwonte: Yes, Rails supports this, and it's called [Lazy Lookup](https://guides.rubyonrails.org/i18n.html#lazy-lookup). Rails automatically scopes the translation key based on the current controller and action, which is why I only need to use t(".header_title") here.

app/views/organizations/adopter_fosterer/likes/index.html.erb
wuletawwonte: https://github.com/rubyforgood/pet-rescue/pull/1037/#discussion_r1793249428

app/views/organizations/adopter_fosterer/adopted_pets/index.html.erb
kasugaijin: Very handy! Thanks for the explanation

kasugaijin (approved):
LGTM!

Issue #1033, opened 2024-10-04
External Form Upload: add ability for staff to import a CSV of user form submissions


kasugaijin:
Log in as staff and go to `/staff/external_form_upload` and you will see a place for organizations to upload data via a CSV from a third party form, e.g., google forms. 

To do
- add a note to the above description that currently we only support google form CSVs
- add a route and controller action (create) that will take the submitted CSV and execute the CSV import service. Note that we currently have `app/services/organizations/csvimportservice` but please add a new directory `importers` to `app/services/organizations` and rename the existing `csvimportservice` to `google_csv_import_service.rb` and put it in that new directory - flexible on naming if you can think of better!
- manually test things. You will need to get an example of a google form CSV
- write tests that ensure the flow works as expected e.g., incude a dummy CSV and make sure you test that it creates Form Answers for users with a matching email, and skips those where there is no email match.

Questons? Ask away...

kasugaijin:
@wandergithub this issue is related to what you just worked on if you are interested?

wandergithub:
Yes, I'm in.

wandergithub:
I have managed to accomplish this so far: 

CSV file:
![image](https://github.com/user-attachments/assets/b016732a-2f4c-499b-9441-7519c1d15899)

FormSubmission created: 
![image](https://github.com/user-attachments/assets/28c23bd5-bc6f-4a47-aa6f-64eb751119ed)
![image](https://github.com/user-attachments/assets/f6a7a5da-d52d-46c2-a1e1-b0ef5839fecf)


Currently, we can upload the FormSubmission with the respective FormAnswers and it seems fine under my limited understanding.

But a few doubts got my attention:
And I don't know if I'm supposed to address these now. For instance, my "email" and "timestamp" had to be manually named. This was not the default name when exported from Google Forms. 

What about nil files and user feedback? I guess we are not addressing these things here??

kasugaijin:
@wandergithub i got an email notification with some questions, but I think you may have deleted the comment? Let me know what questions you have!

PR #1038, opened 2024-10-06
1033 external form upload

wandergithub:
# 🔗 Issue

https://github.com/rubyforgood/pet-rescue/issues/1033
# ✍️ Description

- Add a note for Google form support to description.
- Add a route and create a method to use form upload service to process the csv file
- adds test

# 📷 Screenshots/Demos


wandergithub:
@kasugaijin There is an assertion that is commented. This is because it works perfectly when tested individually. But when using `./bin/rails test:all` it fails. I don't know why the FormSubmission person_id is different from the user_id when tested with the mentioned command but it matches as it should when tested individually. I could not find where is it being affected.

test/controllers/organizations/staff/external_form_upload_controller_test.rb
```diff
@@ -0,0 +1,24 @@
+require "test_helper"
+
+class Organizations::Staff::ExternalFormUploadControllerTest < ActionDispatch::IntegrationTest
+  setup do
+    @organization = ActsAsTenant.current_tenant
+    @file = fixture_file_upload("google_form_sample.csv", "text/csv")
+    @params = {files: [@file]}
+    @user = create(:admin, email: "adopter1@alta.com")
```
kasugaijin, line 8: We should be creating an `adopter` here, instead of `admin`

There are also some changes happening in this PR https://github.com/rubyforgood/pet-rescue/pull/997
- a Person can have multiple form submissions
- a form submission will be created on a person when a new adopter signs up

This means the asserstions on lines 20 and 21 will become unnecessary. We'll want to just test two cases

- when a form submission does not have any form answers (and no CSV timestamp), assert that new form answers are created
- when a form submission does have a csv timestamp matching the timestamp in the CSV, no new Form Answers are created

I think the best thing to do, here, is wait for that PR to get merged, then finish these tests off. Thoughts?

app/controllers/organizations/staff/external_form_upload_controller.rb
```diff
@@ -7,6 +7,17 @@ def index
         authorize! :external_form_upload,
           context: {organization: Current.organization}
       end
+
+      def create
+        authorize! :external_form_upload,
+          context: {organization: Current.organization}
+        files = params.require(:files)
+        files.each do |file|
```
kasugaijin, line 15: I think we should only really allow 1 upload here...it doesn't make sense for more than one file. 

Given we are using `app/views/organizations/staff/shared/_attachment_form.html.erb` partial, we could make the value for the `multiple` attribute an argument that gets passed to the partial. So in this case, it would be `false` and in the other place the partial is used, on the pet, it would be `true`. Thoughts?

Then we can get rid of this loop here. 

Also, curious of the benefit of checking `file.is_a?(ActionDispatch::Http::UploadedFile)`

wandergithub: `file.is_a?(ActionDispatch::Http::UploadedFile)` It is because the received parameter contains other things like `""`. Maybe It will be unnecessary when using just one file.

I don't understand completely the use case due to a lack of background knowledge of the project. If you think a single file is the way to go, I will implement the suggested changes.

kasugaijin: Yes this import service is making the assumption there is only one third party form, and therefore one CSV to upload. It doesn’t currently handle the case where adopter data lives across multiple forms. So it would be good to only allow one file for now.

test/controllers/organizations/staff/external_form_upload_controller_test.rb
wandergithub: Ok, I understand lines 20 and 21 can be removed.

On the other assertions I'm verifying that FormSubmissions and FormAnswers are being created after the upload csv. Where we have 4 answers and questions/cells and  1 form was uploaded so one FormSubmission was created. 

As I understood we may need to add a test for when a csv file matches an existing csv_timestamp entry... Nothing should happen. But all of these are subject to change after PR #997

Then I will remove the unnecessary assertions, address controller comments, and wait for your instructions on the tests.

kasugaijin: @wandergithub FYI this PR has now been merged https://github.com/rubyforgood/pet-rescue/pull/997
So now person has many form submissions

kasugaijin: @wandergithub let me know if you have any blockers on this :)

wandergithub: @kasugaijin Yeah, I would like to know which requirements or specifications changed on this PR. It's not clear to me.

test/controllers/organizations/staff/external_form_upload_controller_test.rb
```diff
@@ -0,0 +1,21 @@
+require "test_helper"
+
+class Organizations::Staff::ExternalFormUploadControllerTest < ActionDispatch::IntegrationTest
```
kasugaijin, line 3: Can we nest the modules instead please?

test/controllers/organizations/staff/external_form_upload_controller_test.rb
```diff
@@ -0,0 +1,21 @@
...
+  setup do
```
kasugaijin, line 4: Ok these tests look good. 
Any reason we set up one adopter and one staff? We should ideally set up two adopters and then assert that one of them will get updated form answers, and the other one does not. Does that make sense?

wandergithub: Yes it does 👍🏽

wandergithub:
image
does not fail locally ??

kasugaijin:
Some thoughts
https://www.loom.com/share/68033246a92b43a98e33fd72e466a625?sid=337cc0d5-0237-4e35-b6a3-d6b3328d3f72

wandergithub:
Ok, I will address it here in this PR.

wandergithub:
I'm blocked, I cant make the factory for `@person1 = create(:adopter, email: "adopter1@alta.com")` to persist the person.
The person with the email "adoper1@..." is not being created. I've used `:with_person` to make changes to the factory but when debugging in the csv_service part of the code the person does not exist for some reason 🤯.

kasugaijin:
> I'm blocked, I cant make the factory for `@person1 = create(:adopter, email: "adopter1@alta.com")` to persist the person. The person with the email "adoper1@..." is not being created. I've used `:with_person` to make changes to the factory but when debugging in the csv_service part of the code the person does not exist for some reason 🤯.

Can you please push your latest changes up to this branch? I'd like to see the full picture before looking at why

kasugaijin:
@wandergithub 
I think I found the issues - see the video
I'd suggest refactoring the CSV import logic first, too. 

https://www.loom.com/share/d462da8a6a534cd0b75b831fcc593786?sid=4cf176c6-89f8-408d-8aa1-289c46fb3c7f

wandergithub:
@kasugaijin Thank you! I really needed that. I will implement the new service logic first to get it out of the way, and then check out everything else.

kasugaijin:
> @kasugaijin Thank you! I really needed that. I will implement the new service logic first to get it out of the way, and then check out everything else.

I am glad it helped!

kasugaijin:
I am just leaving a note for myself here - we should make sure we fully test the refactor of the CSV import service i.e.
- it skips the row if the user exists and the timestamp matches that on the FormSubmisson
- it updates the existing form submission if it is 'empty' i.e. does not have any answers (no timestamp saved)
- it creates a new form submission and adds the form answers if there is no 'empty' form submission and the timestamp is different

@wandergithub You could add the above tests to this PR...but it has already changed a lot for you, so it might be better to not change the scope of this PR any further, and then tackle any of the missing above tests separately to keep things manageable (if you prefer).

wandergithub:
> I am just leaving a note for myself here - we should make sure we fully test the refactor of the CSV import service i.e.
> 
> * it skips the row if the user exists and the timestamp matches that on the FormSubmisson
> * it updates the existing form submission if it is 'empty' i.e. does not have any answers (no timestamp saved)
> * it creates a new form submission and adds the form answers if there is no 'empty' form submission and the timestamp is different
> 
> @wandergithub You could add the above tests to this PR...but it has already changed a lot for you, so it might be better to not change the scope of this PR any further, and then tackle any of the missing above tests separately to keep things manageable (if you prefer).

I added the last two since the first test already exists as the first one in `csv_import_service_tests.rb`. I think I addressed all concerns . I will be attentive to any comments.

app/services/organizations/importers/google_csv_import_service.rb
kasugaijin, line 1: We duplicate the parsing and creating of each  `FormAnswer`
What do you think of extracting this to a method like 

We can move this whole block to a method
```
              row.each do |col|
                next if col[0] == "Email" || col[0] == "Timestamp"

                FormAnswer.create!(form_submission:,
                  question_snapshot: col[0], value: col[1])
              end
            end
```

Something like

```
def create_form_answers(form_submission, row)
              row.each do |col|
                next if col[0] == "Email" || col[0] == "Timestamp"

                FormAnswer.create!(form_submission:,
                  question_snapshot: col[0], value: col[1])
              end
            end
end
```

And where we set the form_submission on line 25, you can do that before calling the method and pass it in, and likewise below, create the new form_submission, and pass it to the method. 

Thoughts?

kasugaijin (commented):
Looking really good @wandergithub! Just one comment to DRY up the service object refactor.

app/services/organizations/importers/google_csv_import_service.rb
wandergithub: Ohh yeah, my bad 🙊

test/controllers/organizations/staff/external_form_upload_controller_test.rb
```diff
@@ -0,0 +1,28 @@
+require "test_helper"
+
+module Organizations
+  module Staff
+    class ExternalFormUploadControllerTest < ActionDispatch::IntegrationTest
+      setup do
+        file = fixture_file_upload("google_form_sample.csv", "text/csv")
+        @params = {files: [file]}
+        user = create(:admin)
```
kasugaijin, line 9: I should have mentioned this earlier but technically we are creating user objects here not person objects. This might confuse someone in the future.

test/services/organizations/csv_import_service_test.rb
```diff
@@ -4,14 +4,14 @@
 module Organizations
   class CsvImportServiceTest < ActiveSupport::TestCase
     setup do
-      person = create(:person)
+      person = create(:adopter)
```
kasugaijin, line 7: Same here re variable naming

kasugaijin (approved):
Approved but please change the variable names in the tests per my suggestion

kasugaijin (approved):
Nice!

wandergithub:
I think that would do it 👍🏽

Issue #1034, opened 2024-10-05
Replace StaffAccount


mononoken:
We replaced AdopterFosterAccount with the introduction of the Person model. We need to remove StaffAccount as well. The main difference will probably be the `deactivated_at` logic having to be moved. Most likely belongs in `User` with other authz stuff.

kasugaijin:
@mononoken I think it's safe to build off this PR https://github.com/rubyforgood/pet-rescue/pull/997 - it's pretty much done, I just have a question on idexes for Alisa.

mononoken:
Whoops that would have been a good idea! @kasugaijin 🫣

PR #1039, opened 2024-10-07
1034 replace staffaccount

mononoken:
# 🔗 Issue
#1034 

# ✍️ Description
This finishes up the Person refactorings from the conference, and it finishes the removal of the "accounts" objects. It changes all logic that was using StaffAccounts to use the authorization layer instead (which uses roles, permissions, and ActionPolicy).

I did a few refactorings to simplify logic while removing the account logic, but otherwise most behavior should be the same.

I did change:
- On the adoptable pets show page, for staff, instead of not showing anything at all in the adopt button section, I added a link to go to the staff pet page for that pet. Simple QOL change. See screenshot below.
- I made Person required for User now. I think all the User records should already have Person IDs so hopefully no problems will occur.

This may cause some conflicts with other open PRs like @MooseCowBear's  #997, so we may want to hold off merging this until the former PRs get merged.

# 📷 Screenshots/Demos
_Embedded link for staff on adoptable pet page to get to the pet's staff page_
image


app/controllers/organizations/staff/staff_controller.rb
```diff
@@ -1,43 +1,32 @@
...
-      format.turbo_stream { render "organizations/staff/staff/update" }
-    end
-  end
-
-  def update_activation
-    if @staff_account.deactivated_at
-      activate
-    else
-      deactivate
+      format.turbo_stream
```
coalest, line 21: Do we want to flash a notice with this turbostream as well? Like we do in the `UserRolesController`? Although not sure how best to separate the messages for activation vs deactivation.

coalest: Also I noticed that there is a lingering reference to `staff_accounts` in the `config/locales/en.yml` file?

coalest: Not sure if the flash messages were working before your changes, but when I check out your branch they don't seem to be working.

app/controllers/organizations/adopter_fosterer/adopter_applications_controller.rb
```diff
@@ -19,7 +19,7 @@ def create
         notice: t(".success", message: MessagesHelper.affirmations.sample)
 
       # mailer
-      @org_staff = User.organization_staff(@pet.organization_id)
+      @org_staff = User.staff
```
kasugaijin, line 22: Just for my sanity as I have not checked, the tenant is set in the mailer scope isn't it?

app/controllers/organizations/staff/staff_controller.rb
```diff
@@ -1,43 +1,32 @@
...
   end
 
-  def deactivate
-    @staff_account.deactivate
-    respond_to do |format|
-      format.html { redirect_to staff_staff_index_path, notice: t(".success") }
-      format.turbo_stream { render "organizations/staff/staff/update" }
+  def update_activation
+    if @staff.deactivated_at
+      @staff.activate
```
kasugaijin, line 14: Not directly related to your code changes, but when calling method to update a record I think it should be a bang `!` method i.e., `activate!` and `deactivate!`. Thoughts? We can address in a new issue to keep this PR on scope, if you prefer.

app/controllers/organizations/staff/staff_controller.rb
kasugaijin: Yeah good catch - adding the flash to `app/views/organizations/staff/staff/update_activation.turbo_stream.erb` should fix that

app/models/staff_account.rb
kasugaijin, line 1: glorious

app/models/user.rb
```diff
@@ -3,6 +3,7 @@
 # Table name: users
 #
 #  id                     :bigint           not null, primary key
+#  deactivated_at         :datetime
```
kasugaijin, line 6: this will make it simpler for deactivating adopters/fosterers also

app/models/user.rb
```diff
@@ -130,6 +125,18 @@ def name_initials
     full_name.split.map { |part| part[0] }.join.upcase
   end
 
+  def deactivate
```
kasugaijin, line 128: Oh these are your methods :D
Can we use ! on these given they are writing to the db?

app/policies/application_policy.rb
```diff
@@ -34,8 +34,10 @@ def verify_organization!
   end
 
   def verify_active_staff!
-    deny! unless user.staff_account
-    deny! if user.staff_account.deactivated?
+    deny! unless user.roles.map(&:name).any? do |role|
+      role == "admin" || role == "super_admin"
+    end
```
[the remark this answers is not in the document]
kasugaijin: ^ yeah that might be cleaner and more usable in other cases

I think we could also use `include?` to make this more readable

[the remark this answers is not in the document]
kasugaijin: or can we use the Rolify `has_role` helper here?

app/views/organizations/adoptable_pets/show.html.erb
```diff
@@ -129,54 +129,51 @@
...
+                    <%= image_tag('pause.png', height: '21') %>
+                  <% elsif 'withdrawn' == @adoption_application.status %>
+                    <%= image_tag('red_no.png', height: '21') %>
+                  <% else %>
+                    <%= image_tag('green_check.png', height: '21') %>
+                  <% end %>
+                </div>
+              <% elsif allowed_to?(:manage?, @pet) %>
+                <%= link_to "Manage #{@pet.name}", staff_pet_path(@pet) %>
+              <%# FIXME: This section was missed during the remove of AdopterFosterProfile but I think it requires FormSubmission to be ready to refactor %>
```
kasugaijin, line 158: Can we create an issue for this if not done so already?

app/views/organizations/staff/staff/update_activation.turbo_stream.erb
```diff
@@ -0,0 +1,4 @@
+<%= turbo_stream.replace_all ".staff_deactivate_toggle_#{@staff.id}",
+  partial: "deactivate_toggle",
+  locals: {staff: @staff}
+%>
```
kasugaijin, line 4: add flash here?

db/schema.rb
```diff
@@ -286,7 +276,8 @@
     t.string "invited_by_type"
     t.bigint "invited_by_id"
     t.integer "invitations_count", default: 0
-    t.bigint "person_id"
+    t.bigint "person_id", null: false
+    t.datetime "deactivated_at"
     t.index ["email"], name: "index_users_on_email", unique: true
```
kasugaijin, line 281: do you think it's worth adding an index to users on person_id?

kasugaijin (commented):
Looking good @mononoken just a couple comments!

mononoken:
Thanks for the comments, y'all! I just worked through all the conflicts from the rebase and will get to your comments hopefully tomorrow.

app/controllers/organizations/staff/staff_controller.rb
mononoken: That's not the typical bang naming convention, and since that convention is fairly established think it is best we not mess with it.

mononoken: Ah, I see that you wanted to change these to use `update!` so understand this idea better.

To me, seeing a bang method indicates that there is a non-bang version. I don't think methods typically have bangs unless a safer version exists too.

app/models/user.rb
mononoken: See other comments. I'll update these to `update!` too

app/policies/application_policy.rb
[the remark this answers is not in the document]
mononoken: I like this idea and am working on it. Investigating multiple tests that started failing when moving this logic.

app/controllers/organizations/staff/staff_controller.rb
kasugaijin: Interesting, I had not thought of it that way. I learned it was to signal a dangerous method that will modify the object permanently, and also will fail loudly. It’s a way to let developers know it’s no ordinary instance method.

This thread is an interesting discussion on it that echoes that, but also some people mention what you state, including Matz by the looks of it. https://stackoverflow.com/questions/612189/why-are-exclamation-marks-used-in-ruby-methods

What’s a good way to let the developer know that calling a particular instance method will actually update the record, if we don’t use a bang? A bang is the convention used where I work…but I am open to ideals and not chained to it.

app/policies/application_policy.rb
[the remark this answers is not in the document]
mononoken: Finished refactoring for this at 123fbace8e91

app/controllers/organizations/staff/staff_controller.rb
mononoken: Created a ticket on the kanban for adding the flash, since it's not in the scope of this PR.

mononoken: Added flash a76573f9b53f

Removed staff_account refs in yml b68048665bf3

@kasugaijin I added the flash in the controller. Do we actually define flash messages in the view file anywhere in the app currently? I'm not sure of the benefit of doing so, but I think that should be a separate PR to refactor for the whole app if there is a reason.

app/controllers/organizations/staff/staff_controller.rb
mononoken: If Matz made a statement on it, I'm inclined to follow his guidance. There are plenty of mutating methods in Ruby that don't have bangs in them, as I see that thread also brings up. Could go into further debate, but ultimately this is dev preferences which are hard to change. We can discuss this more if we think it needs to be in the next meeting to come to a project consensus.

In regards to your question, I hope devs aren't just calling methods without knowing what they do, purely on naming conventions! I don't think Ruby is often written functionally, so I would expect devs to expect something called "activate" to mutate the object.

app/models/user.rb
mononoken: 81c23a7430a2 adds bangs to `update!`

app/views/organizations/adoptable_pets/show.html.erb
mononoken: Started a draft ticket on the kanban. This relies on FormSubmission CRUD being in place which I think it currently isn't since there is no controller for it.

mononoken:
Think I got all the comments addressed and resolved new merge conflicts. Lmk if I missed anything.

kasugaijin (approved):
Thanks @mononoken ! lgmt.

app/controllers/organizations/staff/staff_controller.rb
kasugaijin: To clarify...I agree that devs should not be calling methods wihtout knowing what they do haha...but the `!` is an instant communicator it's a mutating method. But, I am fine with this implementation as well.

app/controllers/organizations/staff/staff_controller.rb
kasugaijin: Yes it's better to stay consistent.

Issue #795, opened 2024-06-01
Filter pets by fostered on staff pet view


mononoken:
## Description

On the staff pets view, `/staff/pets`, we cannot filter pets by if they are fosterable or fostered. Currently, they only show "adopted" and "unadopted". 

These filters work with ransack, and they may benefit from being reworked from their current state.

I haven't fully thought through this. If you have suggestions, let's think about how this works and make a clearer idea of what these options should be and what their names on the UI should be.

The cases desired to show:
- Pets who can be adopted but are not
- Pets that are adopted
- Pets who can be fostered but are not
- Pets that are fostered

## Acceptance criteria

- [ ] Filter pets by all of the cases described

meg-gutshall:
## Ideas for Desired Filters

- Pets who can be adopted but are not --> **Adoptable**
- Pets that are adopted --> **Adoptions**
- Pets who can be fostered but are not --> **Fosterable**
- Pets that are currently or are schedule to be fostered --> **Fosters**

I would like to implement more granular filters like the following:

- Pets that are fostered --> **Current Fosters**
- Pets that were previously fostered --> **Completed Fosters**
- Pets that are scheduled to be fostered --> **Pending Fosters**

It feels like this would require a change in design for the `Match` model, or at the very least implementing a state machine. Personally, I think the `Match` model would be better off as a parent class to `Adoption` and `Foster` models. It can contain shared behavior and logic while the child classes handle more specific functionality.

This is mainly driven by the `foster_statuses` method on the `Match` model. There are more options that should be added to this list of statuses. There are also statuses that should be considered for adoptable pets too, which we can convert into filters.

meg-gutshall:
Thanks @kasugaijin for removing the assigned label. I got way too busy when I came home to finish this up.

For anyone who wants to pick this up, the WIP code is in branch `795-add-pet-filters`. The custom Ransackers aren't working at the moment (the case statement methods). I think it may be because the case statement is checking for an integer when the params would be passing a string to the method.

That would be a good place to pick back up on this issue. Ping me with any questions!

coalest:
I could take this issue on :raising_hand: 

Out of all the filters mentioned by @meg-gutshall, the only one that looks like it would be more effort to implement is the "Completed Fosters" filter (if this would return all pets that were fostered in their past, regardless of the their current status). This seems like it would require an additional table for recording this kind of adoption/fostering history.

[...]

PR #1045, opened 2024-10-08
Add NOT NULL constraint to application_paused field on pets

coalest:
# 🔗 Issue
There is no direct issue for this PR, but it is related to #795. 

# ✍️ Description
In the issue mentioned above, we want to add a filter to pets for their paused status, which is a boolean value with a default set to false. A minor issue is that there is no `NOT NULL` constraint on that column so technically a pet could have `NULL` for their paused status and therefore would be missing when filtering by true or false.

This PR assumes that a `NULL` value for `application_paused` doesn't make sense for a pet.

### DIFF
--- a/app/models/pet.rb
+++ b/app/models/pet.rb
@@ -3,7 +3,7 @@
 # Table name: pets
 #
 #  id                 :bigint           not null, primary key
-#  application_paused :boolean          default(FALSE)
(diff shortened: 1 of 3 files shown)


kasugaijin (approved):
Ah yes good catch!

PR #1046, opened 2024-10-08
795 Add additional filters on staff pet view

coalest:
# 🔗 Issue
#795 

# ✍️ Description
- Added filters for application_paused, published, and placement_type in the staff pets view.
- Changed text in filter from "Unadopted" to "Not adopted"
- Added controller tests
- Moved ransack logic on Pet model to concern

# 📷 Screenshots/Demos
Short demo of new filters: 
[screencast2.webm](https://github.com/user-attachments/assets/79d8ccaf-f490-417c-bb5a-7a9ba1e0896e)

On mobile, the filters are taking up a lot of screen real estate, so if mobile is important could be improved in the future.
Screenshot from 2024-10-08 14-06-26

# Questions
1. For the `placement_type` filter, I wasn't sure if this should just be a simple filter like I have implemented here, or if it should be something more. Because right now it seems confusing that you can have a pet that has already been adopted, but it will show up under "Adoptable" for Placement Type. Is this the expected behavior?

### DIFF
--- a/app/models/concerns/pet_ransackable.rb
+++ b/app/models/concerns/pet_ransackable.rb
@@ -1,0 +1,34 @@
+module PetRansackable
+  extend ActiveSupport::Concern
+
+  class_methods do
+    def ransackable_attributes(auth_object = nil)
+      %w[name sex species breed placement_type application_paused published]
+    end
+
+    def ransackable_associations(auth_object = nil)
+      ["adopter_applications"]
+    end
+
+    def ransackable_scopes(auth_object = nil)
+      [:ransack_adopted, :ransack_birth_date]
+    end
+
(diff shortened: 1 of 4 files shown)


kasugaijin:
@coalest awesome looks good! I’ll finalize my review once I get to play with it.

kasugaijin (approved):
Looks great thank you!
It made me realise we should update the seeds (done) to create pets with different placement types because they were all just fosterable

> For the placement_type filter, I wasn't sure if this should just be a simple filter like I have implemented here, or if it should be something more. Because right now it seems confusing that you can have a pet that has already been adopted, but it will show up under "Adoptable" for Placement Type. Is this the expected behavior?

Good question. I think for MVP this is fine. It's something we can improve if we get feedback saying it's not intuitive.

Issue #984, opened 2024-09-18
Feedback Form: Render it within both dashboard templates instead of redirecting


kasugaijin:
We have the Feedback link in the Staff Dashboard, and the Adopter Foster Dashboard, but this redirects to a new page. Instead, we should render this form within each dashboard layout. You'll probably want two new controllers to handle this. Look at the `Organizations::AdopterFosterer::FaqController` as an example.

Aaryanpal:
@kasugaijin this too

kasugaijin:
@Aaryanpal how's this going?

Aaryanpal:
WIP
@kasugaijin Sorrry for Late Response.
I was out for some work.
Resuming it Now

Aaryanpal:
Hi @kasugaijin,

I'd like to propose a refactoring of our current feedback feature implementation. Here's an overview of the current situation and my proposed changes:

Current Implementation:
- Controller name: `DevContactsController`
- Handles feedback features for both staff and adopters
- Current URL path: `http://localhost:3000/alta/feedback/new`

Proposed Changes:
1. Rename the controller to `FeedbackController` to better reflect its purpose and follow Rails naming conventions
2. Consolidate all feedback-related logic into this single controller
3. Maintain the existing URL structure (`/alta/feedback/new`)

Rationale:
- The current controller name (`dev_contacts_controller`) doesn't clearly communicate its purpose
- After reviewing the codebase, I don't see a compelling reason to maintain separate controllers for staff and adopter feedback

Please let me know your thoughts on this proposed refactoring. I'm happy to provide more detailed implementation plans if needed.

kasugaijin:
Hey @Aaryanpal I appreciate the initiative! I think what you propose makes sense. The ony thing to handle differently if we use a single controller is the layout. I recall the adopter/foster form should be in their dashboard layout.

Aaryanpal:
### For Staff it will look like this
![image](https://github.com/user-attachments/assets/bb734e2b-8cc1-418a-a7d4-a8af85c61882)

### For Adopter it look something like this
![image](https://github.com/user-attachments/assets/c984cc64-e6f7-4b55-b1dc-ffed4e3848d6)

Here's a rephrased version as a message to a dev:

---

[...]

PR #1051, opened 2024-10-09
984 - Feedback Form: Render it within both dashboard templates instead of redirecting 

Aaryanpal:
# 🔗 Issue
#984 

# ✍️ Description

- Added Feedback Controller and Feedback Mailer To Handle Feedback for staff and adopter
- Remove DevContactsController and DevContactsMailer
- Fix the test

# 📷 Screenshots/Demos

[Screencast from 2024-10-09 14-32-46.webm](https://github.com/user-attachments/assets/bf6a7659-dfe3-4b62-bdb4-91e956d92a7b)


kasugaijin:
@Aaryanpal nice looks good! I am pretty slammed but will try to review soon. Hopefully someone else can review as well in the meantime.

app/controllers/feedback_controller.rb
```diff
@@ -1,16 +1,16 @@
...
     @contact = Contact.new
   end
 
   def create
     @contact = Contact.new(contact_params)
-
     if @contact.valid?
-      DevContactsMailer.with(contact_params).send_message.deliver_later
+      FeedbackMailer.with(contact_params).send_message.deliver_later
       redirect_to root_path, notice: I18n.t("contacts.create.success")
```
jmilljr24, line 14:
```suggestion
      redirect_back_or_to root_path, notice: I18n.t("contacts.create.success")
```

jmilljr24 (approved):
Looks good!

I just added one thought/suggestion on the redirect after submission. For example, as an admin user, if felt odd to be redirected to the root page after submitting.  I would probably want to stay on the dashboard? 

Non blocker so feel free to merge if you are happy how it is!!

app/controllers/feedback_controller.rb
kasugaijin: Yes - redirect to dashboard would be preferred!

app/controllers/feedback_controller.rb
```diff
@@ -22,4 +22,14 @@ def create
   def contact_params
     params.require(:contact).permit(:name, :email, :message)
   end
+
+  def set_layout
+    if current_user.nil?
+      "application"
+    elsif current_user.staff_account
```
kasugaijin, line 29: we are deprecating the staff_account very soon. can you instead use the Rolify method `has_role?` and pass in the role you want to check, and the current organization, since roles are scoped to organizatoin on a user. 

so something like 
```
if current_user.has_role?(:admin, ActsAsTenant.current_tenant) ||
  current_user.has_role?(:super_admin, ActsAsTenant.current_tenant)
```

kasugaijin: It might be easier to check for the `adopter` role instead and set the fallback to the staff dashboard...less code

app/views/feedback/new.html.erb
```diff
@@ -0,0 +1,18 @@
...
+          <div class="row">
+            <div class="<%= current_user&.staff_account ? 'col-md-8 mx-auto card p-5 bigger' : 'col-md-12 mx-auto card p-5' %>">
+              <%= render 'form', contact: @contact %>
+            </div>
+          </div>
+        </div>
+      </section>
+    <% end %>
+  <% end %>
+<% end %>
```
kasugaijin, line 18: new line :)

kasugaijin (approved):
Just a couple small comments to address. Otherwise looks good.

kasugaijin:
@Aaryanpal bump - no pressure, just checking in to see if you have blockers?

Issue #885, opened 2024-07-21
Add tests for DefaultPetTask refactor


kasugaijin:
If you look in app/services/organizations we have the DefaultPetTaskService which gets called when we create a new pet, and adds a Task for each DefaultPetTask for that species.
If you look in the Organizations::Staff::PetsController #create you will see we call the service on the new pet

So, to test, we want to

    create the default pet task for dog
    create a dog
    verify that dog has a task with the same name as the default pet task for dogs
    do the same for cat

[PR](https://github.com/rubyforgood/pet-rescue/pull/879)

atbalaji:
Currently I'm not taking this @kasugaijin. I would like to revisit this incase no one else took this.

kasugaijin:
@atbalaji all yours when you want it...just assign yourself.

Aaryanpal:
@kasugaijin can u please assign this

kasugaijin:
@Aaryanpal no need to test front end that looks for DOM elements. Just backend model or controller level makes sense

Aaryanpal:
WIP

Aaryanpal:
WIP

Aaryanpal:
@kasugaijin I am not getting any clue for what use case i need to write the Test Cases.
Any Help Will be Highly Appreciated.

kasugaijin:
@Aaryanpal the original PR added species to the default pet task service, so it would be good to have some tests where we set task A on dogs and task B on cats and make sure they are assigned to cats and dogs when they are created

Aaryanpal:
Thanks for the Guidance @kasugaijin.
Raised the PR

jmilljr24:
@wiliajc87 So PR #879 was merged but we do not have test coverage for it. It added the ability for default pet tasks to be assigns to a general category or by species, ie cat, dog. A new pr for tests was started but was closed before being merged so nothing for testing is done yet.

If you go to the staff dashboard you should see the default pet tasks with the categories. If you create a new pet, the default pet tasks will be added as tasks to this newly created pet.

We want to test that if a default pet task has the category "dog", the newly created dog gets that task. A cat should not get a dog task. 

Let me know if that makes sense and if you would like to be assigned.

jmilljr24:
@wiliajc87 Awesome! Thank you. #1053 is the closed PR that wasn't merge. I might provide more context if you want to take a look.

wiliajc87:
[...]

PR #1053, opened 2024-10-09
885 - Add tests for DefaultPetTask refactor

Aaryanpal:
# 🔗 Issue
#885

# ✍️ Description
- Added Test For Species Dog and Cat Assignment for DefaultPetTaskController

### DIFF
--- a/test/controllers/organizations/staff/default_pet_tasks_controller_test.rb
+++ b/test/controllers/organizations/staff/default_pet_tasks_controller_test.rb
@@ -41,6 +41,34 @@ class Organizations::Staff::DefaultPetTasksControllerTest < ActionDispatch::Inte
           post staff_default_pet_tasks_url, params: @params
         end
       end
+
+      should "assign to dog" do
+        assert_authorized_to(
+          :manage?, DefaultPetTask,
+          context: {organization: @organization},
+          with: Organizations::DefaultPetTaskPolicy
+        ) do
+          post staff_default_pet_tasks_url, params: {
+            default_pet_task: {name: "New Dog Task", species: "Dog", due_in_days: 5}
+          }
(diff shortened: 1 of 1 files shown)


test/controllers/organizations/staff/default_pet_tasks_controller_test.rb
```diff
@@ -41,6 +41,34 @@ class Organizations::Staff::DefaultPetTasksControllerTest < ActionDispatch::Inte
...
+        assert_authorized_to(
+          :manage?, DefaultPetTask,
+          context: {organization: @organization},
+          with: Organizations::DefaultPetTaskPolicy
+        ) do
+          post staff_default_pet_tasks_url, params: {
+            default_pet_task: {name: "New Dog Task", species: "Dog", due_in_days: 5}
+          }
+          has_dog_species = DefaultPetTask.find_by_name("New Dog Task").Dog?
+          assert_equal true, has_dog_species
```
kasugaijin, line 55: Ah so this is testing that it creates a new DefaultPetTask - but what I meant is that it creates a new `Task`
If you look in app/services/organizations we have the `DefaultPetTaskService` which gets called when we create a new pet, and adds a Task for each DefaultPetTask for that species. 

So, to test, we want to 
- create the default pet task for dog
- create a dog
- verify that dog has a **task** with the same name as the default pet task for dogs
- do the same for cat

does that make sense?

kasugaijin: bump :)

jmilljr24:
@Aaryanpal I'm going to close this. If you are still interested, feel free to comment on the issue to be reassigned if no one beats you to it.

Issue #1021, opened 2024-10-02
Pet: Refactor instance method to use the Match model instead of the AdopterApplication model


kasugaijin:
We have this instance method on `Pet`

  def is_adopted?
    adopter_applications.any? { |app| app.status == "adoption_made" }
  end

The single source of truth on an adoption should be the `Match` model, where the `match_type` is `adoption`. Please update the method to query the pet's matches instead.

coalest:
@kasugaijin I can take this

PR #1055, opened 2024-10-10
1021 Refactor Pet instance method to use the Match model instead of the AdopterApplication model

coalest:
# 🔗 Issue
#1021 

# ✍️ Description
- Refactored `Pet#is_adopted?` method to reference `Match` association rather than `AdopterApplication`
- Fixed related tests

### DIFF
--- a/app/controllers/organizations/adoptable_pets_controller.rb
+++ b/app/controllers/organizations/adoptable_pets_controller.rb
@@ -8,10 +8,9 @@ class Organizations::AdoptablePetsController < Organizations::BaseController
   helper_method :get_animals
 
   def index
-    @q = authorized_scope(Pet.includes(:adopter_applications, images_attachments: :blob),
-      with: Organizations::AdoptablePetPolicy).ransack(params[:q])
+    @q = authorized_scope(Pet.all, with: Organizations::AdoptablePetPolicy).ransack(params[:q])
     @pagy, paginated_adoptable_pets = pagy(
-      @q.result,
+      @q.result.includes(:adopter_applications, :matches, images_attachments: :blob),
       limit: 9
     )
     @pets = paginated_adoptable_pets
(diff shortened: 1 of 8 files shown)


coalest:
I'm sure that performance isn't a priority at the moment, but I noticed while refactoring the `is_adopted?` method, that there is an N+1 on the `/staff/pets` page because of this method and the `in_foster?`/`open?` methods are called in order to fetch the status of every pet.

Before changes:
Screenshot from 2024-10-10 16-29-50

After changes:
Screenshot from 2024-10-10 16-30-58

You can see the number of Active Record queries goes down significantly. I can put this into a separate PR, or just stash these changes for now, if we don't want to look at performance yet.

app/controllers/organizations/adoptable_pets_controller.rb
```diff
@@ -8,10 +8,9 @@ class Organizations::AdoptablePetsController < Organizations::BaseController
   helper_method :get_animals
 
   def index
-    @q = authorized_scope(Pet.includes(:adopter_applications, images_attachments: :blob),
-      with: Organizations::AdoptablePetPolicy).ransack(params[:q])
+    @q = authorized_scope(Pet.all, with: Organizations::AdoptablePetPolicy).ransack(params[:q])
```
kasugaijin, line 11: Do you think we should only return pets that are unadopted for this collection?

kasugaijin (approved):
Nice one! Looks like you were N+1 hunting

app/controllers/organizations/adoptable_pets_controller.rb
coalest: Sure, I can do that. Okay if I add that to this PR?

coalest: Added to this PR for now (see last commit).

kasugaijin: Yes in this PR is fine. Thanks.

kasugaijin (approved):
LGTM!

Issue #1057, opened 2024-10-11
Staff: Dashboard index summary stats visual design improvement


kasugaijin:
This is not urgent, but if someone has an eye for design and wants to tackle this - please provide ideas below
We have these sumamry stats on the staff dashboard index for pet tasks, adoption applications, and fosters, but I do not like the large purple text. I think it could be refined, with a different colour and perhaps and icon is the clickable link (I am open to ideas).

![image](https://github.com/user-attachments/assets/282facce-2685-4ebc-84f4-bf1f5294db6d)

Please base any design ideas on our theme, which is found here (look around these theme pages and see if there are elements or designs we can use) https://geeksui.codescandy.com/

sarvaiyanidhi:
Hi @kasugaijin ,

Was just playing around trying to come up with some suggestion... Let me know if you like below option for stats

![staff_dashboard](https://github.com/user-attachments/assets/6cd0e916-dbce-40e2-95b5-746551aa2b20)

kasugaijin:
@sarvaiyanidhi thanks for taking a look! I definitely think that’s better. One opinion, I think we could make the two stats for fosters the same colours as for applications, purple and orange. Otherwise it feels like a lot of different colours. Thoughts ?

sarvaiyanidhi:
Sure, let me play around with couple of more options and post design for reference.

sarvaiyanidhi:
Few options let me know if you like any one of it

Option 1

![staff_dashboard_option_1](https://github.com/user-attachments/assets/ab8d8496-0031-42cc-a6ad-5629788e1c46)

Option 2

![staff_dashboard_option_2](https://github.com/user-attachments/assets/9fa924f4-17b2-4014-80f6-ec297f295ea3)


Option 3

![staff_dashboard_option_3](https://github.com/user-attachments/assets/0cbf41de-7e6a-479f-b6a0-40731a4b479a)

kasugaijin:
@sarvaiyanidhi I prefer option 2! Thank you for these

PR #1060, opened 2024-10-15
1057 - Staff: Dashboard summary stats visual design improvement

sarvaiyanidhi:
# 🔗 Issue
#1057 

# ✍️ Description
Better UI presentation for staff dashboard stats

# 📷 Screenshots/Demos

**Stats View**

staff-dashboard-stats

**Full Page View**

Alta-staff-dashboard


app/views/organizations/staff/dashboard/index.html.erb
```diff
@@ -1,75 +1,103 @@
...
+          </div>
+        </div>
       </div>
       <div class="col-xl-3 col-lg-6 col-md-12 col-12">
-        <%= render DataCardComponent.new do |card| %>
-          <% card.with_title { "Applications" } %>
-          <% card.with_icon_name { "file-text" } %>
-          <% card.with_value do %>
-            <div>
+        <div class="card h-100 shadow-sm border-0">
```
kasugaijin, line 32: Ah - this highlights a deficiency with our DataCardComponent. Ideally we should be able to pass it a `title` and a `body` so that we can use it with these new cards you have made. 

We use the DataCardBody component only in the staff and adopter foster dashboards, both of which will change, so there's no existing logic to handle.

It's your call - do you want to address that in this PR, or would you rather I make a new issue for it?

sarvaiyanidhi: Thank you @kasugaijin.

I updated code to use DataCardComponent for title and icon but could not customize color for title and icon as per earlier design.

Here how it looks now

staff-dashboard-stats

sarvaiyanidhi: Do you want to go ahead with this or should I update code for DataCardComponent to customize css for title and icon as well?

kasugaijin: I think it would be good to have some pre-set options for the title size e.g., classes for `:small` `:medium` `:large` and some colour options for the icon. I think we should keep the title colour as it is for now, so it's consistent across the app. What do you think?

sarvaiyanidhi: I implemented title size and icon color attribute but kept it to minimum. I did not added other classes like "icon-shape rounded-circle p-2" as it was not looking good with current presentation of icon on right side.

This is how it looks with current updates

staff-dashboard-stats

kasugaijin: That look great I think!

app/views/organizations/staff/dashboard/index.html.erb
```diff
@@ -1,75 +1,109 @@
...
       </div>
     </div>
-    <%= render 'pets_with_incomplete_or_overdue_tasks' %>
+
+    <div class="mt-5">
+      <%= render 'pets_with_incomplete_or_overdue_tasks' %>
+    </div>
   <% end %>
-<% end %>
+<% end %>
```
kasugaijin, line 109: New line :)

kasugaijin (approved):
Thank you for taking this on and improving the component! Looks a lot better.

app/views/organizations/staff/dashboard/index.html.erb
sarvaiyanidhi: Done. I think need to add some ERB Lint check as well to avoid such issue.

Issue #1069, opened 2024-10-19
Adoptable Pets Refactor - show collection by species


jmilljr24:
We should render collections based on species
We discussed separate routes per species but think a single controller with URL param that works with Ransack directly e.g., `q%5Bspecies%5D=1`
We'd need to update the UI to allow users to click on each species (e.g. dropdown on the Pets nav link)

To do:
We'd need to enable Orgs to define which species they have (e.g. some are dogs only)

PR #1070, opened 2024-10-19
#1069 Adoptable Pets Refactor - show collection by species 

jmilljr24:
#1069

# ✍️ Description

Creates link's to filtered collections based on species. Removes view with all species in one collections (can be updated simply based on user/org needs and feedback). 

A separate url param was added vs. integrating with ransack. This eliminates multiple issues. The url is human readable. This does not rely on updating the enum if changed in the future and is easier to understand for new developers contributing to the project. A single query is still initiated so no negative performance issues. It also simplifies clearing the user selected ransack search params without affecting the species specific "page".

Additional changes:
-update test (removes some potentially flaky tests)
-update to use nested modules
-remove legacy static home page
-add/update i18n (tried to update most on the pages I interacted with. May have missed some)

To do:
Filter buttons based on the species Org's have ( if no cats, only display button for dogs)

# 📷 Screenshots/Demos

[Uploading Screencast from 2024-10-19 17-56-07.webm…]()


app/controllers/organizations/adoptable_pets_controller.rb
```diff
@@ -1,48 +1,61 @@
...
-  end
+module Organizations
+  class AdoptablePetsController < Organizations::BaseController
+    include ::Pagy::Backend
+
+    skip_before_action :authenticate_user!
+    skip_verify_authorized only: %i[index]
+    before_action :set_likes, only: %i[index show],
+      if: -> { allowed_to?(:index?, Like) }
+    helper_method :get_animals
```
kasugaijin, line 9: An aside comment - I feel like this method should be called `get_breeds` instead

app/views/organizations/adoptable_pets/index.html.erb
```diff
@@ -26,22 +28,24 @@
...
                 <%= f.select :sex_eq, ["Male", "Female"], {include_blank: 'All'}, class: "form-select" %>
               </div>
-              <div class="form-group">
-                <%= f.label :species_eq, "Species" %>
-                <%= f.select :species_eq, Pet.species, {include_blank: 'All'}, class: "form-select" %>
-              </div>              
               <div class="form-group">
                 <%= f.label :breed_eq, "Breed" %>
-                <%= f.select :breed_eq, get_animals, {include_blank: 'All'}, class: "form-select" %>
+                <%= f.select :breed_eq,
```
kasugaijin, line 33: FYI the filter CSS is different between species. It would be good to make the same.
filters

app/controllers/organizations/adoptable_pets_controller.rb
```diff
@@ -1,48 +1,61 @@
...
+      end
     end
-  end
 
-  def set_likes
-    likes = current_user.person.likes
-    @likes_by_id = likes.index_by(&:pet_id)
+    private
+
+    def get_animals(species)
```
kasugaijin, line 51: If we have a before_action that sets `@species` we can get rid of this method parameter and also use the `@species` instance variable in the index action

app/views/organizations/adopter_fosterer/external_form/index.html.erb
```diff
@@ -3,9 +3,25 @@
...
+    <div class="d-flex justify-content-end gap-2 pb-2">
+      <% if user_signed_in? %>
+        <%= link_to " #{t("general.view")} #{t("general.dogs")}",
+        adoptable_pets_path(species: "dog"),
+        class: "btn btn-primary" %>
+        <%= link_to " #{t("general.view")} #{t("general.cats")}",
+        adoptable_pets_path(species: "cat"),
+        class: "btn btn-primary" %>
+      <% else %>
+        <%= link_to t("general.become_an_adopter"),
```
kasugaijin, line 20: Is it possible to reach this route without being authenticated?

kasugaijin (commented):
@jmilljr24 Looking good! I think this is a solid UX improvement. Just a few comments/thoughts.

jmilljr24:
> @jmilljr24 Looking good! I think this is a solid UX improvement. Just a few comments/thoughts.

@kasugaijin Awesome feedback! A couple points you referenced crossed my mind during development and I forgot to circle back to so thank you!!! I will review asap

app/views/organizations/adoptable_pets/index.html.erb
jmilljr24: I agree it seems abrupt switching between dogs/cats. I think(?) the solution would be to hard code the width of the longest named breed. The problem I see is that the max width of widest dog breed is different than the max width of the cat breeds. This would all be based off of our seed data if I were to set this.. What if there is a longer breed added to the DB than our seeds or a different species has an even longer breed than we can account for?

Going back to your point on most user's having a specific species in mind, I do not believe a user would be switching between different species and notice the change in spacing. I am hesitant to hardcore this but maybe I missed something?

app/views/organizations/adopter_fosterer/external_form/index.html.erb
jmilljr24: You are correct, this can be removed.

app/controllers/organizations/adoptable_pets_controller.rb
jmilljr24: I added a scope to the Pet model which I agree makes more sense.

kasugaijin (approved):
LGTM!

Issue #1068, opened 2024-10-19
Different weight units of pet in adoptable pets index vs. show


coalest:
I think I found a small bug. When viewing this pet on `/adoptable_pets`, I see the weight in kilograms. When I click on the pet, and go to the `/adoptable_pets/:id` show page I see a different unit for weight (pounds).

Note the highlighted text in the two screenshots below:

Index view:
![Screenshot from 2024-10-19 16-15-33](https://github.com/user-attachments/assets/b4d749a4-6985-43a8-88d1-827c1ca924f1)

Show view:
![Screenshot from 2024-10-19 16-15-26](https://github.com/user-attachments/assets/bcf2fd62-28b9-4c28-a801-c13c80c5c264)

jmilljr24:
@coalest Nice find! Feel free to investigate. Note there is an option for staff to edit/update the units of weight for a pet in pounds vs. kg. I haven't looked at what units we are saving in the DB. Any help on this is greatly appreciated!!!

PR #1074, opened 2024-10-21
Fix weight units bug in views

coalest:
# 🔗 Issue
#1068 

# ✍️ Description
Kilograms was hard coded in a few of our views, so I switched them to use the pet's `weight_units` instead.

# 📷 Screenshots/Demos


kasugaijin (approved):
Nice catch @coalest!

PR #1093, opened 2024-10-27
#1092 Fix file param for CSV upload

### LINKED ISSUE #1092: Error Uploading CSV
Screenshot from 2024-10-27 10-14-16

jmilljr24:
Resolves #1092 

# ✍️ Description

Fix error with file param and update tests.

Side note. We should add some sort of notice to the user after the file has been submitted. There is no redirect or notice after uploading.

# 📷 Screenshots/Demos

### DIFF
--- a/app/controllers/organizations/staff/external_form_upload_controller.rb
+++ b/app/controllers/organizations/staff/external_form_upload_controller.rb
@@ -11,7 +11,9 @@ module Organizations
       def create
         authorize! :external_form_upload,
           context: {organization: Current.organization}
-        file = params.require(:files).first
+        file = params[:files]
+
+        # Only processes single file upload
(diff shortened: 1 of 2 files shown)


kasugaijin (approved):

kasugaijin:
@jmilljr24 yes I agree. Ideally the `Organizations::Importers::GoogleCsvImportService` call method either returns true (success) or false (failure) and then in the create action on the controller we can handle accordingly with a flash message. 

Did you want to write an issue for this?

jmilljr24:
> @jmilljr24 yes I agree. Ideally the `Organizations::Importers::GoogleCsvImportService` call method either returns true (success) or false (failure) and then in the create action on the controller we can handle accordingly with a flash message.
> 
> Did you want to write an issue for this?

@kasugaijin Yes I will.

PR #1102, opened 2024-10-30
Csv Import Status and Errors

jmilljr24:
# 🔗 Issue

# ✍️ Description

Updates the csv import service to return a Data object with a summary of the import. This will return :success?, :count, :no_match, :errors. Additional arguments can be define in a future PR depending on needs/user feedback.

Minor fixes:
- Wrap transaction around  creating the `FormSubmission` and `FormAnswers`
- Add csv_timestamp to the initial `FormSubmission`. This omission was causing duplicate entries.

# 📷 Screenshots/Demos


app/controllers/organizations/staff/external_form_upload_controller.rb
```diff
@@ -4,18 +4,20 @@ class ExternalFormUploadController < Organizations::BaseController
...
-        file = params[:files]
+        authorize! :external_form_upload, context: {organization: Current.organization}
 
         # Only processes single file upload
-        import_service = Organizations::Importers::GoogleCsvImportService.new(file)
-        import_service.call
+        import = Organizations::Importers::GoogleCsvImportService.new(params[:files]).call
+
+        if import.success?
+          # do something
```
kasugaijin, line 17: Did you want to make an issue for handling these paths? 
I think we want to render a summary table of imports beneath the form input with turbo frame right?

app/controllers/organizations/staff/external_form_upload_controller.rb
```diff
@@ -4,18 +4,20 @@ class ExternalFormUploadController < Organizations::BaseController
...
+        authorize! :external_form_upload, context: {organization: Current.organization}
       end
 
       def create
-        authorize! :external_form_upload,
-          context: {organization: Current.organization}
-        file = params[:files]
+        authorize! :external_form_upload, context: {organization: Current.organization}
 
         # Only processes single file upload
```
kasugaijin, line 13: We limit the import field to only accept one file, too. I am sure you noticed this.

app/services/organizations/importers/google_csv_import_service.rb
```diff
@@ -3,41 +3,61 @@
 module Organizations
   module Importers
     class GoogleCsvImportService
+      Status = Data.define(:success?, :count, :no_match, :errors)
```
kasugaijin, line 6: Just a nit: I think a new line could separate the method from the Status definition to make it easier to read.

app/controllers/organizations/staff/external_form_upload_controller.rb
jmilljr24: > Did you want to make an issue for handling these paths? I think we want to render a summary table of imports beneath the form input with turbo frame right? We could have a green 'Success' or a red 'Import Errors' or something along those lines.

Yes. I wanted to make sure this service was good to go before opening the issue for the front end/controller changes.

app/services/organizations/importers/google_csv_import_service.rb
```diff
@@ -3,41 +3,61 @@
...
       def call
+        row_num = 1
         CSV.foreach(@file.to_path, headers: true, skip_blanks: true) do |row|
+          row_num += 1
           # Using Google Form headers
           email = row["Email"].downcase
           csv_timestamp = Time.parse(row["Timestamp"])
 
           person = Person.find_by(email:, organization: @organization)
           previous = FormSubmission.where(person:, csv_timestamp:)
```
kasugaijin, line 24: do you think we can use a more explicit variable name here over `previous` ? something like `previously_matched_form_submission` or the likes. `previous` just seems a little too generic.

app/controllers/organizations/staff/external_form_upload_controller.rb
jmilljr24: 100%. I added the comment because it may be confusing to see `params[:files]` (plural). The front end currently handling the allowing only one file. Is there a need for some backend control over that? If the front end was bypassed it should raise an error in the service.

app/services/organizations/importers/google_csv_import_service.rb
jmilljr24: Yes, I like that.

app/services/organizations/importers/google_csv_import_service.rb
```diff
@@ -3,41 +3,61 @@
...
+            next
           else
-            create_form_answers(FormSubmission.create!(person:, csv_timestamp:), row)
+            latest_form_submission = person.latest_form_submission
+            ActiveRecord::Base.transaction do
+              # This checks for the empty form submission that is added when a person is created
+              if latest_form_submission.csv_timestamp.nil? && latest_form_submission.form_answers.empty?
+                latest_form_submission.update!(csv_timestamp:)
+                create_form_answers(latest_form_submission, row)
+              else
```
kasugaijin, line 37: can we add a comment on this `else` condition...something like if we get here it means that the person has a new entry in the CSV so we want to import the data...or the likes. I already forgot what getting here means and had to think for a second 👯‍♂️

app/services/organizations/importers/google_csv_import_service.rb
```diff
@@ -3,41 +3,61 @@
...
       def initialize(file)
         @file = file
         @organization = Current.organization
+        @count = 0
+        @no_match = []
+        @errors = []
       end
 
       def call
+        row_num = 1
```
kasugaijin, line 16: do we not have access to a row number helper method with this gem?

jmilljr24: I couldn't find it in the CSV doc's. Would love to be told I'm wrong haha

kasugaijin (commented):
This is a solid improvement that gives more visiblity on the import! @jmilljr24 
Just a few small comments

app/services/organizations/importers/google_csv_import_service.rb
kasugaijin: Monsieur GPT said
The Ruby CSV library does not provide a direct row or row_number helper method for tracking the row number during iteration. However, you can easily track row numbers yourself while iterating through a CSV file by using with_index.

Here’s how you can get the row number with each row in Ruby:

```
CSV.foreach("your_file.csv", headers: true).with_index(1) do |row, index|
  puts "Row #{index}: #{row}"
end
```

app/controllers/organizations/staff/external_form_upload_controller.rb
kasugaijin: hmm we could add a before_action in the controller to check the length of the files array and do X if it is >1

kasugaijin: Given it could be used in multiple controllers, makes sense to add in a controller concern that can be included as needed.

app/services/organizations/importers/google_csv_import_service.rb
jmilljr24: Do you feel that is cleaner that what I currently have? Happy to change it, or change the naming of `row_num` to `index`. Not sure `with_index` gains anything.

kasugaijin: I think with_index is a little cleaner. I think generally if index methods are available it makes sense to use them. So I’ve been told on my PR reviews at work :)

jmilljr24: That does look better now that I typed it out. Thanks!!

app/services/organizations/importers/google_csv_import_service.rb
jmilljr24: New line is removed with standard:fix. But I do agree

kasugaijin: oh weird! Ok well looks like we have to keep it as is otherwise linting will fail

app/controllers/concerns/file_upload_limitable.rb
```diff
@@ -0,0 +1,16 @@
+module FileUploadLimitable
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :single_file_upload_limit, only: [:create]
```
kasugaijin, line 5: can we rename this something like `allow_only_one_attachment`

app/controllers/concerns/file_upload_limitable.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  private
+
+  def single_file_upload_limit
+    return unless params[:files].is_a?(Array)
```
kasugaijin, line 11: do you think using `one?` here will be easier to read?
https://apidock.com/ruby/Enumerable/one%3F

we could return if `params[:files]` only contains one element

jmilljr24: So it interesting how files are handled in rails and I'd have to look into it deeper but when only one file is attached `params[:files].class` returns `ActionDispatch::Http::UploadedFile`. If more than one file is attached you get an array of uploaded files. So for a single file upload `params[:files].one?` returns undefined method.

app/controllers/concerns/file_upload_limitable.rb
jmilljr24: Yes that makes more sense. Any other naming suggestions regarding the concern? Nothing sounded great when I made it but it "works"

app/controllers/concerns/file_upload_limitable.rb
kasugaijin: Ahh yes that makes sense then. I thought it couldn’t be that simple based on how you had it in hindsight.

app/controllers/concerns/file_upload_limitable.rb
kasugaijin: The way we have this now using a before action within the concern will limit the expandability of this concern. We could go a bit more generic on the naming and remove the before action to the controller as needed. So in this case we could call the concern something like AttachmentManageable and then we can keep a wider variety of attachment related methods in here. Does that resonate?

kasugaijin (approved):
Yeah looks great!

Issue #1101, opened 2024-10-30
Prevent editing of emails on User


kasugaijin:
We currently allow users to edit their `User` email address. For example, log in as an adopter and go to dashboard > edit settings, or as staff, go to the dashboard, then top right, click on your avatar > account settings. 

Please disable the email field so people cannot edit it. 

Please also add a model validation on User that prevents changing of the email on `update`

MichaScant:
Hey @kasugaijin can I be assigned this issue?

jmilljr24:
@MichaScant Have at it! Thanks

MichaScant:
Hey, would something like this work for the model validation?

<img width="967" alt="Screenshot 2024-10-30 at 2 39 10 PM" src="https://github.com/user-attachments/assets/79035f40-36d7-41d0-8539-675aac75cfb5">

kasugaijin:
> Hey, would something like this work for the model validation?
> 
[...]

PR #1104, opened 2024-10-30
#1101: Prevent user from changing their email address

MichaScant:
# 🔗 Issue
#1101

# ✍️ Description

- Disabled the ability to edit emails on both the staff and adopter accounts
- Added a validation method in the User model to ensure the email update cannot be changed

# 📷 Screenshots/Demos

Screenshot 2024-10-30 at 2 50 53 PM


kasugaijin (approved):
Awesome thank you @MichaScant

Issue #1058, opened 2024-10-11
Feedback Form: Add topic dropdown and explanation of how to report a bug fix


kasugaijin:
On this feedback page, please add a required select called Subject, that has values: General feedback, Feature request, Bug. Whatever is selected should then become the title of the email sent to the developers via the Mailer.

I think it would also be good to add a link 'How to submit a bug or feature request' above the Message label. When clicked, we use a turbo frame to load the response above the message label. We can use the Feedback Controller #show for this. Now, I **think** placing a turbo frame within a form should work fine, as long as it's a sub-section of the form. If this does not work, you can place this frame/link outside the form, just under the description `Found a bug? Want to request a feature? Contact our dev team with the form below.`

The content rendered in the frame should be a card with the following text (you can refine this text - ask GPT for something better perhaps! But keep it brief):

**Reporting Bugs**
Provide as much detail as possible including:
- the organization name
- the url of the page the bug was found
- the steps that led to the bug
- your expected outcomes versus what you experienced

**Requesting Features**
Provide as much detail as possible including:
- what page the feature should live
- what the feature should do
- why this feature will be useful

<img width="752" alt="image" src="https://github.com/user-attachments/assets/3b479218-cf8c-482a-96d8-a25afcc73a90">

jmilljr24:
On it. I will finalized after #1051 is merged.

jmilljr24:
I played around with the 'how to' link a bit. I'm not in love with link above the message select but maybe I'm not doing it how you imagined? I also tried adding it below the submit button and I kind of like it better but not sure. Just looking for thoughts/idea.

**Option 1**
![Screenshot from 2024-10-12 17-24-54](https://github.com/user-attachments/assets/f6c75dad-a442-45ae-9454-6f9e75232176)
![Screenshot from 2024-10-12 17-24-40](https://github.com/user-attachments/assets/2dbeaa01-ff30-491e-9305-b0f4b0897855)

**Option 2**
![Screenshot from 2024-10-14 14-58-39](https://github.com/user-attachments/assets/278c3a94-4553-4d8e-8c02-7d3df7f297e4)
![Screenshot from 2024-10-14 14-58-49](https://github.com/user-attachments/assets/81097609-05c9-401b-9756-4f80100b211a)

kasugaijin:
@jmilljr24 I know you're working on the pet species PR. Are you also planning to work on this one given your steps above? No pressure either way.

jmilljr24:
[...]

PR #1109, opened 2024-10-31
1009 direct cta buttons to form

Meri-MG:
# 🔗 Issue
#1009 

# ✍️ Description
The CTA buttons on the app root page should take the user to a form for organizations to fill out so they can request access to the app (see how we implemented the Feedback form available in the footer and on dashboards). When the user submits this form, it should send an email to the dev email account (TDB) with the title New Organization Request and provide the details from the form.

In this pr:
- Added `OrganizationAccountRequestsController`.
- Added `OrganizationAccountRequest` model.
- Added a request form.
- Added `OrganizationAccountRequestsMailer`.
- Added tests for the changes.
- Turned the province state field into the text field without using the country state stimulus controller.

# 📷 Screenshots/Demos
Screenshot 2024-10-31 at 09 51 05
Screenshot 2024-11-04 at 14 28 43


app/mailers/organization_account_requests_mailer.rb
```diff
@@ -0,0 +1,15 @@
+class OrganizationAccountRequestsMailer < ApplicationMailer
+  def create_new_organization_account_request
+    @organization_name = params[:name]
+    @requester_name = params[:organization_requester_name]
+    @organization_phone = params[:phone_number]
+    @organization_email = params[:email]
+    # @country = params[:locations][:country]
```
Meri-MG, line 7: I'm having a problem accessing the locations param. could i get any help here?

test/system/organization_account_request_test.rb
```diff
@@ -0,0 +1,25 @@
...
+    should "direct to the new organization's form", js: true do
+      visit root_url
+      click_on "Request an account", match: :first
+
+      fill_in "Organization name", with: "New organization"
+      fill_in "Your name", with: "Doe Smith"
+      fill_in "Organization phone number", with: "12345678"
+      fill_in "Organization email", with: "organization@example.com"
+      select "France", from: "Country"
+      select "Corsica", from: "Province/State"
```
Meri-MG, line 18: I'm having a problem with the province selection in this test which is populated by the stimulus controller.

jmilljr24 (commented):
So I think this can be simplified a bit. It looks like you are trying to use the Organization model and Location model but I don't think we need to do that. We are not creating a new Organization in this PR. This is simply sending some form fields in an email to the Devs. The Dev's will manually make a new Organization in the database.

Take a look at `contacts_controller.rb` and `dev_contacts_controller.rb`. I don't believe you need to add the migration to Organizations or use the Stimulus controller you added.

The nested params could probably be avoided as well. I think it should work with just `params.require(:contact).permit(:name, :email, :message, :requester_name, :city_town)` etc... You may need to update the attr_accessor in `app/models/contact.rb`?

app/controllers/organization_account_requests_controller.rb
```diff
@@ -0,0 +1,34 @@
+class OrganizationAccountRequestsController < ApplicationController
+  skip_before_action :authenticate_user!
+  skip_verify_authorized
+
+  def new
+    @organization_acccount_request = Organization.new
```
jmilljr24, line 6:
```suggestion
     @contact = Contact.new
```

app/controllers/organization_account_requests_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def create
+    @organization_acccount_request = Organization.new(organization_params)
```
jmilljr24, line 10:
```suggestion
  @contact = Contact.new(contacts_params)
```

app/controllers/organization_account_requests_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    if @organization_acccount_request.valid?
+      OrganizationAccountRequestsMailer.with(organization_params)
+        .create_new_organization_account_request.deliver_now
+
+      redirect_to root_path, notice: I18n.t("contacts.create.success")
```
jmilljr24, line 16:
```suggestion
    if @contact.valid?
      OrganizationAccountRequestMailer.with(contact_params).send_message.deliver_later
      redirect_to root_path, notice: I18n.t("contacts.create.success")
```

kasugaijin (commented):
@Meri-MG I agree with @jmilljr24 that is a simpler implementation following how we handle the other contact flows (feedback form, org contact). These just have a controller that instantiates a new contact object that can be used to render the form, and submit to the create action, which sends the email. We just need a new `OrganizationRequest` model that sets the validations so we can call `.valid?` on the object we build on the create action. There's no need to worry about the Organization or Location models in this. The `OrganizationRequest` model will specify the attributes we permit/require. 

See `app/controllers/contacts_controller.rb` and `app/models/contact.rb`
Does this make sense?

app/controllers/organization_account_requests_controller.rb
MooseCowBear: One thing we might want to think about is our current Contact model validates for name, email, message but in this case we will want to validate for different fields.

app/mailers/organization_account_requests_mailer.rb
MooseCowBear: Hi @Meri-MG , the reason this is happening is sort of three-fold. 

1. In the form, you have the model as `Organization` but this is the name of the class and not an instance of a model. And as a result the params are coming back as containing `{ 'locations' =>  ... }` instead of `{ locations_attributes => ... }`. In order to get a `locations_attributes` key in the params, you'd need to pass an instance of the organization model, eg. `Organization.new`. 

2. Passing `organizations_params` to the mailer, there isn't a `locations` key because only `locations_attributes` were permitted

3. Switching `Organization.new` for `Organization` and then trying to access `params[:locations_attributes][:country]` etc in the mailer still wouldn't quite work because there's going to be an extra key to traverse to get to the value you want. The params will actually end up looking like `{ 'locations_attributes' => { '0' => { 'country' => ... } } ...}`. So to pick out the country value you'd end up doing something like `@country = params.dig('locations_attributes', '0', 'country')`

All that being said, switching to using a Contact model in the controller will mean not having to deal with this particular problem

app/controllers/organization_account_requests_controller.rb
MooseCowBear: It might be nice to add a flash message here in the success case

app/views/organization_account_requests/new.html.erb
```diff
@@ -0,0 +1,106 @@
...
+
+                <div class="form-group">
+                  <%= form.text_field :name,
+                    label: 'Organization name',
+                    class: 'form-control',
+                    data: {
+                      slug_from_title_target: 'title',
+                      action: 'keyup->slug-from-title#updateSlug'
+                    } %>
+                </div>
```
MooseCowBear, line 43: Probably organization name and email should also be required. Possibly also phone number?

app/views/organization_account_requests_mailer/create_new_organization_account_request.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    <p>We’ve got a new organization request! Here are the details:</p>
+
+    <ul>
+      <li><strong>Organization Name:</strong> <%= @organization_name %></li>
+      <li><strong>Contact Name:</strong> <%= @requester_name %></li>
+      <li><strong>Organization Phone Number:</strong> <%= @organization_phone %></li>
+      <li><strong>Organization Email:</strong> <%= @organization_email %></li>
+      # <li><strong>Country:</strong> <%= @country %></li>
+      # <li><strong>City/Town:</strong> <%= @city_town %></li>
+      # <li><strong>Province/State:</strong> <%= @province_state %></li>
```
MooseCowBear, line 15: @kasugaijin @jmilljr24 @mononoken One thing to think about. The country and state/province information comes back as codes and not human readable names, which is good for saving records to the database, but for vetting would it be more helpful to have the names?

config/environments/development.rb
```diff
@@ -67,7 +67,7 @@
   config.i18n.raise_on_missing_translations = true
 
   # Annotate rendered view with file names.
-  # config.action_view.annotate_rendered_view_with_filenames = true
+  # config.action_view.annotate_rendered_view_with_filenames = false
```
MooseCowBear, line 70: Why do we need this?

db/migrate/20241024094215_add_customer_name_column_to_organizations.rb
```diff
@@ -0,0 +1,9 @@
+class AddCustomerNameColumnToOrganizations < ActiveRecord::Migration[7.2]
+  def up
+    add_column :organizations, :requester_name, :string
+  end
+
+  def down
+    remove_column :organizations, :requester_name
+  end
+end
```
MooseCowBear, line 9: This can be removed since we won't be creating the new organization until it has been vetted

app/controllers/organization_account_requests_controller.rb
kasugaijin: Yes we will need a new model! I mention in my comment below.

app/views/organization_account_requests_mailer/create_new_organization_account_request.html.erb
kasugaijin: Yes that would be nice. I think we can tackle that in a separate issue.

app/controllers/organization_account_requests_controller.rb
jmilljr24: My train of thought was to use the Contact model as we already use it for `contacts_controller` and `dev_contacts_controller` (being renamed to feedback in #1051  ) and conditionally handle the validations in one model. A new model works a well. 

We will be adding a subject field in #1058 so that could use the same Contact model or would need a new model if we are separating them all out by controller. That will put us at 3 models. (Contact, feedback, OrganizationAccountRequest)

kasugaijin: Yes we could use one. I feel like it would get outgrown quickly though and more of a fan of a dedicated model per form (we can reuse the same model for like forms). We could also move them to a specific folder under app/models to keep them clear and separate from other models that have tables associated.

config/environments/development.rb
Meri-MG: I'm just enabling it locally to see where the code for the particular page comes from.

app/mailers/organization_account_requests_mailer.rb
Meri-MG: Oh, I see. thanks a lot for clarifying.

app/controllers/organization_account_requests_controller.rb
Meri-MG: Do you mean a different type of flash message than the notice?

app/controllers/organization_account_requests_controller.rb
Meri-MG: Thanks everyone for the feedback, I added a new `OrganizationAccountRequest` model.

test/system/organization_account_request_test.rb
Meri-MG: I changed the province select field to a text field to avoid the system test error, as we are no longer building it through the location object.

app/controllers/organization_account_requests_controller.rb
MooseCowBear: @Meri-MG Sorry, I completely looked past the fact that you had already added one 🫠

app/models/organization_account_request.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  before_validation :normalize_phone
+
+  validates :name, :email, :city_town, :country, :province_state, presence: true
+  validates_format_of :email, with: /\A([^@\s]+)@((?:[-a-z0-9]+\.)+[a-z]{2,})\z/i
+  validates :phone_number, presence: true, phone: {possible: true, allow_blank: true}
+
+  private
+
+  def normalize_phone
```
kasugaijin, line 14: Can't we use the `Phoneable` module instead? This method is already there.

`app/models/concerns/phoneable.rb`

config/routes.rb
```diff
@@ -91,4 +90,5 @@
 
   resources :contacts, only: %i[new create]
   resources :dev_contacts, path: "feedback", only: %i[new create]
+  resources :organization_account_requests, only: %i[new create]
```
kasugaijin, line 93: I think we just need `resource`

test/controllers/organization_account_requests_controller_test.rb
```diff
@@ -0,0 +1,45 @@
+require "test_helper"
+
+class OrganizationAccountRequestControllerTest < ActionDispatch::IntegrationTest
+  test "#new" do
+    get new_organization_account_request_path
+
+    assert_response :success
+    assert_select "h1", "New organization account request"
```
kasugaijin, line 8: I think we can get rid of this assertion - it's a personal preference. I started out testing DOM elements heavily in this app and grew to really dislike them because they are so brittle. If anything I think asserting success in the response is sufficient.

kasugaijin (commented):
@Meri-MG this is great! Thank you for digesting the feedback and changing course to approach is this way. Just a few small comments.

config/routes.rb
Meri-MG: you are right 😬

app/models/organization_account_request.rb
Meri-MG: In phonebale, we have a `before_save` callback, and since we aren't saving these records into the database and only validating them, it doesn't work for me, hence i had to add the method separately 😬

Meri-MG: but I noticed now the Organization module has Phoneable included but still defines the method as well, I removed it from there.

Meri-MG: yes, but I cant use the `before_save` callback in my class 🤔

kasugaijin: Ooh I understand now. I wonder if we can conditionally call the before save callback in the module then. Ie it gets called unless the calling class name is this organization requests class? Don’t spend ages on it…but I think we can try that?

kasugaijin: OK no problem. Let's move on with your current!

kasugaijin (approved):
thank you

Issue #1098, opened 2024-10-30
Bug: super admin cannot access /staff/organization/edit


kasugaijin:
Log in as staff and go to the dashboard. In the navbar there should be a Settings link under Admin section. This is not visible, but you can access it by going to `/staff/organization/edit`. 

The issue is in the policy check in `app/views/layouts/dashboard/_sidebar.html.erb` for some reason is returning false so the link does not show.

Meri-MG:
Hello, @kasugaijin, shall i take this one once https://github.com/rubyforgood/homeward-tails/pull/1109 is merged?

PR #1113, opened 2024-11-05
1098 - Fix  settings link under Admin section

Meri-MG:
# 🔗 Issue
#1098 

# ✍️ Description
Log in as staff and go to the dashboard. In the navbar there should be a Settings link under Admin section. This is not visible, but you can access it by going to /staff/organization/edit.

In `OrganizationPolicy`, the `verify_active_staff!` pre-check was failing because the organization method was returning nil. This happened because the organization method in `ApplicationPolicy` was set to always call `record.organization`, which only works when `record` is something associated with an organization (like a CustomPage, User, or another related class). However, if `record` itself is an `Organization`, `record.organization` is not valid. To fix this, I updated the organization method to return record directly when record is an Organization, ensuring the method works correctly for both cases.

# 📷 Screenshots/Demos
Screenshot 2024-11-05 at 10 29 08

### DIFF
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -26,6 +26,8 @@ class ApplicationPolicy < ActionPolicy::Base
   # Define shared methods useful for most policies.
 
   def organization
+    return record if record.is_a?(Organization)
+
     @organization || record.organization
   end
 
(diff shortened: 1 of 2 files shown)


jmilljr24 (approved):
@Meri-MG LGTM!!

PR #1116, opened 2024-11-07
1058 Add Feedback Subject

jmilljr24:
#1058 

# ✍️ Description

A required dropdown selection is added for Subject on the feedback form. Additionally, a link is added to updated a turbo_frame with additional information on how to submit feedback. The placement of the link and/or displayed information could use improvement. 

# 📷 Screenshots/Demos

Screenshot from 2024-11-07 09-35-23
Screenshot from 2024-11-07 09-35-32


config/routes.rb
```diff
@@ -91,4 +91,5 @@
 
   resources :contacts, only: %i[new create]
   resources :feedback, only: %i[new create]
+  get "feedback", to: "feedback#show"
```
jmilljr24, line 94: Adding show to resources would require an id. This creates a route to the show action without an id as we only have one thing to show.

app/views/feedback/_form.html.erb
```diff
@@ -1,40 +1,54 @@
...
     </div>
     <div class="form-group">
-      <%= f.submit I18n.t('general.submit'), class: 'btn btn-primary' %>
+      <%= f.submit I18n.t("general.submit"), class: "btn btn-primary" %>
     </div>
+
+    <%= turbo_frame_tag "feedback", class: "pt-4" do %>
+      <%= link_to "Click for tips on how to submit a request", feedback_path %>
+    <% end %>
   </div>
```
MooseCowBear, line 53: I'm wondering if this would make more sense as a Stimulus controller to toggle the visibility of the tips? Since the tips themselves are just static content, we can eliminate a request to the server.

jmilljr24: Good point. I was going off the issue, but now that you bring this up, I should be able to do it ever simpler with [bootstrap collapse](https://getbootstrap.com/docs/5.3/components/collapse/#how-it-works). @kasugaijin Can you think of any downside diverging from the issue requirements?

kasugaijin: Not at all! Providing the toggle controller is reusable in other places it makes sense as it’s a common feature. If we go that route it’s worth looking at A) stimulus - I think there is already something like this in the controllers as I believe I used toggling in the old adopter profile form. It might need a bit of a tweak. B) our theme also has toggle functionality built right in. So it would be even better to use that if it works. Eg the FAQs use a toggle feature.

config/routes.rb
kasugaijin: Ah, we could set it `resource` instead of `resources` ?

app/controllers/feedback_controller.rb
```diff
@@ -1,35 +1,49 @@
...
 
   def create
-    @contact = Contact.new(contact_params)
-    if @contact.valid?
-      FeedbackMailer.with(contact_params).send_message.deliver_later
-      redirect_to root_path, notice: I18n.t("contacts.create.success")
+    @feedback = Feedback.new(feedback_params)
+
+    if @feedback.valid?
+      FeedbackMailer.with(feedback_params).send_message.deliver_later
```
kasugaijin, line 19: Let's use `deliver_now` for the time being, until we set up background processing.

app/controllers/feedback_controller.rb
```diff
@@ -1,35 +1,49 @@
...
+      path = case set_layout
```
kasugaijin, line 20: Can we move this to a private method e.g., `set_redirect_path` or the likes

app/mailers/feedback_mailer.rb
```diff
@@ -2,8 +2,9 @@ class FeedbackMailer < ApplicationMailer
   def send_message
     @name = params[:name]
     @email = params[:email]
+    @subject = params[:subject]
     @message = params[:message]
 
-    mail to: "devs@email.com", subject: "New Support Message"
+    mail to: "devs@email.com", subject: @subject
```
kasugaijin, line 8: This reminds me, we should set up ENV variables in another issue. Then we can add the homewardtails email address here.

kasugaijin (commented):
Looks good Justin. Just a few comments to consider.

config/routes.rb
jmilljr24: Ah, right! Update - Using bootstrap, we no longer need a show action/route.

app/controllers/feedback_controller.rb
jmilljr24: I moved it to a private method. I don't think we need a before_action.

jmilljr24:
How does this look?
[Screencast from 2024-11-10 10-42-02.webm](https://github.com/user-attachments/assets/4f7677be-2299-4471-b978-895268581974)

app/controllers/feedback_controller.rb
```diff
@@ -1,35 +1,49 @@
...
       "application"
-    elsif current_user.staff_account
+    elsif current_user.has_role?(:adopter, ActsAsTenant.current_tenant)
+      "adopter_foster_dashboard"
+    else
       "dashboard"
+    end
+  end
+
+  def path
```
kasugaijin, line 39: I have been starting at this method as it just doesn't quite feel 'right'. We are defining the path based on the layout. It works. But the layout is defined based on whether there is no `current_user`, the user is an adopter, the user is a staff. It seems to me we should use the same checks for the redirect path. Using the layout feels like it's a proxy, and would break if we ever changed the layouts. The driver here should be the user type, or lack thereof. Thoughts on this?

It's not a hill I am going to die on...but I think it would be clearer.

kasugaijin (approved):
looks great @jmilljr24! Just one comment to consider on that private method. Your call on what you think is best.

app/controllers/feedback_controller.rb
jmilljr24: For sure. I think I originally was trying to keep things DRY and use the set_layout method and I kinda just went down that hole. Then I was lazy and just moved the case into a method without thinking about it more.

Maybe something like
```
  def path
    if current_user.nil?
     root_path
    elsif current_user.has_role?(:adopter, ActsAsTenant.current_tenant)
      adopter_fosterer_dashboard_index_path
    else
      staff_dashboard_index_path
    end
  end
```

Same as set_layout but returning the path instead. Maybe there is a simple way to use set_layout for both but I can't think of it at the moment or even if that is a smart idea.

kasugaijin: Yes I think that works ^. I also wondered about setting both at once...but single responsibility is probably better to stick to when in doubt!

PR #1141, opened 2024-11-13
Workflow deploy weekly fixes

jmilljr24:
# 🔗 Issue

# ✍️ Description

Changed to use the check-run api. Also simplified a bit.

# 📷 Screenshots/Demos


.github/workflows/kamaldeploy.yml
```diff
@@ -20,34 +19,26 @@ jobs:
...
+            if [ "$conclusion" != "success" ]; then
+              echo "Commit "${{ env.SHA }}" has failed checks."
+              exit 1
+            fi
+          done
 
   deploy:
     runs-on: ubuntu-latest
     needs: check_latest_commit
-    if: needs.check_latest_commit.outputs.commit_status == 'success'
```
jmilljr24, line 42: Since we exit with code 1 (failure/error) if there are any fail check, we don't need to check for passing here. This job with never run as it needs "check_latest_commit" job to be be completed without error.

.github/workflows/kamaldeploy.yml
```diff
@@ -20,34 +19,26 @@ jobs:
...
+      - name: Confirm all checks passed
         run: |
-          COMMIT_SHA="${{ env.SHA }}"
-          echo "Checking status of commit ${COMMIT_SHA}..."
-
-          # Get the status of the latest commit using GitHub API
+          # Returns json data on all checks that were run
           response=$(curl -L \
           -H "Accept: application/vnd.github+json" \
-          -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
```
jmilljr24, line 35: We shouldn't need this as the information we need from the api is public.

kasugaijin (approved):

PR #1142, opened 2024-11-13
1106-staff-deactivation-refactor

wandergithub:
# 🔗 Issue

https://github.com/rubyforgood/homeward-tails/issues/1106
# ✍️ Description

Refactor staff deactivation
# 📷 Screenshots/Demos
image


test/controllers/organizations/staff/activations_controller_test.rb
```diff
@@ -0,0 +1,34 @@
...
+  setup do
+    @organization = ActsAsTenant.current_tenant
+    @staff = create(:admin)
+    sign_in @staff
+  end
+
+  context "authorization" do
+    include ActionPolicy::TestHelper
+
+    context "#update_activation" do
```
kasugaijin, line 14: do we need this authorization test here? I think we should leave authorization tests to the dedicated authorization test files, and you have an activations_policy_test file

test/controllers/organizations/staff/activations_controller_test.rb
```diff
@@ -0,0 +1,34 @@
...
+        end
+      end
+    end
+  end
+
+  test "update activation should respond with turbo_stream when toggled on staff page" do
+    user = create(:super_admin)
+    sign_in user
+
+    patch activation_update_activation_url(@staff), as: :turbo_stream
```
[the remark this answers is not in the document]
kasugaijin: Hmm why not test the outcome rather than the mechanism? So, ideally we should compare the states of `@staff` before and then after to assert it changed as we expected. Does that make sense?

app/policies/organizations/activations_policy.rb
```diff
@@ -0,0 +1,18 @@
...
+  class ActivationsPolicy < ApplicationPolicy
+    pre_check :verify_organization!
+    pre_check :verify_active_staff!
+
+    def update_activation?
+      return false if record.id == user.id
+
+      record_role = record.roles.first.name
+
+      if record_role == "super_admin" || record_role == "admin"
```
kasugaijin, line 11: you could shorten this a little to 
`if %w[super_admin admin].include?(record_role)`

config/routes.rb
```diff
@@ -9,6 +9,9 @@
     resources :home, only: [:index]
     resources :adoptable_pets, only: %i[index show]
     resources :faq, only: [:index]
+    resources :activations do
+      patch "update_activation"
```
kasugaijin, line 13: we don't need this custom route anymore (that was the intention of moving this to its own resource). So we can can keep it RESTful and use the `update` action on this controller.

app/controllers/organizations/activations_controller.rb
```diff
@@ -0,0 +1,23 @@
+module Organizations
+  class ActivationsController < Organizations::BaseController
+    def update_activation
+      @user = User.find(params[:activation_id])
```
kasugaijin, line 4: I think it would be more appropriate to send `user_id` in the params. We may need to explicitly pass that in the path helper. So, when you remove this custom action and just use the `update` action it will be something like `activations_path(user_id: user.id)` 
I think rails will yell at you for not having an activations id, but if we change the routes to `resource :activations`, instead of `resources` we should be fine. That make sense?

jmilljr24:
@wandergithub Awesome job! 

The only thing I noticed on my end is that the flash is no longer working. Are you seeing the same thing?

This is on the main branch:
Screenshot from 2024-11-15 09-07-02

wandergithub:
Now that you point that out 🤔 
I will take a look.

wandergithub:
Something is happening with the template, I think related to turbo stream and how the flash message is being handled but I don't fully understand how it works jet...

jmilljr24:
> Something is happening with the template, I think related to turbo stream and how the flash message is being handled but I don't fully understand how it works jet...

The update action in ActivationsController is implicitly looking for `update.turbo_stream.erb` in `views/organizations/activations`. You can explicitly render a file that is in a differently path but we probably should stick with convention unless there is a specific reason you can think of not to.

kasugaijin:
Yeah what @jmilljr24 said. Check out `app/views/organizations/staff/staff/update_activation.turbo_stream.erb`. You need this for your new controller and action, so just move that turbo stream view file to the correct view directory for that new controller and rename it to `update.turbo_stream.erb`. You will note that this file does two things. It replaces the toggle because it should change after changing the activation, and it renders the flash.

jmilljr24:
> Yeah what @jmilljr24 said. Check out `app/views/organizations/staff/staff/update_activation.turbo_stream.erb`. You need this for your new controller and action, so just move that turbo stream view file to the correct view directory for that new controller and rename it to `update.turbo_stream.erb`. You will note that this file does two things. It replaces the toggle because it should change after changing the activation, and it renders the flash.

@kasugaijin I don't believe we need the turbo replace of the toggle any longer. I think that was used when the page was rendering two layouts that were user selected(table vs cards). If the user clicked the toggle in the table, the card wouldn't reflect the change until a page fresh.

kasugaijin:
@jmilljr24 Ah yes we got rid of those cards...good point. We can remove the toggle bit in the turbostream response, then @wandergithub !

wandergithub:
It works now! 
image

Thanks guys. Now I understand better how turbo_streams and the real-time update work.

test/policies/organizations/activations_policy_test.rb
```diff
@@ -0,0 +1,76 @@
...
+      setup do
+        @user = create(:fosterer)
+      end
+
+      should "return false" do
+        assert_equal false, @action.call
+      end
+    end
+
+    context "when user is staff" do
```
kasugaijin, line 46: Ideally this test is called `when user is admin`

test/policies/organizations/activations_policy_test.rb
```diff
@@ -0,0 +1,76 @@
...
+      setup do
+        @user = create(:admin)
+      end
+
+      should "return false" do
+        assert_equal false, @action.call
+      end
+    end
+
+    context "when user is staff admin" do
```
kasugaijin, line 56: and `when user is super admin`

test/policies/organizations/activations_policy_test.rb
kasugaijin, line 1: It would be good to add two more contexts to cover more cases
- when user being mutated is a fosterer and the person doing the action is an admin, superadmin
- when user being mutated is an adopter and the person doing the action is admin, superadmin

kasugaijin (commented):
Looks nice and clean @wandergithub ! I like the refactor. 
Just a couple more policy tests we could add not we are also mutating adopters and fosters. Should be a pretty quick addition. Let me know if you have any questions!

kasugaijin:
@wandergithub just pining to see if you can finish those last few tests? If you're too busy, let me know and I can take care of it.

kasugaijin (approved):
@wandergithub I finished this one off! Thanks for getting it to where it was.

PR #1147, opened 2024-11-16
1143 make section content larger

Isaac3924:
# 🔗 Issue

[https://github.com/rubyforgood/homeward-tails/issues/1143](url)

# ✍️ Description

The issue described was an increase in size for the 3 headings and 3 paragraphs within the section element "bg-white py-4 shadow-sm" on lines 32-61 of app/views/root/index.html.erb. As well as a change to the actual text in some of the paragraphs and one of the headings.

I used the bootstrap utilities to adjust the headings as the custom.scss file asked me to use the utilities over creating my own custom css so as to keep to the same styles. I added in the fs- utilities to the h4 elements giving them the values of fs-2 putting the headers at 2.5rem for font size. I find this to be too big, but found that fs-3 would set it to 1.25rem which was much too small. Adding fs to the paragraph elements proved fruitless as fs-5 had the paragraphs stay at the same font size without it (0.875rem) while the next step up with fs-4 set it to 1rem font size, but this causes the longer paragraphs to reach two lines on the xl breakpoint for media which was the major crux of the issue. 

A solution for this could potentially be reached by either creating a custom class for this in custom.scss (this is detailed further in my second commit message) however, this seems frowned upon due to the comment there seeming to want us to stay within the purview of bootstrap's utilities instead to maintain cohesion in UI and design. With that in mind, I applied the font size values I thought best.

Further changes are straight forward. The config/locales/en.yml file had the string values adjusted to properly display the new messages and headings as requested by the issue. Furthermore, I changed the keys on lines 115 & 116 from personalized and personalized_description to save_time & save_time_description respectively to better reflect the changes made there, as well as change the calls to these keys in app/views/root/index.html.erb lines 54 & 55 of course.

# 📷 Screenshots/Demos

image

### DIFF
--- a/app/views/root/index.html.erb
+++ b/app/views/root/index.html.erb
@@ -51,8 +51,8 @@
             <div class="col-xl-4 col-lg-4 col-md-6 mb-lg-0 mb-4">
                 <div class="d-flex justify-content-center align-items-center">
                     <div class="ms-3">
-                        <h4 class="mb-0 fw-semibold text-center text-dark fs-2"><%= t('.personalized')%></h4>
-                        <p class="mb-0 text-center" style = "color: #373737;"><%= t('.personalized_description')%></p>
+                        <h4 class="mb-0 fw-semibold text-center text-dark fs-2"><%= t('.save_time')%></h4>
+                        <p class="mb-0 text-center" style = "color: #373737;"><%= t('.save_time_description')%></p>
                     </div>
                 </div>
             </div>
--- a/app/views/root/index.html.erb
+++ b/app/views/root/index.html.erb
@@ -35,7 +35,7 @@
             <div class="col-xl-4 col-lg-4 col-md-6 mb-lg-0 mb-4">
                 <div class="d-flex justify-content-center align-items-center">
                     <div class="ms-3">
-                        <h4 class="mb-0 fw-semibold text-center text-dark"><%= t('.free')%></h4>
+                        <h4 class="mb-0 fw-semibold text-center text-dark fs-2"><%= t('.free')%></h4>
                         <p class="mb-0 text-center" style = "color: #373737;"><%= t('.free_description')%></p>
                     </div>
                 </div>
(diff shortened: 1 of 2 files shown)


Isaac3924:
Let me know if I missed anything or if there are any questions to my logic or choices!

kasugaijin:
@all-contributors
please add @Isaac3924 for code.

kasugaijin (commented):
What do you think about making the text beneath the headings a size larger?

Isaac3924:
@kasugaijin Just saw your comment. I think it's doable, but not with the Bootstrap utilities. I see some ways to potentially do it and let the paragraph elements sit on the same line: 

1.) Either increase the section's width to give the paragraph text the room it needs to not move to a second line but that would make the section and the text wider than the rest of the content of the page.
2.) Adjust the padding or margins of individual elements to give more room within the section for the paragraphs. Would need to mess with it more and make sure that spacing between the paragraphs still looks good. They are inheriting a margin to their left from the bootstrap utils. Removing that might help.
3.) Create a custom css class in scss for the paragraph elements that would increase it a bit (0.875rem to about ~0.9rem) and allow it to stay on the same line. This might cause the paragraphs to be very close to each other and mess with readbility, however.
4.) Change the text for the two longer paragraphs to allow for more space, if they're shortened, then could probably increase their font size while keeping them on the same line.

Options 2 and 3's issues would be needing to break away from bootstrap's util's designs which custom.scss seemed to discourage.

kasugaijin:
> @kasugaijin Just saw your comment. I think it's doable, but not with the Bootstrap utilities. I see some ways to potentially do it and let the paragraph elements sit on the same line:
> 
> 1.) Either increase the section's width to give the paragraph text the room it needs to not move to a second line but that would make the section and the text wider than the rest of the content of the page. 2.) Adjust the padding or margins of individual elements to give more room within the section for the paragraphs. Would need to mess with it more and make sure that spacing between the paragraphs still looks good. They are inheriting a margin to their left from the bootstrap utils. Removing that might help. 3.) Create a custom css class in scss for the paragraph elements that would increase it a bit (0.875rem to about ~0.9rem) and allow it to stay on the same line. This might cause the paragraphs to be very close to each other and mess with readbility, however. 4.) Change the text for the two longer paragraphs to allow for more space, if they're shortened, then could probably increase their font size while keeping them on the same line.
> 
> Options 2 and 3's issues would be needing to break away from bootstrap's util's designs which custom.scss seemed to discourage.

Ah thank you for the thorough feedback! Based on that, let's roll with your current changes for now and think about if/how we can change the other text in the future as needed. Thanks @Isaac3924

Issue #1152, opened 2024-11-16
Org Root: add copy to translation file


kasugaijin:
We have a few things on the Org home page `app/views/organizations/home/index.html.erb` that are hard-coded strings. These should be moved to en.yml

Isaac3924:
Could I have this assigned to me?

jmilljr24:
@Isaac3924 All yours!

Isaac3924:
Made a pull request that should resolve this issue! Please let me know if there's further refactroring I need to do, or something I missed!

PR #1159, opened 2024-11-18
Add refactor: added in lines 96-103 within config/locales/en to and a…

Isaac3924:
# 🔗 Issue

[https://github.com/rubyforgood/homeward-tails/issues/1152](url)

# ✍️ Description

Added in lines 96-103 within config/locales/en to and assigned them the string values that were originally in app/views/organizations/home/index which have been removed and now reference them within the en yml file.

# 📷 Screenshots/Demos

image
image


app/views/organizations/home/index.html.erb
```diff
@@ -81,11 +81,11 @@
 
       <div class="d-flex flex-column justify-content-center align-items-center">
         <div class="mb-3">
-          <%= image_tag('paw-account.svg', width: '150', alt: "Create and account") %>
+          <%= image_tag('paw-account.svg', width: '150', alt: "Create an account") %>
         </div>
         <div>
           <p class="text-black">
-            1. Create Account
+            <%= t("general.step_1") %>
```
jmilljr24, line 88: What do you think about putting all of the steps in an ordered list? Then we won't need to add the step number to the translation. The translation key would be more descriptive `t(".create_account")`

config/locales/en.yml
```diff
@@ -93,6 +93,14 @@ en:
...
     browse: "Browse"
     view: "View"
+    about_us: "About us"
+    adopt: "Adopt"
+    become_adopter_now: "Become an adopter now!"
+    how_to_apply: "How to apply for adoption"
+    step_1: "1. Create Account"
+    step_2: "2. Complete Questionnaire"
+    step_3: "3. Browse Pets"
+    step_4: "4. Click Apply"
```
jmilljr24, line 103: Just a nit, but I'm thinking it would be a little more organized to scope these like  `en.organizations.home.index`. "About us" could probably stay in default as it's used on the root page as well (needs to be updated in a separate issue)

If you change to this scope, this will allow you to use rails i18n lazy lookup.

`t(".how_to_apply"` will look for `en.organizations.home.index.how_to_apply`

jmilljr24 (changes requested):
@Isaac3924 Thanks! Nothing major, just a couple of comments.

Side note:
To link the issue from your pr you can just type `#1152`. Also if you were to put it with a [keyword](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) such as "resolves #1152 " the issue will be closed automatically if the pr is merged. Not required by any means, just wanted to share.

Also we probably didn't need screenshots for this pr, but screenshots of the site/ui can be helpful for frontend changes.

app/views/organizations/home/index.html.erb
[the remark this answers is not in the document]
Isaac3924: Can do!

app/views/organizations/home/index.html.erb
Isaac3924: Will need to look at it again, but should be doable!

config/locales/en.yml
Isaac3924: I like it, yeah! I was wondering whether to have it in general or not tbh. Will refactor it today!

Isaac3924:
I've done the changes you've asked @jmilljr24 . Let me know if you see any discrepancies!

app/views/organizations/home/index.html.erb
```diff
@@ -72,54 +72,64 @@
...
+        <div class="d-flex flex-column justify-content-center align-items-center">
+          <div class="mb-3">
+            <%= image_tag('paw-account.svg', width: '150', alt: "Create an account") %>
+          </div>
+          <div>
+            <p class="text-black">
+              <li>
+                <%= t('.create_account') %>
+              </li>
+            </p>
```
jmilljr24, line 92: I don't believe this `p` tag is helping us. Can you just add the class to `li` and remove `p`? 

Also I didn't notice this before but can you double check that the text/number is center aligned to the icon above? It seems like the text is aligned but not including the number. Probably happened when switching to the ordered list. If this suggestion doesn't look right please say so.

Maybe there is a bootstrap class that can help, if not adding a style attribute such as `list-style-position: inside;` on the `li`

jmilljr24 (changes requested):
Just one more thing to take a look at. Otherwise this is great!

app/views/organizations/home/index.html.erb
Isaac3924: I'll take another look, sure! Will agree that the p tag could probably just be done away with. But yes, I think this is being caused by the switch to the ordered list. I'll dig into it and hopefully have it done soon!

Isaac3924:
@jmilljr24 Just finished it up! removed the p tags and essentially replaced them with li tags. And added in the `list-style-position: inside;` you mentioned. Didn't need to do it for each individual li element, though. Was able to just apply it to the ol on line 82. That centered all of the text with the consideration of the list's numbers.

jmilljr24 (approved):

jmilljr24:
@Isaac3924 Looks great! Thank you

Issue #1062, opened 2024-10-16
Adoptable Pets: Add empty state


kasugaijin:
If you go to `/adoptable_pets` and perform a search that returns no results, we don't show anything. 

Please add an empty state and render it if the adoptable pets collection is empty. It should say "We don't have any pets matching your search criteria." 

![Image](https://github.com/user-attachments/assets/319d8b8a-f05c-4b2c-bf35-ffb1faacabbb)

Gabe-Torres:
I'll take this one!

kasugaijin:
FYI @Gabe-Torres this PR https://github.com/rubyforgood/homeward-tails/pull/1070 is creating separate adopatable pet pages, by species, but it's using the same controller. So, the empty state rendering shouldn't be impacted much. But, thought I would let you know. P.S. let me know if you have any questions.

kasugaijin:
@Gabe-Torres how's this going?

Gabe-Torres:
I'll have this pr submitted asap!

kasugaijin:
Thanks @Gabe-Torres no major rush! Just checking in :)

Gabe-Torres:
PR submitted!

Gabe-Torres:
Here is the new new. Sorry for the confusion! 🐶 

This PR has the latest main and no previous commits from my last attempt. 
I will delete the previous branch once this issue is closed.

PR #1163, opened 2024-11-19
1062 empty state message for no match

Gabe-Torres:
# 🔗 Issue #1062 

# ✍️ Description

This PR implements adding an empty state message when a user searches for a pet that yields no results. This will help with user confusion by giving a descriptive flash alert - " We don't have any pets matching your search criteria. "

The user will then be redirected to the species they previously searched for. Another option is only to flash a message without the redirect. 

Is there a preference? Not sure if the redirect is needed but it seemed like the right thing to do!

# 📷 Screenshots/Demos

https://jam.dev/c/353f28c2-9742-4b8a-bdba-6603774203a3

### DIFF
--- a/app/controllers/organizations/adoptable_pets_controller.rb
+++ b/app/controllers/organizations/adoptable_pets_controller.rb
@@ -27,6 +27,11 @@ module Organizations
       )
 
       @pets = paginated_adoptable_pets
+
+      if @pets.empty?
+        flash[:alert] = "We don't have any pets matching your search criteria."
+        redirect_to adoptable_pets_path(species: @species)
+      end
     end
 
     def show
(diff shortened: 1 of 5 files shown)


kasugaijin:
Hey @Gabe-Torres thank you!
I should have been more clear in the issue, my bad. We want this empty state to render in the index view, not a flash.
So in the view we would have a conditional something like
```
if there are pets in the collection
  show the results cards
else
  show the empty state message
```

the empty state should be a centered card with the provided text from the issue in it. Does this make sense?

image


image

Gabe-Torres:
hmmm i attempted to start over with the latest main but it didnt get rid of the previous commits that are no longer relevant. Now my commit history on this branch is out of order. To advoid confusion and trouble I'm going to make a new branch and PR! 😅

Issue #1084, opened 2024-10-23
Fosterer Invites: Enable staff to enter fosterer phone number and address when inviting


kasugaijin:
This is a bit more of a complex project. It's not too hard, but involves a bit of nuance. Hopefully my notes are useful, but ask if you have any questions.

Currently, staff can only invite a fosterer with a first name, last name and email. This is because invitations are done on `User` and we reserve the `User` model for authentication and authorization only, so do not store other personal data on it. Instead, we use an associated `Person` model to hold first and last name, email and phone number (and whatever else we need in the future). We also store addresses in a `Location` model.

Invites on the `User` model are handled by the `Organizations::Staff::InvitationsController`. The limitation of the implementation is that we are confined to a single invitation resource to handle inviting staff, and fosterers, so it's an ugly controller. It will do for now, though. 

Desired Outcome
**We need to enable staff to enter all of the relevant data for a fosterer once, when they invite them. This includes email, first name, last name, phone number and location attributes.**

My thoughts are:
- we can add these new fields to the fosterer invitation form (phone, and location attributes)
- then in `Organizations::Staff::InvitationsController` when we create the `User`, we then update its `Person` with the phone number (note that a Person gets created in [this callback](https://github.com/rubyforgood/homeward-tails/blob/64e94fd969a77cb0f85e66b1ec9c4e0f4aa6f29d/app/models/user.rb#L72) when a user is created)
- then create the person's associated location record
- we should just need to permit these params and they should be available to use
- If all these records persist, then we call `invite!`. 

Note, you will need to add a `belongs_to :person` on `Location` and add a migration for a `person_id`, and enable `Person` to accept nested attributes for location. We should not validate `person_id` on `Location`, however, as Locations can exist without a person. We shouldn't need to provide an organization id.

Furthermore, staff currently have the ability to edit a fosterer (`Person`) (see `app/controllers/organizations/staff/fosterers_controller.rb`). Please update this edit form to show, but disable editing, the `email` field. We must not allow editing email, because the Person's email should always match the associated User email (herein lies the issue of storing duplicate data on two models, but is the path we decided for various reasons). Staff should also be able to update location details on this form, so please add these fields.

mononoken:
My understanding

## AC
- [ ] Add phone to fosterer invite form (on Person record)
- [ ] Add location to fosterer invite form (on Person record)
- [ ] Disable email field for fosterer edit

mononoken:
@kasugaijin 
[...]

PR #1169, opened 2024-11-21
1084 add info fosterer invite

mononoken:
# 🔗 Issue
#1084 

# ✍️ Description

We are adding phone number and locations to the fosterer forms. Staff can invite fosterers with this info included or they can add/edit it in the edit form. Neither attribute is required, so they don't have to include the info if they don't want to.

Note, one case I noticed not addressed in this PR: If a fosterer had location data and the staff wanted to completely remove it instead of replacing it, that is currently not supported. Locations were implemented as a separate model, so we would have to add a delete action for the fosterer's location. I can follow up with that if we think that is necessary.

Also, do we want to display the location in the table? 

# 📷 Screenshots/Demos

_fosterer invite form_

SCR-20241123-ohqe

_fosterer edit form_

SCR-20241123-oilv


.gitignore
```diff
@@ -53,3 +53,6 @@
 .DS_Store
 
 /config/credentials/production.key
+
+# LazyVim config file
+.lazy.lua
```
mononoken, line 58: Sneaking in gitignore for LazyVim config file.

app/controllers/organizations/staff/invitations_controller.rb
```diff
@@ -74,4 +82,12 @@ def after_accept_path_for(_resource)
       root_path
     end
   end
+
+  # TODO: We should consider removing this duplicated logic, so we don't have to do this.
```
mononoken, line 86: I'm not sure we need this info in both User and Person, especially the names. Think we can remove these from User. 

Removing duplicated email may be a bit more complicated since it's used for auth purposes, so maybe that can remain.

app/models/location.rb
```diff
@@ -2,32 +2,43 @@
...
-#  index_locations_on_organization_id  (organization_id)
-#
-# Foreign Keys
-#
-#  fk_rails_...  (organization_id => organizations.id)
+#  index_locations_on_locatable  (locatable_type,locatable_id)
 #
 class Location < ApplicationRecord
-  acts_as_tenant(:organization)
+  belongs_to :locatable, polymorphic: true
```
mononoken, line 20: This change can close this story on the kanban that hasn't been converted to an issue yet: https://github.com/orgs/rubyforgood/projects/16/views/1?filterQuery=location&pane=issue&itemId=72155107

config/database.yml
```diff
@@ -18,8 +18,8 @@ default: &default
   adapter: postgresql
   encoding: unicode
   host: localhost
-  username: <%= ENV['DATABASE_USERNAME'] %>
-  password: <%= ENV['DATABASE_PASSWORD'] %>
+  username: <%= ENV.fetch('DATABASE_USERNAME') { 'postgres' } %>
+  password: <%= ENV.fetch('DATABASE_PASSWORD') { 'password' }  %>
```
mononoken, line 22: Added the docker compose defaults here

config/initializers/faker.rb
```diff
@@ -0,0 +1,3 @@
+if Rails.env.test?
+  Faker::Config.locale = "en-US"
```
mononoken, line 2: This is useful for things like phone validations

db/migrate/20240902111455_change_adopter_foster_account_null_on_adopter_applications.rb
```diff
@@ -1,6 +1,10 @@
 class ChangeAdopterFosterAccountNullOnAdopterApplications < ActiveRecord::Migration[7.1]
   def change
-    remove_index :adopter_applications, column: [:adopter_foster_account_id], name: :index_adopter_applications_on_adopter_foster_account_id
-    change_column_null(:adopter_applications, :adopter_foster_account_id, true)
+    if index_exists?(:adopter_applications, :adopter_foster_account_id)
```
mononoken, line 3: I noticed building the db from migrations instead of the schema was not working because of these group of migrations. I wrapped these changes in conditionals as a quick fix instead of trying to edit our migration history.

db/migrate/20241121050334_add_phone_number_to_people.rb
```diff
@@ -0,0 +1,12 @@
+class AddPhoneNumberToPeople < ActiveRecord::Migration[7.2]
+  def change
+    add_column :people, :phone_number, :string, limit: 15
```
mononoken, line 3: Replaced the `phone` column with `phone_number` to match the organizations table naming convention. Plus, I think it is more explicit.

Also added some simple checks to sanitize data.

docker-compose.yml
```diff
@@ -9,7 +8,7 @@ services:
       POSTGRES_USER: ${DATABASE_USERNAME:-postgres}
       POSTGRES_PASSWORD: ${DATABASE_PASSWORD:-password}
     ports:
-      - 5432
+      - 5432:5432
```
mononoken, line 11: Incomplete port mapping

docker-compose.yml
```diff
@@ -1,4 +1,3 @@
-version: '3.8'
```
mononoken, line 1: Version was deprecated awhile ago.

test/models/location_test.rb
```diff
@@ -2,11 +2,7 @@
 
 class LocationTest < ActiveSupport::TestCase
   context "validations" do
-    should validate_presence_of(:country).with_message("Please enter a country")
-    should validate_length_of(:country).is_at_most(50).with_message("50 characters maximum")
```
mononoken, line 6: Ugh. Shoulda matchers don't have a built-in way to handle validations with conditionals. I'll just remove these unless anyone has objections and wants to write stubs for these.

app/controllers/organizations/staff/invitations_controller.rb
kasugaijin: Yeah I agree...storing the same data in two places is not good practice. Email is the only one that we might want to do this for until we think of a good solution.

app/javascript/controllers/country_state_controller.js
kasugaijin, line 1: FYI: We currently have this issue https://github.com/rubyforgood/homeward-tails/issues/1170 to remove the city-state gem because it is unreliable in production (see link in issue). So if this gets merged before the PR for that issue I will let that author know this form is also going to need addressing.

app/models/location.rb
```diff
@@ -2,32 +2,43 @@
...
-  validates :country, presence: true, length: {maximum: 50, message: "50 characters maximum"}
-  validates :city_town, presence: true, length: {maximum: 50, message: "50 characters maximum"}
-  validates :province_state, presence: true, length: {maximum: 50, message: "50 characters maximum"}
+  with_options(
+    length: {maximum: 50, message: "50 characters maximum"}
+  ) do |location|
+    location.validates :country, presence: true, if: :province_state_present?
+    location.validates :province_state, presence: true, if: :city_town_present?
+    location.validates :city_town, presence: true
+  end
```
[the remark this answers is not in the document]
kasugaijin: I believe there is a way to do this with an attribute we set on the model class as well, e.g., foster_location, and we can pass that attribute in params from the controller, and then conditionally call validations if that attribute is true. 

e.g.,

```
attribute :location_required, :boolean

      with_options if: :location_required do
        validations...
      end
```

then in the controller add `location_required: true` to the build params passed to the model. 

I am not saying we have to do it this way...but it might read cleaner? Your call.

kasugaijin (approved):
Looks good @mononoken

app/javascript/controllers/country_state_controller.js
mononoken: Thanks for the heads up! I'll let them know I just merged this.

kasugaijin: Thanks @mononoken please point to the two forms they’ll need to address

PR #1172, opened 2024-11-21
Org Account Request - fixes

kasugaijin:
# 🔗 Issue
Found some issues with typos and markup formatting

image

# ✍️ Description

# 📷 Screenshots/Demos


app/controllers/organization_account_requests_controller.rb
```diff
@@ -3,14 +3,14 @@ class OrganizationAccountRequestsController < ApplicationController
   skip_verify_authorized
 
   def new
-    @organization_acccount_request = OrganizationAccountRequest.new
```
kasugaijin, line 6: typo in `account`

kasugaijin:
Ah test failed. I will fix later.

app/views/organization_account_requests/new.html.erb
kasugaijin, line 1: This is mostly format fixing as there were div tags missing and inconsistent use of form tags

test/system/organization_account_requests_test.rb
kasugaijin, line 1: This is a prime reason I don't like system tests most of the time. The juice is not worth the squeeze. They fail because an element in the UI changes, so require a lot of work to upkeep. We test the model and controller for this feature, and that to me is enough.

app/views/organization_account_requests/new.html.erb
```diff
@@ -1,92 +1,106 @@
...
+          <div class="card-body pt-0">
+            <div class="row mt-3">
+              <div class="col-lg-6">
                 <div class="form-group">
                   <%= form.text_field :name,
-                    label: 'Organization name',
-                    class: 'form-control',
-                    required: true
-                   %>
+                                      label: t('organization_account_requests.form.organization_name'),
```
jmilljr24, line 34: I prefer using/seeing the "Lazy" look up. I care less about it being shorter but more importantly is forces a consistent organizational structure in the en.yml file. 

This example is only a small change to the translation file but I think contributors tend to look at the code we have and try to emulate it. The translation file could get out of hand pretty quickly if people make up a different look up path for each pr.

```suggestion
                                      label: t('.form.organization_name'),
```

app/views/organization_account_requests/new.html.erb
```diff
@@ -1,92 +1,106 @@
...
+                                      class: 'form-control',
+                                      required: true
+                  %>
                 </div>
 
                 <div class="form-group">
-                  <%= form.text_field :requester_name, label: 'Your name', class: 'form-control' %>
+                  <%= form.text_field :requester_name,
+                                      label: t('organization_account_requests.form.your_name'),
```
jmilljr24, line 42:
```suggestion
                                      label: t('.form.your_name'),
```

config/locales/en.yml
```diff
@@ -53,6 +52,10 @@ en:
     new:
       header: Request an Organization Account
       description: Submit the form below if you are a pet rescue and would like your own customizable website and management platform on Homeward Tails. We'll follow up with you shortly.
+    form:
+      header: Organization Details
+      organization_name: Organization name
+      your_name: Your name
```
jmilljr24, line 58: See other comment. This would scope under `new`
```suggestion
      form:
        header: Organization Details
        organization_name: Organization name
        your_name: Your name
```

jmilljr24 (approved):
@kasugaijin Just some thoughts on i18n. No big deal.

app/views/organization_account_requests/new.html.erb
kasugaijin: Ah yes good call. I forgot about the lazy lookup. I will implement that.

config/locales/en.yml
kasugaijin: ah yea you mention that here haha

Issue #1048, opened 2024-10-09
Staff: As staff I can review an adopter's information on their application


kasugaijin:
When logged in as staff, I can go to `/adoption_application_reviews` to see all of the applications on a pet. But I cannot click on the adopter name to see their information.

Every User has a Person model. So, each adopter's information is stored on the `person.latest_form_submission.form_answers`. This will give you a list of records, each with a question snapshot and value (answer). (Nice to know, these form answers will come from a CSV uploaded by staff that matches the Person's email with a response in the CSV from a form like Google form). We need to show these questions and answers to the staff. Note, that we don't need to worry about grouping questions, for MVP we will just show a list of questions and aswers, however they come in the CSV (it's likely grouped in the original form anyway so that grouping should be loosely retained here).

To do:
- add a route and show action for `Organizations::Staff::AdoptionApplicationReviewsController` that will show the adopter's information. This information is stored as form answers on the adopter's Person model's latest_form_submission.
- authorize the route to make sure the staff is in the same organization as the adopter, and has `review_adopter_applications` permission
- add the adopter's name as a link in the table for adopter applications, and when I click on it, redirect me to the show page 
- the show page will be very simple, rendered in the staff dashboard per the sketch below
- I think the edit route/action on this controller is redundant - please check and remove it if so
- Please add some seed data so each adopter user has some form answers on their form submission

Questions and answers
![Image](https://github.com/user-attachments/assets/a2a33831-a508-40ef-a9ea-021a104d5af8)

Aaryanpal:
@kasugaijin Will Please Assign this

kasugaijin:
@Aaryanpal how's this going?

kasugaijin:
Issue up for grabs!

sarvaiyanidhi:
Hi @kasugaijin I would like to work on this issue if it is still open..

kasugaijin:
All yours @sarvaiyanidhi

sarvaiyanidhi:
Hi @kasugaijin ,

I believe that I need to upload some sample data from this location staff/external_form_upload

Do you have any sample format of CSV which can be used to upload details, basically I need to know apart from person email field what all other fields needs to be uploaded?

Not sure if required but I think we should also provide sample CSV format to download so that user can update data in it and upload it  back.

kasugaijin:
[...]

PR #1177, opened 2024-11-22
1048 - Review an adopter's information on their application

sarvaiyanidhi:
# 🔗 Issue
#1048 

# ✍️ Description
- Implemented functionality to see adopter submitted answers for its application

# 📷 Screenshots/Demos

image_2024-11-21_230410983

Alta-application-review-answers


sarvaiyanidhi:
Few points on feature implementation as per issue description

- Have implemented same UI to display application answers as FAQ page so that there is consistency in design across application
- Have not removed edit/update method code from controller as mentioned in issue as notes edit functionality on index page does not work if we remove it.

app/controllers/organizations/staff/adoption_application_reviews_controller.rb
```diff
@@ -41,6 +41,10 @@ def update
     end
   end
 
+  def show
```
kasugaijin, line 44: I think we should add the same authorization check as we do on the index. You'll just need to call it in the Show action and add an alias action in `Organizations::AdopterApplicationPolicy`

db/seeds/01_alta.rb
```diff
@@ -229,6 +229,44 @@
...
+        question_snapshot: question[:name],
+        form_submission: submission,
+        organization: @organization,
+        created_at: submission.created_at,
+        updated_at: submission.created_at
+      )
+    end
+  end
+
+  puts "Created #{FormAnswer.count} form answers"
```
kasugaijin, line 268: I think we can get rid of this log, just because we don't really do it for any other resource and there's nothing exceptional about this resource such that we need to show logs for it

test/controllers/organizations/staff/adoption_application_reviews_controller_test.rb
```diff
@@ -71,6 +71,17 @@ class Organizations::Staff::AdoptionApplicationReviewsControllerTest < ActionDis
         end
       end
     end
+
+    context "#show" do
```
kasugaijin, line 75: We are not actually authorizing requests to this route yet in the controller. I think this is passing because it's just checking if the user authorization on that policy for `manage?`, which is should pass. See my note about authorization in the controller. We just need to call `authorize!...` and add `show` to the `set_alias` method in the policy class.

test/controllers/organizations/staff/adoption_application_reviews_controller_test.rb
```diff
@@ -132,4 +143,27 @@ class Organizations::Staff::AdoptionApplicationReviewsControllerTest < ActionDis
       end
     end
   end
+
+  context "Viewing application details" do
+    setup do
+      @user = create(:admin)
+      sign_in @user
+
+      @person = create(:person)
```
kasugaijin, line 152: Here I think we can actually focus on the `User` resource. So, you can create an `adopter` (see the factory for `User` if not sure what I mean). Then, I believe that when we create a `User` record it will automatically create a `Person` (because user belongs to person). Then, you can use that `Person` record instead of creating a new one. I believe you will need to keep the rest here to create the `FormSubmission` etc.

test/integration/organizations/staff/adoption_application_reviews_test.rb
```diff
@@ -10,6 +10,12 @@ class Organizations::Staff::AdoptionApplicationReviewsTest < ActionDispatch::Int
     create(:adopter_application, status: :successful_applicant, form_submission: form_submission)
     create(:adopter_application, status: :adoption_made, form_submission: form_submission)
     @custom_page = create(:custom_page, organization: ActsAsTenant.current_tenant)
+
+    # Setup for show view tests
+    @person = create(:person)
```
kasugaijin, line 15: Same here about creating an `adopter` instead.

test/integration/organizations/staff/adoption_application_reviews_test.rb
```diff
@@ -24,6 +30,18 @@ class Organizations::Staff::AdoptionApplicationReviewsTest < ActionDispatch::Int
       follow_redirect!
       assert_equal I18n.t("errors.authorization_error"), flash[:alert]
     end
+
+    should "not be able to view application details" do
+      @user = create(:user)
+      sign_in @user
+
+      get staff_adoption_application_review_path(@adopter_application)
```
kasugaijin, line 38: I wonder if this is actually failing because of `Organizations::AdopterApplicationPolicy` `manage?` method, or in the pre-check for `verify_active_staff!`

test/integration/organizations/staff/adoption_application_reviews_test.rb
```diff
@@ -61,6 +79,40 @@ class Organizations::Staff::AdoptionApplicationReviewsTest < ActionDispatch::Int
       assert_equal("some notes", @under_review_app.notes)
     end
 
+    context "viewing application details" do
```
kasugaijin, line 82: Personally, I am no longer a fan of asserting div elements in tests because it makes the overhead of changing the UI much higher. Can we remove these ones in this integration test?

test/integration/organizations/staff/adoption_application_reviews_test.rb
```diff
@@ -74,6 +126,15 @@ class Organizations::Staff::AdoptionApplicationReviewsTest < ActionDispatch::Int
         follow_redirect!
         assert_equal "Unauthorized action.", flash[:alert]
       end
+
+      should_eventually "not be able to view application details" do
```
kasugaijin, line 130: What is this testing?

app/views/organizations/staff/adoption_application_reviews/show.html.erb
```diff
@@ -0,0 +1,24 @@
+<%= render DashboardPageComponent.new(crumb: :dashboard_applications, crumb_options: [@application]) do |c| %>
+  <% c.with_header_title do %>
```
kasugaijin, line 2: Can we update this header to be 'Form Answers'
And then underneath the breadcrumbs, we can show the person's name and email
image

Can we update this to

app/views/organizations/staff/adoption_application_reviews/show.html.erb
```diff
@@ -0,0 +1,24 @@
...
+     <!-- row -->
+    <% @form_answers.each do |form_answer| %>
+      <div class="justify-content-md-between mb-4 mb-xl-0 gx-3">
+      <!-- card -->
+      <div class="card">
+        <div id="<%= dom_id form_answer %>"class="card-body d-flex flex-sm-row flex-column justify-content-between border-bottom">
+          <div class="d-flex align-items-center">
+            <div>
+              <strong class="fs-4" >Q: <%= form_answer.question_snapshot %></strong>
+              <p class="">A: <%= form_answer.value %> </p>
```
kasugaijin, line 16: We can get rid of the margin on the bottom of this `<p>`
image

db/seeds/01_alta.rb
```diff
@@ -229,6 +229,44 @@
     end
   end
 
+  questions = [
```
kasugaijin, line 232: This is duplicated in both seed files and it's quite a bit of code. What do you think about either moving this to a single file that is used by both seeds, OR, even simpler, just use the Faker gem like you did in the tests? While the gem won't make proper questions and answers for pets, it is fine for dev and way less overhead to manage.

app/controllers/organizations/staff/adoption_application_reviews_controller.rb
sarvaiyanidhi: Done.. Added authorization for the show action

db/seeds/01_alta.rb
sarvaiyanidhi: Done

test/controllers/organizations/staff/adoption_application_reviews_controller_test.rb
sarvaiyanidhi: Have updated show method with authorize code and added alias in policy class

test/controllers/organizations/staff/adoption_application_reviews_controller_test.rb
sarvaiyanidhi: Yes, I checked the User factory and found that adopter already creates form submission so updated code accordingly.

test/integration/organizations/staff/adoption_application_reviews_test.rb
sarvaiyanidhi: Done

test/integration/organizations/staff/adoption_application_reviews_test.rb
sarvaiyanidhi: Have removed assert select from this as well as controller test to remove dependency on UI

test/integration/organizations/staff/adoption_application_reviews_test.rb
sarvaiyanidhi: I tried to replicate the way it was done for index page but it was not actually working as expected so removed it. Infact existing test for index also uses should_eventually which is not actually passing but just skipping it. I think we can removed deactivated staff block but have kept old code as it is for now.

app/views/organizations/staff/adoption_application_reviews/show.html.erb
sarvaiyanidhi: Is this better?

Alta-application-review-answers

app/views/organizations/staff/adoption_application_reviews/show.html.erb
sarvaiyanidhi: Done

db/seeds/01_alta.rb
sarvaiyanidhi: Populated questions and answers using Faker to simplify code

app/views/organizations/staff/adoption_application_reviews/show.html.erb
kasugaijin: @sarvaiyanidhi can we put the email below the user's name and make it a hyperlink?

Joe Brando
[joe@brando.com](joe@brando.com)

test/integration/organizations/staff/adoption_application_reviews_test.rb
kasugaijin: Thank you!

app/views/organizations/staff/adoption_application_reviews/show.html.erb
sarvaiyanidhi: Done

Alta-application-review-answers

Issue #1182, opened 2024-11-24
Set up BugSnag


mononoken:
See Slack for credentials

PR #1184, opened 2024-11-25
1182 setup bugsnag

mononoken:
# 🔗 Issue

#1182

# ✍️ Description

This should setup BugSnag to log errors on prod. I added a little extra configuration beyond just installation. 

Would be good to test this out. We can do so calling `Bugsnag.notify("test error")` in prod and see if it shows up on the dashboard.

Here is the info that should automatically be captured: https://docs.bugsnag.com/platforms/ruby/rails/automatically-captured-data/. We may want to modify some of it. For example, it captures some user info automatically. Any privacy concerns with any of that? Especially if we do decide to surface some errors in Slack, we will want to consider sanitizing this. I think organization context should be included since we include the slug in the url.

Here are the BugSnag Rails docs: https://docs.bugsnag.com/platforms/ruby/rails/

# 📷 Screenshots/Demos

n/a


config/initializers/bugsnag.rb
```diff
@@ -0,0 +1,7 @@
+if Rails.application.credentials.bugsnag_api_key.present?
+  Bugsnag.configure do |config|
+    config.api_key = Rails.application.credentials.bugsnag_api_key
+  end
+else
+  Bugsnag.configuration.logger = ::Logger.new("/dev/null")
```
mononoken, line 6: If no key, send to abyss.

app/controllers/application_controller.rb
```diff
@@ -14,9 +21,11 @@ def switch_locale(&action)
...
-  rescue_from ActionPolicy::Unauthorized do |ex|
-    flash[:alert] = t("errors.authorization_error")
-
-    redirect_back fallback_location: root_path
+  def log_and_reraise(error)
+    unless KNOWN_ERRORS.include?(error.class)
+      Bugsnag.notify(error)
+      error.instance_eval { def skip_bugsnag = true }
+    end
+    raise
```
mononoken, line 29: Default behavior of BugSnag is to log unhandled errors. This config will also log handled errors, except authz errors. It also should avoid logging unhandled errors twice.

If there are more handled errors we don't want to log, we could add them to the const.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,17 @@
...
-
   before_action :set_current_user
   around_action :switch_locale
 
+  KNOWN_ERRORS = [ActionPolicy::Unauthorized]
+  rescue_from StandardError, with: :log_and_reraise
+  rescue_from ActionPolicy::Unauthorized do |ex|
+    flash[:alert] = t("errors.authorization_error")
+
+    redirect_back fallback_location: root_path
```
jmilljr24, line 12: This is [soft deprecated](https://api.rubyonrails.org/v7.1/classes/ActionController/Redirecting.html#method-i-redirect_back) fyi

```suggestion
    redirect_back_or_to root_path
```

jmilljr24 (approved):
@mononoken lgtm. Just an fyi

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,17 @@
 class ApplicationController < ActionController::Base
   before_action :authenticate_user!
   verify_authorized unless: :devise_controller?
-
   before_action :set_current_user
   around_action :switch_locale
 
+  KNOWN_ERRORS = [ActionPolicy::Unauthorized]
```
kasugaijin, line 7: Should we put all error logging logic in a concern to include on application controller? Just keeps it in once place, out of the way of domain logic that might need to live on application controller.

config/initializers/bugsnag.rb
kasugaijin: Do we want to scope this to an environment as well? e.g. `if env.production? then...`

kasugaijin:
> For example, it captures some user info automatically. Any privacy concerns with any of that? Especially if we do decide to surface some errors in Slack, we will want to consider sanitizing this.

As a matter of best practice we should filter certain data from all logs. Ideally we should set up an initializer that does this:

```
Rails.application.config.filter_parameters += [
  :password,
  :password_confirmation,
  :credit_card_number,
  :ssn,
  :api_key,
]
```

app/controllers/application_controller.rb
mononoken: I would rather not in this case. The code isn't going to get reused anywhere else. And the order of many of the actions in ApplicationController are dependent on each other. The controller is small now too.

config/initializers/bugsnag.rb
mononoken: This is done by the API key call. This will only load if the API key is present, which it should only be for prod.

kasugaijin: Can we add a comment for that then? I don’t think that would be immediately obvious to a future dev reading this without some nuanced understanding.

mononoken: In meeting today we discussed and thought this would be okay without comment. It is agnostic of the env and is relying on the credentials being available. Think it should be able to follow without comment. We can add one in the future though if we have any confusions.

Issue #1153, opened 2024-11-17
Investigate: why bumping active storate validations gem causes tests to fail


kasugaijin:
https://github.com/rubyforgood/homeward-tails/pull/1145 in this PR made by Dependabot, it bumps Active storage validations gem and the tests fail. Please find out why. 

Active storage validations is used to...validate active storage attachments e.g., file formats and size. I'd start with checking our that PR and running the tests locally and start debugging. I suspect it's something silly and small (famous last words).

PR #1197, opened 2024-11-29
Dependabot update fail fix

MooseCowBear:
# 🔗 Issue

#1153 
# ✍️ Description

Super simple fix to get the test to pass but something we might want to think about is the why. 

This test started to fail because of updates to the gem that were supposed to make the content type check not rely on the filename in order to determine whether the file was the right kind of file because that is an unreliable way to check for content type. However, when no content type is provided (as is the case in this test), the way the gem gets the content type is not by looking at the content type property but by looking at the supplied filename. And since the filename passed to the test ended in `png` the validation passed. 

The readme now suggests turning on spoofing protection (which will become the default in version 2) which looks like it will check the contents of the file to determine content type, but I haven't been able to get that to work without breaking a bunch of other tests. Anyway, if we want that feature, it's something to investigate. 

# 📷 Screenshots/Demos


jmilljr24 (approved):
@MooseCowBear Nice! 

If you want, you can make an issue for the changes mentioned. I would just tag is "on hold" if you think there is more to be discussed.

kasugaijin (approved):
Ah nice work.

Issue #1192, opened 2024-11-27
UI Mobile: How to apply for adoption overflow


jmilljr24:
On the Organization home page the steps on how to apply do not wrap. Fix for screen sizes below [Bootstrap medium](https://getbootstrap.com/docs/5.3/layout/breakpoints/#available-breakpoints) < 768px. Size small screens could possibly be displayed 2 by 2 grid. Extra small will need to be vertical. The icon size may make this quite long to scroll through on mobile so feel free to adjust and see what works best.

![Screenshot from 2024-11-27 11-26-36](https://github.com/user-attachments/assets/2b5385cd-208b-4855-bc44-65632836efc0)

Aaryanpal:
@jmilljr24 this too

jmilljr24:
@Aaryanpal have at it!

PR #1202, opened 2024-11-30
#1192 fix the ui for mobile

Aaryanpal:
# 🔗 Issue
#1192 

# ✍️ Description
Introduce Grid for working in mobile tablet and desktop
# 📷 Screenshots/Demos

[as.webm](https://github.com/user-attachments/assets/f8e613c0-e544-4b75-8702-44da9e63f91d)

### DIFF
--- a/app/views/organizations/home/index.html.erb
+++ b/app/views/organizations/home/index.html.erb
@@ -66,66 +66,56 @@
 </section>
 
 <!--//? how to apply section ?//-->
-<section id="how-to-apply">
-  <div class="container">
-
+<section class="pt-5 pb-5" id="how-to-apply">
+  <div class="container mb-md-5 mt-md-5">
     <!--//? section header -->
-    <div class="d-flex justify-content-start mb-2">
+    <div class="text-center mb-5">
       <h2 class="section-heading text-uppercase text-xl">
         <%= t('.how_to_apply') %>
       </h2>
     </div>
-
     <!--//? cards section - info (not bootstrap card)  -->
-    <div class="d-flex flex-wrap justify-content-around align-items-center px-4">
-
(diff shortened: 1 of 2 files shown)


jmilljr24 (approved):
@Aaryanpal Thanks. The 2x2 grid looks great. I kinda figured the single column would look a little long but this is better than it overflowing. I'm the wrong person for UI design but if you have have any idea to make it look better feel free. Maybe even just some smaller icons when it is a single column. If you want to leave it as is, you can merge this. 

One other nit, was there a reason the list items didn't work for you? My thought was the list items numbers will be translated via the browser for i18n without us having to do that manually for each language. Obviously the lines will still need translating either way but why repeat something that already has a built in feature. Does that make sense?

jmilljr24:
Resolves #1192

Aaryanpal:
> @Aaryanpal Thanks. The 2x2 grid looks great. I kinda figured the single column would look a little long but this is better than it overflowing. I'm the wrong person for UI design but if you have have any idea to make it look better feel free. Maybe even just some smaller icons when it is a single column. If you want to leave it as is, you can merge this.
> 
> One other nit, was there a reason the list items didn't work for you? My thought was the list items numbers will be translated via the browser for i18n without us having to do that manually for each language. Obviously the lines will still need translating either way but why repeat something that already has a built in feature. Does that make sense?


Thank you for the feedback, much appreciated! 😊

I opted to structure the file this way to maintain uniformity across the codebase and ensure everything aligns with our existing patterns. I totally understand your point about leveraging the browser's built-in i18n for numbering—it’s a great feature and could simplify translations. My approach here was primarily about consistency, but I’m open to revisiting it if you feel the native list functionality is a better fit.

Let me know your thoughts, and I can adjust accordingly. Also, I’ll give some thought to UI tweaks, like the smaller icons, and see if I can refine it further. Thanks again for your input!

PR #1208, opened 2024-12-02
External Form Upload: Display import summary / errors 

khiga8:
# 🔗 Issue

https://github.com/rubyforgood/homeward-tails/issues/1112

# ✍️ Description

This PR displays import summary feedback when a file is uploaded through the external form upload.

* I opted to display a visually prominent banner right below the attach control, rather than top of the page. This banner is injected via turbo (following rec in https://github.com/rubyforgood/homeward-tails/issues/1112), and the page doesn't reload, so I believe placing it right by the control that the user just interacted with would make it more visible. 
* A red banner is shown for an error state, and a green banner is shown for a success state. 
* I set a `role="alert"` on the heading of the dynamically injected content, so that assistive technology users, such as screen reader users, are made aware of this new content via announcement when it's added to the page.

# 📷 Screenshots/Demos

Screenshot of green success banner placed below the upload control with the heading, 'File successfully scanned'

Screenshot of red error banner placed below the upload control, with the heading, 'File scanned: 1 error(s) present'


app/services/organizations/importers/google_csv_import_service.rb
```diff
@@ -13,6 +13,7 @@ def initialize(file)
       end
 
       def call
+        # Validate that file is in CSV format before we read from it.
```
jmilljr24, line 16: @khiga8 Thanks for pointing this out! This started a discussion in our last meeting about other validations we should be checking for in this service. I opened PR #1211 which adds a file type validation and checks for the correct headers.

Hopefully I didn't step on your toes, but I figured the validations were outside the scope of this issues and didn't want to cause any confusion or add more work. If you have a better way to validate though I'd love to see it!!

The controller

khiga8: This is great! Thank you :)

Issue #1170, opened 2024-11-21
Implement Country/State select ability


kasugaijin:
We currently use city-state gem to get a list of states for a given selected country. However, this gem is not reliable in production and this [thread](https://github.com/thecodecrate/city-state/issues/3) does not inspire confidence in the long run. So, we will keep things simple and roll our own. 

This is a two-part issue

Part 1)
- remove the city-state gem
- remove the routes for countries and states
- remove the states controller and its associated view
- remove states tests

Part 2)
- we only use the country and state form fields in the Organization Settings form
- update this select so countries options are Canada, USA, Mexico (only these for now) 
- states - use stimulus to dynamically update the state select to only show the values for the states of the selected country. So, if I select USA, it updates the State select to show USA states.
- States should be displayed with their full name e.g. California, but saved with their abbreviateion CA

Tip: To get a full list of the states for each country, use the city-state gem in your rails console to get the values i.e., `CS.states(:US)` will give you the states for USA. Then you can use these in Stimulus.

sarvaiyanidhi:
@kasugaijin I can take up this issue.. Can you please assign it to me?

kasugaijin:
@sarvaiyanidhi all yours!

mononoken:
FYI, @sarvaiyanidhi:

Just merged #1169 which added some city-state usage in the fosterer edit form and fosterer invite form. It also touched the `country_state_controller.js` a little. Hopefully, it's not too much work for you. Thank you!

PR #1209, opened 2024-12-03
1170 - Country State Select Implementation

sarvaiyanidhi:
# 🔗 Issue
#1170 

# ✍️ Description
- Remove existing city-state gem implementation
- Custom code implementation of country and state using Stimulus

# 📷 Screenshots/Demos


kasugaijin:
@sarvaiyanidhi how's this going? Let me know if you have any questions!

sarvaiyanidhi:
> @sarvaiyanidhi how's this going? Let me know if you have any questions!

Hey @kasugaijin Almost 80% is done, just facing some issue with displaying selected state although it is saving correctly in database and need to update test cases. Will push code soon with all changes.

kasugaijin:
OK thanks for the update @sarvaiyanidhi !

sarvaiyanidhi:
Hi @kasugaijin Just an update I could not spend much time in this week to close on this.. Will try to work on this over the weekend and close it. Thanks

kasugaijin:
@sarvaiyanidhi no worries thank you!

sarvaiyanidhi:
Hi @kasugaijin ,

I have pushed code for review. Following are the changes done to implement feature
- Removed dependency of country-state gem
- Loaded required countries and states through YML file in initializer code so that in future if we want to extend with more countries it can be easily done
- Other code is almost same as previous implementation except for one attribute added for state in JS controller file as it was not retaining value in droplist in earlier code as well.

Let me know if you need any other changes.. Thanks

app/javascript/controllers/country_state_controller.js
kasugaijin, line 1: New line :)

config/countries_states.yml
kasugaijin, line 1: New line :)

app/views/organizations/staff/fosterer_invitations/_form.html.erb
kasugaijin, line 1: These country and province state inputs are identical, aren't they? How about we put them in a partial that we can pass the form instance to? The partial could live somewhere like `app/views/shared/location/_country.html.erb` and `app/views/shared/location/_province_state.html.erb`

sarvaiyanidhi: We can put in partial and use that in form but I thought it would be too much optimization as one has to refer two additional files just to go through one form code but on other hand if we want to change anything for country and state it will be easy to just do in one file.

Should I go ahead and create partials?

kasugaijin: Yeah I think that given this is a custom built feature it's nice to have it in one place and each form just renders those partials and passes the form context. It's a more maintainable approach IMO.

kasugaijin: Maybe just go with the one partial for now. These inputs will always be used together, so they can live in one file.

kasugaijin (commented):
@sarvaiyanidhi please also update the Org Account Request form to use this new implementation `/organization_account_request/new`

image

kasugaijin (approved):
Awesome thank you @sarvaiyanidhi

PR #1211, opened 2024-12-05
Add CSV Import File Validations

### LINKED ISSUE #1210: Add custom error for email column missing in csv import
We want to add a custom error that gets raised in the import service to alert the user that they need the "email" column in their csv.

jmilljr24:
Resolves #1210 

# ✍️ Description

This adds a validation check within the csv import service. The file type is checked and the headers are checked for the column named "Email". We still have the issue where an organization could name their "Email" column something different like "Email Address" which would produce an error.

# 📷 Screenshots/Demos


kasugaijin:
>We still have the issue where an organization could name their "Email" column something different like "Email Address" which would produce an error.

I think it's normal to have requirements on headers for CSV uploads. So, we could make a note on the upload page that there must be an `email` header

kasugaijin (commented):
What do you think about moving the validation logic to its own single responsibility? We can use these validations for file type and email header for CSV imports from all third party services. Maybe we can keep it in `app/services/organizations/importers/concerns`?

jmilljr24:
> What do you think about moving the validation logic to its own single responsibility? We can use these validations for file type and email header for CSV imports from all third party services. Maybe we can keep it in `app/services/organizations/importers/concerns`?

That does make sense if we have a separate import service for each third party service. I'm not sure if that is ultimately what we will need to do. There is a good chance this service could handle many different third party csv's without being overly complex. And we wouldn't need to repeat a lot of code. I'm hopeful it will be as simple as handling the "email" column and potentially the "Timestamp" column as those two are what the service depends on. The rest of the columns just get added as question/answer. I can't imagine any other things that would cause an issues from one form type to another.

If we can keep one clean and simple service, I don't see the need to move the validation logic. If you think it is more organized to move to a concern even with only one import service, I'm happy to do it.

kasugaijin:
Ah yes…if we only need one service for all third party forms it makes sense to keep as is!

app/services/organizations/importers/google_csv_import_service.rb
```diff
@@ -13,42 +13,57 @@ def initialize(file)
...
       end
 
       private
 
+      def validate_file
+        raise FileTypeError unless @file.content_type == "text/csv"
+
+        first_row = CSV.foreach(@file.to_path).first
+        raise EmailColumnError if first_row.nil?
+        raise EmailColumnError unless first_row.include?("Email")
```
kasugaijin, line 61: shall we add lowercase `email` here just in case the user doesn't adhere to case requirement?

app/services/organizations/importers/google_csv_import_service.rb
```diff
@@ -13,42 +13,57 @@ def initialize(file)
...
         Status.new(@errors.empty?, @count, @no_match, @errors)
       end
 
       private
 
+      def validate_file
+        raise FileTypeError unless @file.content_type == "text/csv"
+
+        first_row = CSV.foreach(@file.to_path).first
+        raise EmailColumnError if first_row.nil?
```
kasugaijin, line 60: If I am understanding this correctly, we are raising `EmailColumnError` if the first row is empty, right? Should we have a different error class to handle this case? Something like `NoDataError` (I am sure there is a better name than that).

app/services/organizations/importers/google_csv_import_service.rb
kasugaijin, line 1: Nice work! I like the structure.

app/services/organizations/importers/google_csv_import_service.rb
```diff
@@ -13,42 +13,57 @@ def initialize(file)
...
           end
-        rescue => e
-          @errors << [index, e]
         end
         Status.new(@errors.empty?, @count, @no_match, @errors)
       end
 
       private
 
+      def validate_file
```
kasugaijin, line 56: In the vein of single responsibility, we could have separate validation methods for each validation. Might be something to consider if we add more validations. I am fine either way in this case.

app/services/organizations/importers/google_csv_import_service.rb
jmilljr24: I think I'm confused on what you are asking.  If the header row doesn't include "Email"(capitalized) the error is raised. The service is currently using "Email". Do you want "email" to work as well?

kasugaijin: Sorry yeh that’s right. We may as well allow Email and email as valid because users won’t adhere to case rules half the time.

app/services/organizations/importers/google_csv_import_service.rb
jmilljr24: Yes it is checking for an empty file as an edge case (came up during testing). I just reused the same error because technically the file does not have the "Email" column but I can add a more specific error.

app/services/organizations/importers/google_csv_import_service.rb
```diff
@@ -13,42 +13,62 @@ def initialize(file)
...
+        raise FileTypeError unless @file.content_type == "text/csv"
+
+        first_row = CSV.foreach(@file.to_path).first
+        raise FileEmptyError if first_row.nil?
+
+        email_headers = ["Email", "email", "Email Address", "email address"]
+        email_headers.each do |e|
+          @email_header = e if first_row.include?(e)
+        end
```
jmilljr24, line 65: This is the simplest way I can think of at the moment to allow both "Email" and "email". Since I had to do this change anyways, it is easy enough to add new naming variation's so I added the stakeholders as well.

There isn't any extra logic to deal with the scenario of multiple headers having an email name.

kasugaijin (commented):
Another thought, do we need to also validate the timestamp header the same way we do for email?

app/services/organizations/importers/google_csv_import_service.rb
kasugaijin: Yeah I like this.

jmilljr24:
> Another thought, do we need to also validate the timestamp header the same way we do for email?

Good point. It is the default for google but someone could alter their csv and we should handle that.

kasugaijin (approved):
Looks great! I think we should update the CSV import page to state those two headers need to be present for it to work. I will add that in a PR I am making with small fixes here and there.

PR #1212, opened 2024-12-06
remove /

kasugaijin:
# 🔗 Issue
Fixes the URLs generated in the mailers. They have an extra `/` before the slug...which breaks the URL.

# 📷 Screenshots/Demos


jmilljr24 (approved):
@kasugaijin looks good.

PR #1233, opened 2024-12-09
Remove adoptable pets policy

kasugaijin:
# ✍️ Description
There was a bug - if I created a published pet, it would show on the adoptable_pets page. If I created a match for that pet that was a foster, the pet would disappear from the index page. This was because of the policy removing any pets with matches. It got me looking at the policy, and I think it is superfluous. It is a totally public route, and we don't need a policy class to determine what pets we show on that page. We can handle that in the controller based on the query we make. I prefer we get rid of the policy class (and any tests). This also fixes the aforementioned bug.

# 📷 Screenshots/Demos


app/controllers/organizations/adoptable_pets_controller.rb
```diff
@@ -55,6 +51,8 @@ def set_likes
 
     def set_species
       @species = params[:species]
+
+      redirect_back_or_to root_path if @species.nil?
```
kasugaijin, line 55: I prefer to keep the action method happy path only, so moving this here

app/controllers/organizations/adoptable_pets_controller.rb
kasugaijin, line 1: I don't think there's any need to have the ability for staff to see the show page of an adoptable pet that has been adopted, or is not published. They can manage pets via the dash scoped routes just fine.

We could keep a policy class, technically, but it doesn't make much sense when we can just handle the desired behaviour simply within the controller. It's less overhead, and easier to digest.

test/integration/adoptable_pet_show_test.rb
```diff
@@ -28,6 +29,12 @@ class AdoptablePetShowTest < ActionDispatch::IntegrationTest
       assert_response :success
     end
 
+    should "see a pet with a current foster" do
```
kasugaijin, line 32: This scenario is currently not possible in Main because the policy is rejecting pets with **any** matches, including those that are fosters. So, this is a needed fix.

app/views/organizations/adoptable_pets/show.html.erb
kasugaijin, line 1: image

jmilljr24 (approved):
@kasugaijin lgtm!

app/controllers/organizations/adoptable_pets_controller.rb
kasugaijin: Yeah it took me a minute to get used to this approach but now I really like it. Keeping the action happy path only and handling edges in other methods. It's a way to keep controllers very clean. I would counter and say that it's really not two different responsibilities. The method is setting the species, and handling the event that it is nil accordingly.

app/controllers/organizations/adoptable_pets_controller.rb
kasugaijin: I hear that. I thought of this, too. I ended up taking the simplest approach possible, because it should be enough to work as needed. And, generally, we should always write the simplest code we can to achieve a goal (I think). In this case, we are assuming this is a useful feature, but it might not be. I'd rather we deploy the simple path, and be told it needs changing, than assume it's something that will be used, and have the potential overhead of addtional logic in a policy class. 

Futher, I think it's more efficient for us to query what we need, rather than query a larger collection, and let the policy dictate what data we should use.

I think the policy classes are awesome, but they don't have to be used in every case. The controller can also be used to handle behaviours/flows.

kasugaijin: @mononoken I am going to merge now as I want to fix this bug in production! If we want to revisit this policy class let's do that on one of our calls!

Issue #1056, opened 2024-10-10
Clicking pet when viewing adoption applications as adopter fosterer leads to unauthorized error


coalest:
### Reproduction steps:
1. Log in as an adopter, eg. `adopter1@alta.com:123456`
2. Go to the `Adoption Applications`page
3. Click on a pet's name.
4. See the error message like screenshot below:
![Screenshot from 2024-10-10 20-27-17](https://github.com/user-attachments/assets/4bfbf45a-0400-4dd3-abb4-65e47de3b71f)

### Expected behavior:
In theory I have already put in an application to adopt these pets, so I would expect to be directed to a page with information about that pet (either generic information or information on the application I made).

### Actual behavior:
I see an error message flash that says "You are not authorized to perform this action."
In the logs, I see a request was sent to `GET /alta/adoptable_pets/1` failed the policy check and resulted in a 302 redirect.
I believe because [this part of the expression](https://github.com/rubyforgood/pet-rescue/blob/main/app/policies/organizations/adoptable_pet_policy.rb#L11) returns `false` as the pet does have a match (the current user).

princekumarg12:
Hi,
I would like to take on the task regarding the authorization issue with the adoption application page. I believe I can help identify and resolve the underlying policy check problem.

Please let me know if I can be assigned to this task.

kasugaijin:
@princekumarg12 all yours!

princekumarg12:
@kasugaijin i found this while debugging

There are two reasons for this bug:

1. user.staff_account is nil, which causes authorization to be denied.
See the screenshot for reference.
![Screenshot from 2024-10-13 00-28-54](https://github.com/user-attachments/assets/95119c04-0f7d-41f9-8a31-b4f7e8117825)

2. The user does not have permission to "manage_pets."

kasugaijin:
Ah good catch! So we have this PR https://github.com/rubyforgood/pet-rescue/pull/1039 to remove the StaffAccoubt model. Can you please check out that branch and see if the bug is fixed? It might not need fixing in that case.

princekumarg12:
I checked out the branch 1034-replace-staffaccount and found that the issue still persists.

kasugaijin:
Ah ok thank you! I think it would make sense to hang on until that branch https://github.com/rubyforgood/pet-rescue/pull/1039 is merged because we are changing the architecture around a bit, and I think that would impact this fix.

kasugaijin:
[...]

PR #1234, opened 2024-12-09
1056 disable adopted pets link

fivefootbot:
# 🔗 Issue
http://localhost:3000/alta/adopter_fosterer/applications

# ✍️ Description
I disabled the link to a pet's profile if the adoption has been made unless the user is a staff member. 

# 📷 Screenshots/Demos
Screenshot 2024-12-09 at 6 24 22 PM


app/assets/images/about1.jpg
kasugaijin, line 1: Why are all of these assets in this PR?

test/helpers/adopter_applications_helper_test.rb
kasugaijin, line 1: Same here we don't need to handle the staff context

kasugaijin (commented):
Was this linked to a specific issue? I don't think it's a major issue if a person clicks to see the adoptable pet path of an adopted pet. I believe we show a status of adopted on that page? But I cannot recall if this is an issue, or something you considered should be added? I guess it would be cleaner to prevent the click through to that page, but then I would just check if the `pet` instance `is_adopted?` and then handle accordingly. I don't think we need a helper here.

jmilljr24:
@kasugaijin #1056

kasugaijin:
> @kasugaijin #1056 This might be resolved by you removing the policy in #1233

Yes you are right because it redirects if the pet is adopted or not published. I think we can actually close this PR @fivefootbot

jmilljr24:
@kasugaijin #1056 


> > @kasugaijin #1056 This might be resolved by you removing the policy in #1233
> 
> Yes you are right because it redirects if the pet is adopted or not published. I think we can actually close this PR @fivefootbot

@kasugaijin I don't think its a bad idea to still disable the link for adopted pets. It will just redirect back as is but not the best to have a link for no reason. I can do the review for this if you want. Should just need `link_to_unless pet.is_adopted?, pet.name, adoptable_pet_path(pet))  `

kasugaijin:
> @kasugaijin #1056
> 
> > > @kasugaijin #1056 This might be resolved by you removing the policy in #1233
> > 
> > 
> > Yes you are right because it redirects if the pet is adopted or not published. I think we can actually close this PR @fivefootbot
> 
> @kasugaijin I don't think its a bad idea to still disable the link for adopted pets. It will just redirect back as is but not the best to have a link for no reason. I can do the review for this if you want. Should just need `link_to_unless pet.is_adopted?, pet.name, adoptable_pet_path(pet)) `

Yes you’re right it’s better ux to not link the pet name if the pet is adopted! Disregard my comment above this still adds value here :)

app/views/organizations/adopter_fosterer/adopter_applications/_applications_table.html.erb
```diff
@@ -24,7 +24,7 @@
               'coming_soon.jpg', class: 'rounded-3', style: "height: 50px; width: 50px; object-fit: cover;") %>
               <div class="ms-3">
                 <h4 class="mb-0">
-                  <%= link_to pet.name, adoptable_pet_path(pet) %>
+                  <%= link_to_if(link_to_application?(app, current_user), pet.name, adoptable_pet_path(pet)) %>
```
jmilljr24, line 27:
```suggestion
                  <%= link_to_unless pet.is_adopted?, pet.name, adoptable_pet_path(pet)) %> 
```

jmilljr24 (changes requested):
@fivefootbot Thanks for working on this! Sorry if there was any confusion.

In `app/models/pet.rb` we have the method `is_adopted?` which will simplify this a bit. I like the `link_to_if` you used. I didn't know that existed :smiley: 

So the only two files with any changes that should be in this PR are applications_table and application_cards. If you can revert the other files that would be great. No new test required for this PR. If that causes any troubles for you feel free to start with a new branch off main and create a new PR. No big deal. 

Side note: When you create a pull request, where it says 🔗 Issue, you can type `#1056` with the original issue number and it will link it automatically. This makes it easier for anyone reviewing the PR to find the issue. Also you will note there is a failed check for "Ruby on Rails CI / Standard (pull_request) ". After you make your changes locally, run `bin/rails standard:fix`

Let us know if you have any questions here or in slack!!

app/views/organizations/adopter_fosterer/adopter_applications/_applications_table.html.erb
kasugaijin: TIL `link_to_if` and `link_to_unless` exist!

app/assets/images/about1.jpg
fivefootbot: I think I accidentally branched off of a previous branch. I'll get this cleaned up.

jmilljr24:
#1247

PR #1241, opened 2024-12-11
Rails 8

jmilljr24:
# 🔗 Issue

# ✍️ Description

 Upgrade to Rails 8. I attempted to keep any new rails 8 defaults unless we had something configured differently. I also changed the annotate gem to annotaterb as it supports rails 8 and the previous seems to be abandoned. 

I did some user flows and fixed anything minor that I came across. Please verify the docker development setup works.

# 📷 Screenshots/Demos


bin/setup
```diff
@@ -14,20 +13,22 @@ FileUtils.chdir APP_ROOT do
   # Add necessary setup steps to this file.
 
   puts "== Installing dependencies =="
-  system! "gem install bundler --conservative"
```
jmilljr24, line 17: This is not in the new default setup file. Was this something for the docker setup?

config/application.rb
```diff
@@ -20,6 +20,10 @@ class Application < Rails::Application
     #
     # Added to manage the tenants within the path
     config.middleware.use OrganizationMiddleware
+    # Please, add to the `ignore` list any other `lib` subdirectories that do
+    # not contain `.rb` files, or that should not be reloaded or eager loaded.
+    # Common ones are `templates`, `generators`, or `middleware`, for example.
+    config.autoload_lib(ignore: %w[assets tasks])
```
jmilljr24, line 26: Do we need to add anything to this?

bin/setup
kasugaijin: That came from the old baja pet rescue app, and I cannot recall why. I probably followed a blog on setting up the rails app at that time, and that was a noted step, if it wasn't included in Rails 7 by default.

app/controllers/static_pages_controller.rb
kasugaijin, line 1: How come this action is being added here? We have an `Organizations::FaqController` (which I think can easily be moved to this staic pages controller) but we'd want to do all or nothing here?

app/controllers/organizations/staff/custom_form/questions_controller.rb
kasugaijin, line 1: Why is this action being added?

app/controllers/static_pages_controller.rb
jmilljr24: @kasugaijin Below was added as a default config. It raises an error because `faq` is in `skip_verfiy_authorized`. I didn't look into why faq is included (if you know off the top of your head why please tell). Same with questions controller. I haven't had a chance to look into it.

 We can remove the config if it's something we are not concerned about.
```
  # Raise error when a before_action's only/except options reference missing actions.
  config.action_controller.raise_on_missing_callback_actions = true
```

jmilljr24: I went ahead and removed faq from `skip_verify_authorized` as it was not being used in this controller. Moving faq to the static pages controller could be a separate issue if you want.

app/controllers/organizations/staff/custom_form/questions_controller.rb
jmilljr24: Looks like show wasn't need in the before action. There is no show page at this time.

bin/setup
jmilljr24: Ok. I'll leave it out for now.

kasugaijin (approved):
Awesome! Thanks for taking care of this @jmilljr24 
I will let you merge this. If you have time, can you please do a manual deploy and check nothing explodes in production?

PR #1244, opened 2024-12-11
Form submission refactor

kasugaijin:
# 🔗 Issue
Issue: https://github.com/rubyforgood/homeward-tails/issues/1232

- changes associations so that adopter application belongs to person instead of form submission
- removes broadcast on adopter applications
- Adds a new form submission route scoped to the person to display their form submissions. User can click a given form submission to get its associated forms answers via turbo frame, from a new form submission scope route.
- Adds policy class for these two resources
- updates tests to handle new associations

We could consider pagination. I was thinking it would be useful to paginate the form submissions so we only ever have 3-5 rows in that table. But, I think it's better UI not to paginate form answers because if you want to go back to a given question/answer while reviewing details it will be annoying to do it with page links. I think it's easier to just scroll. Open to other thoughts on this, though. I suggest adding pagination in another issue.

formstuff

Empty State
image

# ✍️ Description

# 📷 Screenshots/Demos


app/models/adopter_application.rb
```diff
@@ -2,38 +2,33 @@
...
   acts_as_tenant(:organization)
   belongs_to :pet, touch: true
-  belongs_to :form_submission
+  belongs_to :person
 
-  has_one :person, through: :form_submission
-
-  validates :pet_id, uniqueness: {scope: :form_submission_id}
-
-  broadcasts_refreshes
```
kasugaijin, line 36: This was causing issues in production - we are not set up to use this pub sub so will remove.

jmilljr24: Generally I would agree to remove it.  I'm wondering if it would be okay in this instance to comment it out with the intention to implement soon with rails 8 action cable.

kasugaijin: It’s pretty easy to add back in. I’d rather remove it until we actually implement actioncable otherwise we could have commented out code sitting for a while.

app/controllers/organizations/staff/form_answers_controller.rb
kasugaijin, line 1: This is an endpoint we hit with turbo frame to load the form answers. I figured the policy is exactly the same as FormSubmission, so re-used that policy instead of creating a separate policy.

app/controllers/organizations/staff/form_submissions_controller.rb
```diff
@@ -0,0 +1,25 @@
+module Organizations
+  module Staff
+    class FormSubmissionsController < Organizations::BaseController
+      layout "dashboard"
+
+      before_action :context_authorize!, only: %i[index]
+      before_action :set_person
+
+      def index
+        @form_submissions = @person.form_submissions.order(created_at: :desc)
```
jmilljr24, line 10: `created_at` will be different than `csv_timestamp`. Do we want to use the csv timestamp? 

We will have to deal with form submissions without the csv timestamp(custom forms) as well.

kasugaijin: Yes that makes sense!

kasugaijin: Second thoughts, I suppose it depends on how we want to define the FormSubmission. Is it a capture of when they filled out the Google Form, or when the staff imported the data?

jmilljr24: I'm not sure how valuable the date of import is. For example, when a new org does their first import they may have multiple submissions from one person on separate dates. The import date would be the same day for all submissions.

When our custom forms is fully functional, then yes the date the `form submission` is created would make sense because that is when the form is submitted.

kasugaijin: Ok if we go with the csv_timestamp, we have to assume it will always be available on FormSubmission, at least until we start using custom forms. We do not validate this attribute on the model or database. I think that now we create the FormSubmission in the CSV Import service, we should at least have a validation for csv_timestamp. When we start to use the custom forms, we can update this validation to only run on CSV imports. Thoughts?

kasugaijin: I updated to show both the csv_timestamp and the import date just so it's clear what each date is for. I also decided to show the HHMM for the csv_timestamp in case someone submits twice in one day - then the staff can tell which set of answers they are checking.

image

jmilljr24: Looks good! We can always add a user sort-able feature in the future.  FYI this isn't a blocker for me, I just haven't had a chance to fully review the whole pr yet.

My first thought on validating the csv_timestamp presence is that it makes sense. After thinking through the import more I'm not sure. I'm trying not to get too into the weeds on what people might do to their csv date, but as of right now the the import will work just fine if the timestamp column entry is nil (the header has to be present). So for example, the row could have valid data and a matching email but no timestamp ( maybe the row was added manually by someone?) I don't think I'd want to skip that row.

Maybe a simpler approach is to conditionally display the timestamp for each submission in the view. That resolves the current issue as well as when the custom form is live.

jmilljr24 (approved):
@kasugaijin Awesome!!

PR #1254, opened 2024-12-14
add aliases

jmilljr24:
# 🔗 Issue

# ✍️ Description

This add a few Kamal aliases.

Each of these connects to the currently running container on the production server.
`kamal shell` - opens bash in the containner
`kamal console` - opens the rails console
`kamal logs` - display's the last 10 lines of the productions log and follows as its updated

# 📷 Screenshots/Demos


kasugaijin (approved):
Nice.

PR #1255, opened 2024-12-14
Implement a reusable empty state partial for all places that need them

### LINKED ISSUE #1224: Implement a reusable empty state partial for all places that need them
We don't do a great job at showing empty states in the app. We show them sometimes, and other times not. The task in this issue is to implement a single empty state partial. Given most of our empty states are used with filters, we can have default text in the partial like 'No results matching your search criteria', and to handle other cases, e.g., 'You have no liked pets', we can pass the string we want to display, and if provided, show that instead. 

Please place this partial somewhere top level because it will be used throughout the app, e.g., `app/views/shared` and then update all places we should use an empty state. 

Off the top of my head, this includes

- Staff > pets, adoption applications, fosters, default pet tasks (where we use filters)
- The adoptable pets path
- Adopter Fosterer > liked pets, adopted pets, fostered pets
- click around and see if you can find another place...but I think that's it?

We have some places that display an empty state in a table (like adopter fosterer > adoption applications, or Staff > Pet photos/attachments). It might make sense to leave these as is and not use the new partial. You can be final judge of that.

jp524:
# 🔗 Issue
Resolves #1224. 

# ✍️ Description
Modified all relevant views to use the new `empty_state` partial. I liked the look of the existing card in `/adoptable_pets` so I decided to reuse that in the partial.

# 📷 Screenshots/Demos
Here's a sample of views using the new partial:

**staff > pet > photos**
staff   pet   photos

**staff > foster pets**
staff   foster pets

**adopter fosterer > adopted pets**
adopter fosterer   adopted pets

**adoptable pets**
adoptable pets


kasugaijin:
@jp524 cool I will take a look tomorrow!

app/views/organizations/staff/adoption_application_reviews/_search_results.html.erb
kasugaijin, line 1: New line :)

app/views/shared/_empty_state.html.erb
kasugaijin, line 1: New line :)

config/locales/en.yml
kasugaijin, line 1: I think the key could be called `empty_state` on each scope. It would just make it easier to use rather than have to look up specific keys. Thoughts? I'd say this is non-blocking, but if you agree, would be a nice change for simplicity.

I believe it is fine to re-use the same key for each nested scope scope.

kasugaijin (commented):
Looks great @jp524 
Just one comment, and a merge conflict (we have been merging a lot recently)

config/locales/en.yml
jp524: Works for me! I'll make the change.

kasugaijin (approved):
Awesome @jp524 !
Thank you

PR #1263, opened 2024-12-16
CSV Import - correctly report validation error for timestamp

jmilljr24:
# 🔗 Issue

# ✍️ Description

The validation error is correctly returned now. I also fixed the row index. Headers are skipped so the count was off by one.

Screenshot from 2024-12-16 06-15-31

# 📷 Screenshots/Demos


kasugaijin (approved):
Should we remove the validation for `csv_timestamp` on `FormSubmission` with this now? Alternatively, we could use a dummy value, but given it's `datetime` I am not sure how useful this might be.

PR #1264, opened 2024-12-16
Add solid cable and solid queue

jmilljr24:
# 🔗 Issue

# ✍️ Description

Solid Cable is a database-backed Action Cable (websockets) adapter. Removes the need for something like Redis.

Solid Queue is a DB-based queuing backend for [Active Job](https://edgeguides.rubyonrails.org/active_job_basics.html). This will allow us to add background jobs.

A few minor fixes were also added.

References:
https://github.com/rails/solid_cable
https://github.com/rails/solid_queue
# 📷 Screenshots/Demos


config/database.yml
```diff
@@ -91,10 +104,18 @@ test:
 # for a full overview on how database connection configuration can be specified.
 #
 production:
-  <<: *default
-  host: <%= ENV["DB_HOST"] %>
-  pool: 5
```
jmilljr24, line 96: Default is 5. No need to define here.

config/environments/production.rb
```diff
@@ -55,9 +55,6 @@
   config.assume_ssl = true
   config.force_ssl = true
 
-  # Include generic and useful information about system operation, but avoid logging too much
-  # information to avoid inadvertent exposure of personally identifiable information (PII).
-  config.log_level = :info
```
jmilljr24, line 60: Missed this in rails 8 upgrade. defined twice.

kasugaijin (approved):
Nice. It's nice not to have to use Redis when you don't have to. 
@jmilljr24 can you please link to the resources you followed to implement this in the description? I am assuming it's the rails guides...but maybe others.

Issue #1220, opened 2024-12-08
Staff Create Pet: add tooltips to the published and paused applications toggles


kasugaijin:
These two toggles might not be intuitive to understand what they are for, so it would be good to add an icon and tooltip description  for each. See an example of a tooktip in action on our theme here (for password strength): https://geeksui.codescandy.com/geeks/pages/security.html 

Application Paused: 'If applications are paused, the public can see the pet but cannot submit an adoption application.'
Published: 'If a pet is published it is visible to the public.'

<img width="380" alt="image" src="https://github.com/user-attachments/assets/a82acdc2-ba37-4717-bae9-e9e83284182c">

mgrigoriev8109:
Can I work on this one @kasugaijin ?

kasugaijin:
For sure all yours @mgrigoriev8109

mgrigoriev8109:
Unassigning myself from this if anyone else wants to work on it - 

assuming no one else has implemented actual responsive tooltips yet, I think the temporary fix we want to have is to include those two messages in parenthesis next to the Switch buttons.

edwinthinks:
@kasugaijin I think actually you can just change the labels to avoid the tooltip.

"Application Paused" -> "Accept new adoption applications"
"Published" -> "Pet visible to public?"

Sometimes I think we can add a bit more details in the toggle to make it more clear rather than go for the tooltip where you'd have to dive into the tooltip to understand.

kasugaijin:
@edwinthinks yes we went with modified labels a s a workaround in the meantime. The tooltip is definitely a nice to have, and you're right, to be used cautiously.

edwinthinks:
@kasugaijin yeh tooltips is good but I don't often think people use them to learn more about anything. Would be a good additional touch if you find people don't know what those mean.

PR #1267, opened 2024-12-17
feat: add new pet tooltip #1220

mgrigoriev8109:
# 🔗 Issue
[[](https://github.com/rubyforgood/homeward-tails/issues/1220)](https://github.com/rubyforgood/homeward-tails/issues/1220)

# ✍️ Description
Adds two new tooltips for Staff users that are creating a new pet:
1. Application Paused: "If applications are paused the public can see the pet, but cannot submit an adoption application."*note that I moved the comma around because it sounded a bit better, but I can keep the original wording if it's preferred*
2. Published: 'If a pet is published it is visible to the public.'
Screenshot 2024-12-16 at 9 07 28 PM

Also fixes a broken tooltip that was present under Adoption Applications. It was supposed to render properly when the New Adoption button was disabled, but wasn't.
Screenshot 2024-12-16 at 9 07 48 PM

A note - I'm not familiar with this repo and its asset pipeline, and wasn't sure the best way to apply the two javascript scripts to these views. I saw in another Ruby For Good repo that uses tooltips they just threw the script directly into the view, and it worked that way. I tried adding a `app/javascript/controllers/pets_controller.js` with 
```
import { Controller } from "@hotwired/stimulus"

export default class extends Controller {
  connect() {
  const tooltipTriggerList = document.querySelectorAll('[data-bs-toggle="tooltip"]')
  const tooltipList = [...tooltipTriggerList].map(tooltipTriggerEl => new bootstrap.Tooltip(tooltipTriggerEl))
  }
}
```

buuuuut it didn't work at all. If there's a better place for me to add in the javascript please let me know.


kasugaijin:
@mgrigoriev8109 can't we use the data attributes and classes that are used by our theme? See the example link I provided in the issue.

kasugaijin:
@mgrigoriev8109 per the meeting tonight, let's just add the text in parentheses next to each radio. We should make sure the radios are always on separate lines at all widths so it looks OK.

mgrigoriev8109:
I'm gonna close out this PR, and unassign myself from the issue for the time being.

PR #1269, opened 2024-12-17
Adding tooltip to awaiting data status for new adopters

jasonwang7517:
# 🔗 Issue
[1228: Add tooltip to the awaiting data status for adopters](https://github.com/rubyforgood/homeward-tails/issues/1228)

# ✍️ Description
Adds a new tooltip when adopters have submitted applications but have not yet had a questionnaire reviewed by staff.

# 📷 Screenshots/Demos
image


app/views/organizations/adoptable_pets/show.html.erb
```diff
@@ -154,6 +156,21 @@
                     <%= image_tag('red_no.png', height: '21') %>
                   <% else %>
                     <%= image_tag('green_check.png', height: '21') %>
+                    <% if 'awaiting_review' == @adoption_application.status %>
+                        <span class="d-inline-block" tabindex="0" 
+                              data-bs-toggle="tooltip" title="Please make sure you fill out the adoption questionnaire so staff have enough information to process your application.">
+                          <i class="bi bi-info-circle fs-4 text-secondary ps-1"></i>
+                        </span>
+
+                      <script>
```
jasonwang7517, line 165: Very new to Bootstrap so if this `script` tag and the `stylesheet` link on line 1 are unnecessary, please let me know. Still learning!

app/controllers/organizations/adopter_fosterer/adopter_applications_controller.rb
```diff
@@ -50,7 +50,8 @@ def application_params
       :pet_id,
       :form_submission_id,
       :status,
-      :profile_show
+      :profile_show,
+      :person_id
```
kasugaijin, line 54: Why is this attribute added?

app/views/organizations/adoptable_pets/show.html.erb
```diff
@@ -1,3 +1,5 @@
+<%= stylesheet_link_tag "https://cdn.jsdelivr.net/npm/bootstrap-icons@1.10.5/font/bootstrap-icons.css", rel: "stylesheet" %>
```
kasugaijin, line 1: We should not need any CDN, or custom scripts.
Our app is using a theme - see the issue where I link to it. Check that example I provided for a tooltip - you should be able to use the same data attributes and classes to make this tooltip work. 

Further, this tooltip should only show when we display the 'awaiting data' status...it doesn't make sense to show it for any other status, because the tooltip is specific to that status.

app/controllers/organizations/adopter_fosterer/adopter_applications_controller.rb
jasonwang7517: Without, it throws 
```
 Unpermitted parameter: :person_id. Context: { controller: Organizations::AdopterFosterer::AdopterApplicationsController, action: create, request: #<ActionDispatch::Request:0x0000ffff601571c8>, params: {"authenticity_token"=>"[FILTERED]", "adopter_application"=>{"pet_id"=>"8", "person_id"=>"15"}, "commit"=>"Apply to Adopt", "controller"=>"organizations/adopter_fosterer/adopter_applications", "action"=>"create"} }
```
for me.

app/views/organizations/adoptable_pets/show.html.erb
jasonwang7517: I just read through the Bootstrap 5 documentation and it seems we'll still have to add a script to initialize the tooltip since they're opt-in for performance reasons, right? [Bootstrap Documentation](https://getbootstrap.com/docs/5.0/components/tooltips/)

I read that some themes will auto-initialize but I'm not sure if this theme does that since I can't find any a `main.js` or `theme.js` or any other file that contains similar code.

app/controllers/organizations/adopter_fosterer/adopter_applications_controller.rb
kasugaijin: Ah yes this might actually fix the production error I was seeing this morning 👯

app/views/organizations/adoptable_pets/show.html.erb
kasugaijin: Try the theme implementation. The JS is bundled with it and other JS featues work from it.

app/controllers/organizations/adopter_fosterer/adopter_applications_controller.rb
jasonwang7517: heck yeah

app/views/organizations/adoptable_pets/show.html.erb
jasonwang7517: Yeah I tried

```
<button type="button" class="btn btn-secondary" data-bs-toggle="tooltip" data-placement="top" title="Tooltip on top">
  Tooltip on top
</button>
```

but it doesn't seem to pop up without the script.

https://github.com/user-attachments/assets/c17cf9ac-9dc3-4223-8562-d7d260580b04

kasugaijin:
@jasonwang7517 thanks for your back and forth on this. We decided on our call this evening to just keep it super simple for MVP. Instead of adding the tooltip with custom JS, could you please just add the text in a paragraph underneath the status? We will just show the plain text at all times instead. We could add this in a card to make it look nicer (see the empty state parial for an example). Does this make sense?

jasonwang7517:
> @jasonwang7517 thanks for your back and forth on this. We decided on our call this evening to just keep it super simple for MVP. Instead of adding the tooltip with custom JS, could you please just add the text in a paragraph underneath the status? We will just show the plain text at all times instead. We could add this in a card to make it look nicer (see the empty state parial for an example). Does this make sense?

Got it!

jasonwang7517:
> @jasonwang7517 thanks for your back and forth on this. We decided on our call this evening to just keep it super simple for MVP. Instead of adding the tooltip with custom JS, could you please just add the text in a paragraph underneath the status? We will just show the plain text at all times instead. We could add this in a card to make it look nicer (see the empty state parial for an example). Does this make sense?

I wasn't sure if by "at all times" you meant no matter what the status was so I kept it under an `awaiting_review` check but I can take that out if we truly want it for any application status!

kasugaijin (approved):

PR #1273, opened 2024-12-17
feat: add alert when missing attachment #1220

mgrigoriev8109:
# 🔗 Issue
https://github.com/rubyforgood/homeward-tails/issues/1221

# ✍️ Description
Adds an alert when a staff user attempts to upload 0 files or images

# 📷 Screenshots/Demos
`bin/dev`, log in as Staff, head to http://localhost:3000/alta/staff/pets/2?active_tab=photos 

Click Attach without any photo uploaded, and see the new alert.
Screenshot 2024-12-17 at 3 04 09 PM

Add an image and check that the success alert pops up as it was intended.
Screenshot 2024-12-17 at 3 04 35 PM

Do the same for Files
Screenshot 2024-12-17 at 3 05 06 PM
Screenshot 2024-12-17 at 3 05 24 PM


mgrigoriev8109:
Need to tinker with this a bit more, should've ran the tests locally - looks like having my method as a before_action for those actions is breaking some stuff.

app/controllers/concerns/attachment_manageable.rb
```diff
@@ -5,4 +5,15 @@ def allow_only_one_attachment
     flash.now[:alert] = t(".attachment_manageable.upload_limit_exceeded")
     render turbo_stream: turbo_stream.replace("flash", partial: "layouts/shared/flash_messages")
   end
+  
+  def alert_for_no_images
+    if params[:action] === 'attach_images' and params[:pet][:images].length > 1
```
kasugaijin, line 10: `==` in Ruby if you want equality
https://dev.to/baweaver/understanding-ruby-triple-equals-2p9c `===` in Ruby means something else

kasugaijin (commented):
Looks good! Just a couple small comments.

app/controllers/concerns/attachment_manageable.rb
kasugaijin, line 1: For the tests that are failing, you might want to throw a debugger in the tests to see what the params look like in the test context. It looks like what you are doing here to check if there is an attachment works in the browser, but not in the tests.

mgrigoriev8109:
@kasugaijin this is ready for a re-review. The `:images` and `:files` params behaved differently, just like you predicted. 

In a live environment for some reason they're populated with an empty string, so they become an array of `["", {image-object}, {image-object}]`. 

I figured instead of adding empty strings to the test environment it'd be cleaner to just remove those empty strings in the method. Let me know if this works, or if I should implement it differently.

app/controllers/concerns/attachment_manageable.rb
```diff
@@ -2,7 +2,24 @@ module AttachmentManageable
   def allow_only_one_attachment
     return unless params[:files].is_a?(Array)
 
-    flash.now[:alert] = t(".attachment_manageable.upload_limit_exceeded")
+    flash.now[:alert] = t("attachment_manageable.upload_limit_exceeded")
     render turbo_stream: turbo_stream.replace("flash", partial: "layouts/shared/flash_messages")
   end
+
+  def show_alert_if_attachment_missing
+    if !params[:pet][:images].nil?
```
kasugaijin, line 10: I think it would be more idiomatic to use
`if params[:pet][:images].present?` rather than a double negative

app/controllers/concerns/attachment_manageable.rb
```diff
@@ -2,7 +2,24 @@ module AttachmentManageable
...
+      no_empty_images = params[:pet][:images].reject { |image| image == "" }
```
kasugaijin, line 11: `no_empty_images` could just be renamed `images`, same for `files`.

app/controllers/concerns/attachment_manageable.rb
```diff
@@ -2,7 +2,24 @@ module AttachmentManageable
...
+    elsif !params[:pet][:files].nil?
+      no_empty_files = params[:pet][:files].reject { |image| image == "" }
```
kasugaijin, line 13: I'd also add a comment on why we are rejecting the ""

app/controllers/concerns/attachment_manageable.rb
mgrigoriev8109: Did this in [refactor: use present? instead of nil?](https://github.com/rubyforgood/homeward-tails/pull/1273/commits/2454e2d0883c9115aa2f2132d03d65f9dd1ba5a8)

app/controllers/concerns/attachment_manageable.rb
mgrigoriev8109: Did this in [refactor: change no_empty_images to images and files](https://github.com/rubyforgood/homeward-tails/pull/1273/commits/22d84b32d7593e16f3339229da03991124a44a7f)

app/controllers/concerns/attachment_manageable.rb
mgrigoriev8109: Did this in [chore: add comment on why we remove empty string from files / images …](https://github.com/rubyforgood/homeward-tails/pull/1273/commits/3e3bdb83b7e572b9a2baef81bff2f67f11da91fb)

kasugaijin (approved):
@mgrigoriev8109 thank you!

PR #1277, opened 2024-12-18
Docker db

kasugaijin:
# 🔗 Issue
Configures Cable and Queue to run in Docker. 
People running locally now need to add the host for both Dbs in application.yml

**When merged, we should add a note in the slack channel as a heads up for people.** Most people probably won't check the readme.

# ✍️ Description

# 📷 Screenshots/Demos


kasugaijin:
Nope all good! @jmilljr24

PR #1281, opened 2024-12-19
Csv background job

jmilljr24:
# 🔗 Issue

# ✍️ Description

This adds a background job to process the csv import. Some changes had to be made to the import service but where minor enough where the service could be run independent of the background job if needed.

This uses [Active Job](https://guides.rubyonrails.org/active_job_basics.html) for the background job and uses `Turbo::StreamsChannel` to broadcast the update when the job is completed.

Tests updated.

# 📷 Screenshots/Demos

[Screencast from 2024-12-19 07-19-35.webm](https://github.com/user-attachments/assets/291fed5e-fc2f-4674-80b4-63f5c1e4eb81)


app/controllers/organizations/staff/external_form_upload_controller.rb
```diff
@@ -11,8 +11,15 @@ def index
 
       def create
         authorize! :external_form_upload, context: {organization: Current.organization}
-        import = Organizations::Importers::CsvImportService.new(params[:files]).call
-        render turbo_stream: turbo_stream.replace("results", partial: "organizations/staff/external_form_upload/upload_results", locals: {import: import})
+        file = params[:files]
+        @blob = ActiveStorage::Blob.create_and_upload!(io: file, filename: file.original_filename)
+
+        CsvImportJob.set(wait: 1.second).perform_later(@blob.signed_id, Current.organization.id)
```
jmilljr24, line 17: I don't love this wait but there was an issue due to timing. If an exception was raised (in the validation_file method) the `blob.signed_id` would not match the id created in the view. The view needs the broadcast channel to receive the message that the import is complete. 

I need to see if using something other than the blob.signed_id will be sufficient for the pub/sub channel.

app/services/organizations/importers/csv_import_service.rb
```diff
@@ -4,61 +4,66 @@ module Organizations
...
-
-            if person.nil?
-              @no_match << [index, email]
-            elsif previously_matched_form_submission.present?
-              next
-            else
-              ActiveRecord::Base.transaction do
-                create_form_answers(FormSubmission.create!(person:, csv_timestamp:), row)
-                @count += 1
+          @file.download do |data|
```
jmilljr24, line 17: Background jobs should use a file that has been uploaded to a storage service. This steams the download and yields the chunks.

app/services/organizations/importers/csv_import_service.rb
```diff
@@ -4,61 +4,66 @@ module Organizations
...
+            validate_file(data)
+            CSV.parse(data, headers: true, skip_blanks: true).each_with_index do |row, index|
```
jmilljr24, line 19: `.foreach` only works with a file path. This parses the actual data. `with_index` does not work so I've added the offset where index is used elsewhere.

app/services/organizations/importers/csv_import_service.rb
```diff
@@ -4,61 +4,66 @@ module Organizations
...
-            @errors << [index, e]
           end
         end
-        Status.new(@errors.empty?, @count, @no_match, @errors)
+        Turbo::StreamsChannel.broadcast_replace_to @file.signed_id,
+          target: "results",
+          partial: "organizations/staff/external_form_upload/upload_results",
+          locals: {
+            import: Status.new(@errors.empty?, @count, @no_match, @errors)
+          }
```
jmilljr24, line 48: Broadcast the import status when the import is complete.

app/views/organizations/staff/external_form_upload/_upload_results.html.erb
```diff
@@ -1,21 +1,21 @@
-<turbo-frame id="results">
```
jmilljr24, line 1: This was not actually doing anything with a turbo frame so a div is more appropriate.

app/views/organizations/staff/shared/_attachment_form.html.erb
```diff
@@ -16,19 +16,18 @@
...
+              Images must be .jpg or .png under 1MB.
+            <% end %>
           </small>
         </div>
         <div class="ps-2">
           <%= form.submit t("general.attach"), class: "btn btn-outline-success" %>
         </div>
       </div>
     <% end %>
-    <%= turbo_frame_tag "results" %>
```
jmilljr24, line 32: Moved to the external form index. This partial is used in multiple locations and the results are really only need for the csv import.

app/services/organizations/importers/csv_import_service.rb
```diff
@@ -4,61 +4,67 @@ module Organizations
   module Importers
     class CsvImportService
       Status = Data.define(:success?, :count, :no_match, :errors)
-      def initialize(file)
+      def initialize(file, current_user_id)
         @file = file
-        @organization = Current.organization
+        @organization = ActsAsTenant.current_tenant
```
kasugaijin, line 9: Will this tennant scope always be reliable with this process running in another thread? I am wondering if we should set the org based off the User.

kasugaijin (commented):
Nice! @jmilljr24 Very cool. Just one comment re tenant. It would be good to get his into main and test it soon. I will be adding baja to prod later on today (they won't be uploading form data any time soon as there's not much point without adopters signing up so no rush)

app/services/organizations/importers/csv_import_service.rb
jmilljr24: I was concerned of that as well. From the docs:
```
ActsAsTenant supports

    ActiveJob - ActsAsTenant will automatically save the current tenant in ActiveJob arguments and set it when the job runs.

```

I did a "trust but verify" the best I could and it appears it works as it should. I kind of prefer sticking with what is build in with ActsAsTenant than relying on the user. If I was running this service without using a job, I'd want to get the org through ActsAsTenant. 

I added a debugger to the gem to get a better idea how it was working with its active job extension. It basically takes the current tenant and merges it into the job data hash when the job is created and queued. When the job is actually run, it gets the tenant id and sets the current_tenant. The csv service then runs within that scope.

I'll have to eat my words if it breaks in production :grin:

Screenshot from 2024-12-19 14-24-12

kasugaijin: Oh snap, yes you're right - they even have a note in their docs on this. That's really nice. 
https://github.com/ErwinM/acts_as_tenant?tab=readme-ov-file#background-processing-libraries

jmilljr24:
I'm open to suggestions on testing.

I decided to remove some things from the controller test as it seemed beyond the scope of the controller and overly complicated things. The csv service has good test coverage so I don't think we need to duplicate that in the controller test. I didn't add any tests for the job simply because there really is no additional logic in the job other than calling the service.

Issue #1219, opened 2024-12-08
Contact/Feedback forms: if I am authenticated my email address is already loaded into the form


kasugaijin:
If a user is authenticated (adopter/foster or staff) and they go to either the contact form, or the feedback form, they have to fill in the email field. It would be better UX to auto-populate this field IF the user is already authenticated i.e., `current_user` returns an object. Please update these two forms to do that.

Tests not required.

fivefootbot:
I'd like to take on this issue!

kasugaijin:
@fivefootbot all yours

PR #1282, opened 2024-12-19
Feedback and contact forms will auto-fill a user's email if they are …

fivefootbot:
…logged in

#1219  🔗 Issue

# ✍️ Description
Modified feedback and contact controllers to auto-fill a user's email if the user is logged in. If a user is not logged in, the contact and feedback forms are still accessible but the email will not auto-fill.

# 📷 Screenshots/Demos
Screenshot 2024-12-19 at 1 24 39 PM


kasugaijin (approved):
Nice! @fivefootbot

PR #1283, opened 2024-12-19
Some more small fixes

kasugaijin:
# 🔗 Issue
Fixes bugs found by @mgrigoriev8109 [here](https://rubyforgood.slack.com/archives/C05JB50LEN5/p1734574101268989)

# ✍️ Description

# 📷 Screenshots/Demos

### DIFF
--- a/app/views/organizations/adoptable_pets/index.html.erb
+++ b/app/views/organizations/adoptable_pets/index.html.erb
@@ -26,7 +26,7 @@
               </div>
               <div class="form-group">
                 <%= f.label :sex_eq, "Sex" %>
-                <%= f.select :sex_eq, ["Male", "Female"], {include_blank: 'All'}, class: "form-select" %>
+                <%= f.select :sex_eq, [["Male", "male"], ["Female", "female"]], {include_blank: 'All'}, class: "form-select" %>
               </div>
               <div class="form-group">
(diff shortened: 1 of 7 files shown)


app/views/organizations/adoptable_pets/index.html.erb
```diff
@@ -26,7 +26,7 @@
               </div>
               <div class="form-group">
                 <%= f.label :sex_eq, "Sex" %>
-                <%= f.select :sex_eq, ["Male", "Female"], {include_blank: 'All'}, class: "form-select" %>
+                <%= f.select :sex_eq, [["Male", "male"], ["Female", "female"]], {include_blank: 'All'}, class: "form-select" %>
```
jmilljr24, line 29: This is fine for now. The create pet form just gives a drop down so we really should only have downcase. I'm guessing the issue stemmed from seed data. We could a maybe add a before_save to the model.

jmilljr24 (approved):
@kasugaijin Looks good. Just one thought.

app/views/organizations/adoptable_pets/index.html.erb
kasugaijin: The pet create form will save a downcase sex, but the ransack filter was using capitalized sex. I think a better approach would be to just use an enum. That should get rid of the case issue for good.

PR #1285, opened 2024-12-19
update the custom form flow

kasugaijin:
# 🔗 Issue
It didn't feel right the way we had the browse pets buttons under the iframe form after signing up. So I have updated to go to include a button to go to dashboard instead, and added more detail to the text. 

New sign up
image

When form has not been submitted or imported
image

When form has been submitted and imported
image

# ✍️ Description

# 📷 Screenshots/Demos


jmilljr24 (approved):
@kasugaijin lgtm

PR #1300, opened 2024-12-25
Improve bin/setup by indicating application.yml needs adjusting

edwinthinks:
# 🔗 Issue
N/A

# ✍️ Description
I tried loading up homeward-tails after some time and noticed we aren't handling the `application.yml` or indicating to the newcomer that it needs adjusting.

# 📷 Screenshots/Demos
CleanShot 2024-12-25 at 11 33 56@2x

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -126,6 +126,7 @@ group :development, :test, :staging do
 end
 
 group :development, :test do
+  gem "colorize"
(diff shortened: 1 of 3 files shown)


edwinthinks:
ahh its already in the README. I don't think we need this then

Issue #1288, opened 2024-12-20
Fosterer Invite: fix input layout


kasugaijin:
For the person details (first name, last name, email, phone), put the input in two columns on 2 rows so the form looks more even (these should stack on smaller screens)

<img width="1092" alt="image" src="https://github.com/user-attachments/assets/50c5e2c2-4657-4b1a-a550-4c99e1f113d9" />

edwinthinks:
I can take this one :)

kasugaijin:
@edwinthinks Christmas commits!

edwinthinks:
Jeje, where can I find this page btw?

kasugaijin:
Log in as staff and go to dashboard > fosterers > invite

edwinthinks:
Got it, and I opened a PR. I took some design liberties here and there.

Taking a look at the app; I'd love to do a design revision on a few spots.

kasugaijin:
@edwinthinks yes please do! Do you want to make a PR for those?

PR #1301, opened 2024-12-25
Improve design of the Foster Invite form using vertical form inputs

edwinthinks:
# 🔗 Issue
#1288 

# ✍️ Description
Clean up some stylings while making some design decisions along the way. One thing I decided to try out was to put all input fields in the same column rather than trying to split them. I've been told that it is best practice for design to not do two column forms. The reasoning to me makes sense, you don't want to change a person's context when reading from left to right and downwards. This also solves the problem with handling mobile mode as it already is stacked on top of each other.

I recommend sticking to this pattern since there will less work figuring out what the next styling should be. My rule of thumb is stick to the vertical forms until it breaks and you need to introduce something else more unique.

# 📷 Screenshots/Demos

Desktop:
CleanShot 2024-12-25 at 12 58 22@2x

Mobile:

CleanShot 2024-12-25 at 12 58 31@2x

### DIFF
--- a/app/views/organizations/staff/fosterer_invitations/_form.html.erb
+++ b/app/views/organizations/staff/fosterer_invitations/_form.html.erb
@@ -1,54 +1,54 @@
-<div class="card mb-4">
-  <div class="card-body">
-    <!-- form -->
-    <%= bootstrap_form_with model: user, url: invitation_path(user) do |form| %>
-      <div class="row gx-3">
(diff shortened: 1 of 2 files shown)


kasugaijin (approved):
@edwinthinks ah yes single column is great. I agree we should stick to this. It makes sense based on your rationale.

PR #1302, opened 2024-12-25
Increase size of the dashboard for adopters to be fluid to take advantage of viewport space

edwinthinks:
# 🔗 Issue
N/A

# ✍️ Description
Thought we should increase the container of the adopter dashboard to be fluid. I think we are losing space for content via the container restricting the size of the dashboard. See snapshots below:

# 📷 Screenshots/Demos

CleanShot 2024-12-25 at 13 06 56@2x

CleanShot 2024-12-25 at 13 06 43@2x


kasugaijin (approved):
Good call @edwinthinks

Issue #1180, opened 2024-11-23
App root page - add new images


kasugaijin:
These are some simple UI changes to update images on the app root page.

There are three images to update. Each image should show a screenshot of the feature, like the others already do. Please run the final images you create here through [tinypng](https://tinypng.com/) so the file is compressed. The images you create should be the same dimensions as the other images, and on the same background div that is rounded and gray. You will note the existing images already have rounded corners. You might be able to achieve this with CSS instead, I'll let you decide what makes most sense.

Please also update the Manage People section to use this  'Create accounts and manage staff 'image from the Figma:
https://www.figma.com/design/0jVgYASUJy0KiX3BVc3dFM/Tasks?node-id=1150-254&t=ykWedzLOvulPZhF1-1

Invite and manage people
Log in as staff and go to `/staff/staff` and add a pull page of staff users (invite a load of fake accounts) and take a screenshot of this
<img width="924" alt="image" src="https://github.com/user-attachments/assets/fbb19d0f-ff0c-4adf-8408-2d7672acd5dd">

Adopter fosterer dashboard image
For this image, you will need to log in as an adopter and go to `/adopter_fosterer/applications` and take a screenshot of this page.
<img width="941" alt="image" src="https://github.com/user-attachments/assets/d6e40d60-a3bb-416e-84a8-d2c36d2e8e1c">

This foster image is the correct image, but for some reason, it is blurred. See if you can make a clearer one.
<img width="1124" alt="image" src="https://github.com/user-attachments/assets/de033169-5aff-4610-a781-0fb0c07d3823">

Aaryanpal:
@kasugaijin will take this up!
Thanks

Aaryanpal:
@kasugaijin is this issue completed?

kasugaijin:
@Aaryanpal nope the current images are placeholders! Look at them carefully and you’ll see

Aaryanpal:
you marked this as completed i mean?

kasugaijin:
Oh I am an idiot. Haha yes that was a mistake. Running on little sleep right now! Sorry.

jasonwang7517:
@kasugaijin I think this issue can be marked as resolved right?

PR #1305, opened 2024-12-28
Issue #1180: Add new images to app root page

jasonwang7517:
# 🔗 Issue
[Issue](https://github.com/rubyforgood/homeward-tails/issues/1180)

# ✍️ Description
- Adds new screenshots to the root page
- Replaces gray background with CSS shadow 

# 📷 Screenshots/Demos
https://github.com/user-attachments/assets/d95c929c-8629-48ad-8f3f-ba7b1356a8d0


app/assets/images/invite_fosterers.png
jasonwang7517, line 1: This image is a little fatter since when it's set to the same ratio as the other pictures, we get the card view which is full of "coming soon" pictures. Not sure if that matters a ton but I thought it was worth pointing out.

kasugaijin: I think this is fine for now! Better than the placeholder image we had.

kasugaijin:
@jasonwang7517 this is looking good. 
Can we get rid of this outer background entirely so we are just left with the image (rounded with small shadow)?

image

jasonwang7517:
> @jasonwang7517 this is looking good. Can we get rid of this outer background div entirely so we are just left with the image (rounded with small shadow)? If we need the div, can we just remove the border shadow from it so it is imperceptable?
> 
> image

Sure! I made the shadows small instead of large and it may be a little hard to see so let me know if you want me to change the shadows back to large. Here is how they look small:
image

kasugaijin:
@jasonwang7517 I think small looks fine!

kasugaijin (approved):

Issue #1081, opened 2024-10-23
Org root page: add browse species buttons conditionally


kasugaijin:
Currently, the browse cats and browse dogs buttons are hard coded. However, not every organization is going to have both cats and dogs. So, we should build the necessary buttons based on the unique pet species represented in the organization. 

For example, if you queried the unique species on all pet records for a given org, and it returned only 'dog', then we should only show a dog button...and so on. 

I was thinking about what we should do when there are no adoptable pets for a given species. For now, let's still show the button, but we should show an empty state. [This related issue](https://github.com/rubyforgood/homeward-tails/issues/1062) is adding empty states.

Regarding implementation, we will be handling this in the `Organizations::HomeController`. I think we should add a scope on `Pet` called something like `organization_species` that returns a list of unique species for the org. Then we can use this list to build the buttons.

![Image](https://github.com/user-attachments/assets/0b1b12c8-7ce0-4129-baf4-c0875485ec54)

Yuji3000:
I'd be happy to take this issue

kasugaijin:
@Yuji3000 all yours

jmilljr24:
Just to point out/clarify, the Pet model is scoped to the current organization using acts_as_tenant. @kasugaijin I think using `organization_species` could cause confusion. Since the model is already scoped to the org, it doesn't really tell me what the method is doing.  Maybe `unique_species` or `species_present` , thoughts?

kasugaijin:
@jmilljr24 I think `unique_species` is proably the clearest name

kasugaijin:
@Yuji3000 let us know if you have any questions!

Yuji3000:
@kasugaijin thank you

kasugaijin:
@Yuji3000 are you still able to work on this? Let me know if not and I can reassign.

Yuji3000:
I'll have a PR in this week. Sorry, extremely busy.

kasugaijin:
Someone please finish this PR :)

jasonwang7517:
@kasugaijin I can take this after my current issue.

PR #1306, opened 2024-12-28
Add unique species scope and test

Yuji3000:
# 🔗 Issue
#1081 [Issue Link](https://github.com/rubyforgood/homeward-tails/issues/1081)

# ✍️ Description
Added scope "unique_species"  of an ORG for future conditional rendering of species buttons. Simple test also added to make sure a single species isn't returned more than once. 

# 📷 Screenshots/Demos
Screenshot 2024-12-27 at 7 39 03 PM

Screenshot 2024-12-27 at 7 38 47 PM

### DIFF
--- a/app/models/pet.rb
+++ b/app/models/pet.rb
@@ -95,6 +95,9 @@ class Pet < ApplicationRecord
     query.where(published: true).order(:breed).distinct.pluck(:breed)
   }
 
+  # Returns an array of unique species of an organization
+  scope :unique_species, -> { distinct.pluck(:species) }
+  
   attr_writer :toggle
 
   # check if pet has any applications with adoption pending status
--- a/test/models/pet_test.rb
+++ b/test/models/pet_test.rb
@@ -83,6 +83,19 @@ class PetTest < ActiveSupport::TestCase
         assert_not Pet.with_photo.include?(pet_without_image)
       end
     end
+
+    context ".unique_species" do
+      should "returns an array of unique species" do
+        create(:pet, species: "Dog")
+        create(:pet, species: "Cat")
+        create(:pet, species: "Dog")
+    
+        unique_species = Pet.unique_species
+    
+        assert_equal ["Cat", "Dog"], unique_species.sort
+        assert_equal 1, unique_species.count("Dog")
+      end
+    end
   end
 
   context "#has_adoption_pending?" do


kasugaijin:
@Yuji3000 I see it's still draft so forgive me if I am stating something you intend to work on. This scope looks good. Now we should use this in places where we show the 'See <species>' buttons e.g., the org home page hero section. We should render a button (that sends the correct species param to the controller) for each unique species. 

Thanks!

Yuji3000:
@kasugaijin Thanks for clarifying.

kasugaijin (commented):
@Yuji3000 this is looking good!
We also show the species in the `Pets` dropdown in the top nav, so it would be good to update this
image

And as an adopter on your dashboard we have these buttons. 
image

I think it would be good to add the conditional button logic in the view into a partial that can be used in both cases we show buttons. Thoughts?

app/models/pet.rb
```diff
@@ -95,6 +95,9 @@ class Pet < ApplicationRecord
     query.where(published: true).order(:breed).distinct.pluck(:breed)
   }
 
+  # Returns an array of unique species of an organization
+  scope :unique_species, -> { distinct.pluck(:species) }
```
kasugaijin, line 99: Do you think this should also limit the species where the pet is adoptable? For example, if there are cats and dogs available for adoption, then all cats are adopted, do we still want to shot the view cats button? Or should we just show buttons for pets where there are currently adoptable animals?

app/views/organizations/home/index.html.erb
```diff
@@ -11,13 +11,11 @@
...
           <div class="d-flex justify-content-center">
             <% if user_signed_in? %>
-
-              <%= link_to "#{t("general.browse")} #{t("general.dogs")}",
-              adoptable_pets_path(species: "dog"),
-              class: "btn btn-primary btn-lg rounded-1 ms-2" %>
-              <%= link_to "#{t("general.browse")} #{t("general.cats")}",
-              adoptable_pets_path(species: "cat"),
-              class: "btn btn-primary btn-lg rounded-1 ms-2" %>
+              <% @unique_species.each do |species| %>
```
kasugaijin, line 14: This would be good to add to a partial that can be re-used across the app. I think it would make sense for it to live in `app/views/organizations/shared`

test/models/pet_test.rb
```diff
@@ -83,6 +83,19 @@ class PetTest < ActiveSupport::TestCase
         assert_not Pet.with_photo.include?(pet_without_image)
       end
     end
+
```
kasugaijin, line 86: If we decide not to show the species for pets without adoptable animals, we can also test that here.

app/models/pet.rb
Yuji3000: yeah, I think to only show adoptable animals.

Issue #1286, opened 2024-12-19
Adopter Authentication: Add Oauth with google


kasugaijin:
We should enable adopters to register and subsequently authenticate with Oauth via Google.

sarvaiyanidhi:
Hi @kasugaijin Can you please assign it to me?

kasugaijin:
All yours! @sarvaiyanidhi 
As part of the setup for this I think we might need set set up google API with a client ID and key. Please let me know what we need and I can set it up and give these to you for testing, and I can also add to the production credentials.
Also, please place the omniauth logic for User in a concern, something like Omniauthable

sarvaiyanidhi:
Thanks @kasugaijin Will share details with you. I will start working on the feature in 1-2 days and get back to you.

sarvaiyanidhi:
Hi @kasugaijin ,

I implemented logic by creating same credentials in my personal account but we need to setup google client_id and client_secret in credentials file to get flow working.

Also, we need to mention Authorized redirect URIs in credentials page.
Right now, I had to add below two redirect URIs to get flow working for both organizations

http://localhost:3000/alta/users/auth/google_oauth2/callback
http://localhost:3000/baja/users/auth/google_oauth2/callback

I need to try and work around to see if it is possible to get working without organization in the URI as not sure if it is feasible to add URI for each new organization

Also, I don't see any development.yml.enc in code base which I had to generate to place development credentials so is it ok to commit this file?

I still need to fix few issue in code like it is not allowing to create same google login for baja if already created for alta .

FYI I have followed this reference https://avohq.io/blog/social-login-rails-google-github

Thanks

kasugaijin:
@sarvaiyanidhi awesome thanks for the update. So a to do for me is to create a google client id and secret for homeward tails? I will check the resource you followed tomorrow for the details on that.

as for credentials, please don’t upload any dev credentials. I recommend creating rails credentials for development and referring to that. See how we refer to the rails credentials in other places in the app. Rails will point to the correct credentials file (dev or prod) based on the environment. I can handle the prod credentials.

Good point on the redirect url. Can’t we create this url dynamically using the org slug? The paths should be the same, just the slugs will be dynamic I believe?

sarvaiyanidhi:
Thanks @kasugaijin Above mentioned URLs are added in Google credentials form as shown below

[...]

PR #1309, opened 2024-12-30
1286 - Adopter Authentication: Add Oauth with google 

sarvaiyanidhi:
# 🔗 Issue
#1286 

# ✍️ Description
- Implemented Google login using omniauth-google-oauth2 gem
- By default adopter role is set with user on signup
- Need to setup and add google client_id and client_secret in credentials file to get flow working

Format for adding credentials

```
google:
  client_id: CLIENTID_VALUE
  client_secret: CLIENTSECRET_VALUE
```

# 📷 Screenshots/Demos

google_signup

google_login


sarvaiyanidhi:
@kasugaijin Do you want me to update settings page by removing password fields for user logged in using Google?

pet-rescue-settings

kasugaijin:
> @kasugaijin Do you want me to update settings page by removing password fields for user logged in using Google?
> 
> pet-rescue-settings

Yes that makes sense. Do you know if the password field is populated if an adopter signs up via oath? I think there’s something because I believe it’s a model validation?

kasugaijin:
@sarvaiyanidhi in `app/services/organizations/create_service.rb` could you please add a comment `# Be sure to add the callback URL for the new org in for Google OAuth dev account`

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,18 @@
+class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
```
kasugaijin, line 1: Can we nest this please :)
```
module Users
  OmniauthCallbacksController < ...
```

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+    if @user.persisted?
+      sign_in_and_redirect @user, event: :authentication
+      set_flash_message(:notice, :success, kind: "Google") if is_navigational_format?
+    else
+      session["devise.google_data"] = request.env["omniauth.auth"].except(:extra)
+      redirect_to new_user_registration_url, alert: @user.errors.full_messages.join("\n")
+    end
+  end
+
+  def failure
```
kasugaijin, line 15: Where does this action get used? There's no route for it and I don't see it referenced in the action above it.

app/models/concerns/omniauthable.rb
```diff
@@ -0,0 +1,27 @@
+module Omniauthable
+  extend ActiveSupport::Concern
+
+  included do
+    devise :omniauthable, omniauth_providers: [:google_oauth2]
+  end
+
+  class_methods do
+    def from_omniauth(auth)
+      where(provider: auth.provider, uid: auth.uid, organization_id: Current.organization.id).first_or_create do |user|
```
kasugaijin, line 10: Hmm. If I am staff authenticating with OAut, is this not going to set an adopter role on my account when I authenticate with OAuth?

We should only set the adopter role on the user IF the user is new. We should not touch roles of users who already exist, no?

kasugaijin: Ah nevermind...my misunderstanding of how `first_or_create` works. The block is skipped if the user exists. Nice.

kasugaijin (commented):
@sarvaiyanidhi looking good. 
I am part way through setting up the developer account and getting the credentials. I am stuck on domain verificiation and waiting on someone else to help me with that (I don't have access to the domain). Then I will be able to push a commit to your branch with the updated production credentials for this :)

sarvaiyanidhi:
> > @kasugaijin Do you want me to update settings page by removing password fields for user logged in using Google?
> Yes that makes sense. Do you know if the password field is populated if an adopter signs up via oath? I think there’s something because I believe it’s a model validation?

Yes, when we create user we do set password for the user using Devise method. Here is the the code which generates password in app/models/concerns/omniauthable.rb

```
 def assign_attributes_from_auth(auth)
    self.email = auth.info.email
    self.password = Devise.friendly_token[0, 20]
    self.first_name = auth.info.first_name if respond_to?(:first_name)
    self.last_name = auth.info.last_name if respond_to?(:last_name)
  end
```

app/controllers/users/omniauth_callbacks_controller.rb
sarvaiyanidhi: Sure.

kasugaijin:
> > > @kasugaijin Do you want me to update settings page by removing password fields for user logged in using Google?
> > > Yes that makes sense. Do you know if the password field is populated if an adopter signs up via oath? I think there’s something because I believe it’s a model validation?
> 
> Yes, when we create user we do set password for the user using Devise method. Here is the the code which generates password in app/models/concerns/omniauthable.rb
> 
> ```
>  def assign_attributes_from_auth(auth)
>     self.email = auth.info.email
>     self.password = Devise.friendly_token[0, 20]
>     self.first_name = auth.info.first_name if respond_to?(:first_name)
>     self.last_name = auth.info.last_name if respond_to?(:last_name)
>   end
> ```

Yes that’s right ok cool. Then yeah it doesn’t make sense to show the password field if the user is authenticated with oauth.

sarvaiyanidhi:
> Yes that’s right ok cool. Then yeah it doesn’t make sense to show the password field if the user is authenticated with oauth.

Done. Have pushed the changes

edit_settings

app/controllers/users/omniauth_callbacks_controller.rb
sarvaiyanidhi: Done. Pushed the changes

kasugaijin:
@sarvaiyanidhi I have pushed a commit to add the client ID and secret after setting up the dev account > oauth, and will give it 24 hours so the callback links register. Hopefully I have set things up properly :D

sarvaiyanidhi:
> @sarvaiyanidhi I have pushed a commit to add the client ID and secret after setting up the dev account > oauth, and will give it 24 hours so the callback links register. Hopefully I have set things up properly :D

Sure. Let me know if you need anything from my end. Thanks

kasugaijin (approved):

Issue #1304, opened 2024-12-27
Add tooltip feature


kasugaijin:
There’s a few places in the app where tooltip functionality would be handy. 

This thread here has an implementation using Stimulus (preferred) so take a look and see if you can get it working. You might not need all of the JavaScript, so just get the basics working to show the tooltip on hover and remove it after hover 

https://rubyforgood.slack.com/archives/C05JB50LEN5/p1734639082814269

A good place to implement this first off is on the new pet form. Log in as staff and go to the create pet form and you will see two radios and the bottom with descriptions. These descriptions should be moved to the tooltip. And we should use an icon next to the field name that indicates the user can hover for information.

jasonwang7517:
I can take this!

jasonwang7517:
@kasugaijin Can you clarify what descriptions should be moved? I can't seem to find any information to put in the tooltip on the new pet form.
![image](https://github.com/user-attachments/assets/baed27f0-f6c8-430a-8f07-4ef5bf5bfc65)

kasugaijin:
@jasonwang7517 application paused - pet is visible to public but they cannot make applications
Published - pet is visible to public

jasonwang7517:
Sorry for dragging my feet on this, have been busy with interview prep but will find time later this week to bang it out!

PR #1313, opened 2025-01-03
Issue #1304 - Add tooltip feature 

jasonwang7517:
# 🔗 Issue
[Original issue](https://github.com/rubyforgood/homeward-tails/issues/1304)

# ✍️ Description
- Creates a basic Stimulus controller that can be used for future tooltip instances
- Adds a tooltip to the toggle sections of the create pet form

# 📷 Screenshots/Demos
https://github.com/user-attachments/assets/9a566d70-7612-4fe9-9e29-25ae359c8573


app/views/organizations/staff/pets/_form.html.erb
```diff
@@ -80,14 +80,30 @@
...
       <div class='form-group'>
-        <div class="form-check form-switch ps-3">
-            <%= form.hidden_field :application_paused, value: false %>
-            <%= form.check_box :application_paused, { class: "form-check-input", role: "switch", id: "flexSwitchCheckChecked" }, true, false %>
+        <div class="form-check form-switch d-flex align-items-center ps-3">
+          <%= form.hidden_field :application_paused, value: false %>
+          <%= form.check_box :application_paused, { class: "form-check-input", role: "switch", id: "flexSwitchCheckChecked" }, true, false %>
+          <div data-controller="tooltip" data-action="mouseenter->tooltip#showTooltip mouseleave->tooltip#hideTooltip" class="mb-3 ms-1" tabindex="0">
+            <i class="fe fe-help-circle"></i>
+            <div data-tooltip-target="tooltip" class="tooltip bs-tooltip-top fade" role="tooltip" style="display: none;  position:absolute;">
```
kasugaijin, line 88: Nice!
I think we could make this even simpler and easier to use throughout the app if we turn the markup that shows as a the tooltip into a partial we can then use elsewhere in the app. I am thinking we could create a `_tooltip.html.erb` partial in `app/view/organizations/shared` that looks something like 

```
<div data-controller="tooltip" data-action="mouseenter->tooltip#showTooltip mouseleave->tooltip#hideTooltip" class="mb-3 ms-1" tabindex="0">
            <i class="fe fe-help-circle"></i>
            <div data-tooltip-target="tooltip" class="tooltip bs-tooltip-top fade" role="tooltip" style="display: none;  position:absolute;">
              <div class="tooltip-inner">
                :text
              </div>
            </div>
          </div>
```

We can pass the text we want to show as an argument to the partial to make it versatile. 

e.g. 

` <%= render partial: "organizations/shared/tooltip", locals: { text: "this is what the user sees" }`

jasonwang7517: Always coming in with the hot ideas! I gotchu

jasonwang7517:
unironically committed this code hours ago and as I was about to go to sleep I remembered and woke up to push it.

just thought it was worth nothing.

kasugaijin (approved):
@jasonwang7517 looks great!

PR #1316, opened 2025-01-06
Allow duplicate email between orgs

kasugaijin:
# 🔗 Issue

# ✍️ Description
Per [this discussion](https://rubyforgood.slack.com/archives/C05JXC199M2/p1736129153581599) I have put up this PR as a proposal to allow users to register across organizations with the same email address. This will create separate user records scoped to each org, which I think is fine, considering the organizations do not have any cross functionality or data sharing. This is the simplest approach to allow a user to register on multiple pet rescue orgs, and is likey a need given we might have multiple orgs using our app on the Baja peninsula alone. 

We will still ensure uniqueness on email under the Org scope with this validation that already exists on User
```
validates :email, presence: true, uniqueness: {scope: :organization_id}, format: {
    with: /\A([^@\s]+)@((?:[-a-z0-9]+\.)+[a-z]{2,})\z/i
  }
```

# 📷 Screenshots/Demos

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -29,7 +29,7 @@
 #
 # Indexes
 #
-#  index_users_on_email                 (email) UNIQUE
+#  index_users_on_email                 (email)
 #  index_users_on_invitation_token      (invitation_token) UNIQUE
 #  index_users_on_invited_by            (invited_by_type,invited_by_id)
 #  index_users_on_invited_by_id         (invited_by_id)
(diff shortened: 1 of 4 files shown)


jmilljr24:
Do we want any other validations that [Validatable](https://github.com/heartcombo/devise/blob/main/lib/devise/models/validatable.rb) includes?

kasugaijin:
> Do we want any other validations that [Validatable](https://github.com/heartcombo/devise/blob/main/lib/devise/models/validatable.rb) includes?

Yeah, I think we should validate password and confirmation for presence and length, for sure.

Issue #1319, opened 2025-01-14
CSV Import file input - update the file type description


kasugaijin:
Log in as staff and go to the external form upload route and look at the file input description, this should say `Files must be .CSV...`. Please fix. This input uses a partial, so please update that to handle CSV type.

<img width="1020" alt="image" src="https://github.com/user-attachments/assets/853af348-cc07-488d-b3f7-cbf3bb656b76" />

jasonwang7517:
@kasugaijin I can take this. I know I have that platform issue outstanding but that seems kind of blocked I'm low-key mad confused about how to do it.

kasugaijin:
@jasonwang7517 all yours!

jmilljr24:
@jasonwang7517 Looks like this issue has a PR that was merged, however I have a bug locally after that commit.

When I upload a valid csv file, I get the error "File must be CSV." Are you able to reproduce this error?

https://github.com/user-attachments/assets/75321255-c8d8-47fa-84db-95443b35aab7

jasonwang7517:
I can take a look in a bit!

jmilljr24:
@jasonwang7517 Looks like it was just a minor fix in #1333 . No bid deal. Thanks!!!

PR #1325, opened 2025-01-19
Issue #1319 - Update file type description and add CSV file type validation

jasonwang7517:
# 🔗 Issue
[Original Issue](https://github.com/rubyforgood/homeward-tails/issues/1319)

# ✍️ Description
- Updates text under upload input to inform user that only CSV files will be accepted
- Adds a validation within the `create` action of the controller to display an alert if user uploads a file type other than CSV.

# 📷 Screenshots/Demos
_New copy and error message on top_
image

### DIFF
--- a/app/controllers/organizations/staff/external_form_upload_controller.rb
+++ b/app/controllers/organizations/staff/external_form_upload_controller.rb
@@ -12,10 +12,11 @@ module Organizations
       def create
         authorize! :external_form_upload, context: {organization: Current.organization}
         file = params[:files]
+
         unless file.content_type == "text/csv"
           redirect_to staff_external_form_upload_index_path, alert: "File must be a CSV."
         end
-        
+
         @blob = ActiveStorage::Blob.create_and_upload!(io: file, filename: file.original_filename)
 
         CsvImportJob.perform_later(@blob.signed_id, current_user.id)
--- a/app/controllers/organizations/staff/external_form_upload_controller.rb
+++ b/app/controllers/organizations/staff/external_form_upload_controller.rb
@@ -3,7 +3,8 @@ module Organizations
     class ExternalFormUploadController < Organizations::BaseController
       include AttachmentManageable
       layout "dashboard"
-      before_action :validate_attachment, only: [:create]
+      before_action :allow_only_one_attachment, only: [:create]
+      before_action :handle_incorrect_file_format_when_csv_expected, only: [:create]
 
       def index
         authorize! :external_form_upload, context: {organization: Current.organization}
@@ -21,11 +22,6 @@ module Organizations
         respond_to do |format|
           format.turbo_stream
         end
-      end
-
-      def validate_attachment
-        allow_only_one_attachment
-        handle_incorrect_file_format_when_csv_expected
       end
     end
   end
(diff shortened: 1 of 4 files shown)


app/views/organizations/staff/shared/_attachment_form.html.erb
```diff
@@ -18,7 +18,7 @@
                           hide_label: true %>
           <small class="form-text text-muted">
             <% if attachment_type == 'files' %>
-              Files must be .pdf, .png, or .jpeg under 2MB.
+              File must be .csv under 2MB.
```
kasugaijin, line 21: I think this change will impact all other places we use this partial and will now always say `File must be .csv under 2MB.` even in places where we expect a pdf. This is going to be wrong. 

What we should do is add a new conditional option, like `csv` and add this message there. So when I call the partial I pass the `csv` attachment type. 

Does this make sense?

jasonwang7517: Oh yeah my bad! I gotchu tho

app/controllers/organizations/staff/external_form_upload_controller.rb
```diff
@@ -12,6 +12,11 @@ def index
       def create
         authorize! :external_form_upload, context: {organization: Current.organization}
         file = params[:files]
+
```
kasugaijin, line 15: This can be added to the `module AttachmentManageable`. Given it's something that we would want to use across multiple controllers we should have it in a module, then call it in a `before_action` on the `#create` method. 

Something like 
```
def handle_wrong_file_format_for_csv
  unless params[:files].content_type == "text/csv
    redirect_to staff_external_form_upload_index_path, alert: "File must be a CSV."
   end
end
```

And call it on the `class ExternalFormUploadController` **after** the `before_action :allow_only_one_attachment, only: [:create]` on the `#create` action.

app/controllers/organizations/staff/external_form_upload_controller.rb
```diff
@@ -21,6 +22,11 @@ def create
           format.turbo_stream
         end
       end
+
+      def validate_attachment
```
kasugaijin, line 26: Can we leave this as two separate `before_action` calls :)
It makes it easier to read them both at the top, instead of scrolling down to read both.

kasugaijin (commented):
I also just had a thought. 

First we check `show_alert_if_attachment_missing` which renders a flash via turbostream if no file is given. 

I THINK that if you click attach without any file attached we might get an error with your new method because `params[:files].content_type` will try to call `content_type` on `nil`. So, I think we should instead redirect in the `show_alert_if_attachment_missing` method instead of sending a turbbostream flash. We can keep the same flash message. 

Can you verify I am not going insane and test this? Does this make sense?

jasonwang7517:
> I also just had a thought.
> 
> First we check `allow_only_one_attachment` which renders a flash via turbostream if no file is given.
> 
> I THINK that if you click attach without any file attached we might get an error with your new method because `params[:files].content_type` will try to call `content_type` on `nil`. So, I think we should instead redirect in the `allow_only_one_attachment` method instead of sending a turbbostream flash. We can keep the same flash message. This should then mean we don't even call your new method because it will redirect in cases where there's not attachment.
> 
> Can you verify I am not going insane and test this? Does this make sense?

Yeah you're right! That being said I added the safe navigation operator `&.` which fixes the issue.

kasugaijin (approved):
Nice one

Issue #1227, opened 2024-12-08
Edit User Registration redirect path - update for staff and adopter/fosterer


kasugaijin:
If you log in as a fosterer or adopter and you edit your user registration (e.g., add an avatar) when you save it will redirect you to the org home page. But, we should change this. If the user is an adopter or fosterer, it should redirect to the adopter fosterer dashboard `adopter_fosterer/dashboard` and if it staff it should redirec to the `/staff/dashboard`

We may as well add a controller test for these

Gabe-Torres:
I'll take this one!

jmilljr24:
@Gabe-Torres Thanks!

kasugaijin:
@Gabe-Torres just so I am not going crazy, you had a PR up for this then closed it, right? Are you planning to add a new one? No rush...just getting my assumptions right.

Gabe-Torres:
@kasugaijin Not going crazy! I had a PR up but closed it by mistake. I just pushed a new one 👍🏽

PR #1329, opened 2025-01-20
Update: User Registration Redirect Path

Gabe-Torres:
# 🔗 Issue
#1227 

# ✍️ Description
    This update utilizes Devise's after_update_path_for method to handle
    user redirection after updating their registration.

    If the user is part of the adopter or fosterer roster, and associated
    with a valid organization they will be redirected to their adopter/fosterer dashboard.

    If the user is an admin or super admin, and associated with a valid
    organization they will be redirected to their staff dashboard.

    If neither condition is met, the user will be redirected to the root path.

    Adds: Controller tests for user registration updates:
    Test case for adopter/fosterer user redirection.
    Test case for staff user redirection.

    Adds: update on registration test file to account for new update redirect

# 📷 Screenshots/Demos

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -64,6 +64,19 @@ class RegistrationsController < Devise::RegistrationsController
       :avatar)
   end
 
+  def after_update_path_for(resource)
+    role = resource.roles.find_by(resource_id: Current.organization.id)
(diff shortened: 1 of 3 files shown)


kasugaijin (approved):
Nice! @Gabe-Torres

Issue #1296, opened 2024-12-21
Avatarable Concern: Update validations


jmilljr24:
The avatarable concern currently uses `between` for the size validation. Chane this to have max file size of `less_than: 1.megabyte`. Also add i18n to the messages(content_type and size), preferably using lazy lookup.

We are using the [Active Storage Validations Gem](https://github.com/igorkasyanchuk/active_storage_validations)

Meri-MG:
hi, @jmilljr24 , can i try working on this one?

jmilljr24:
@Meri-MG Have at it!!

PR #1331, opened 2025-01-20
1296 - Avatarable Concern: Update validations

Meri-MG:
# 🔗 Issue
#1296 

# ✍️ Description
- Updated the avatarable concern to use `less_than` for the size validation instead of `between`, utilizing the default i18n message provided by Active Storage Validations.
- Applied lazy lookup for the customized content type validation message.
- Removed tests related to the minimum size of the avatar, as the minimum size is no longer defined.


app/models/concerns/avatarable.rb
```diff
@@ -7,9 +7,8 @@ module Avatarable
     has_one_attached :avatar unless ENV["PRECOMPILING"]
 
     # TODO:  move these validation strings to a locale file
```
kasugaijin, line 9: Can we remove this comment now?

kasugaijin (commented):
@Meri-MG looks good! Looks like we can remove that comment now.

Meri-MG:
> @Meri-MG looks good! Looks like we can remove that comment now.

done! I have to read comments more carefully 🫠

kasugaijin (approved):

Issue #1326, opened 2025-01-19
Staff dash index: refactor mobile view


kasugaijin:
Log in as staff and go to the dashboard where we show a quick summary of things in cards. Then go to mobile view and click a link in the tasks stats table. You will note that the table where we render rows of pets with overdue or incomplete tasks is way down at the bottom of the page and the user has no idea that clicking the tasks card should update that table at the bottom of the page.

**To do**
Delete the Rescues card as it’s doing nothing

Move the Tasks card to the bottom so the table render directly beneath the task card

Remove the sex column in the table because it’s not needed and the table looks funky on mobile with it (be mindful if this is a partial used elsewhere as then we can’t simply just delete the column so may want to just bring the html into the view file instead of using a partial)

![Image](https://github.com/user-attachments/assets/2f52aeff-658f-4f7b-87cb-c2070a35e090)

jasonwang7517:
I can take this!

PR #1332, opened 2025-01-20
Issue #1326: Refact mobile view of staff dashboard

jasonwang7517:
# 🔗 Issue
[Original issue](https://github.com/rubyforgood/homeward-tails/issues/1326)

# ✍️ Description
- Deletes unused Rescue card
- Moves Tasks card to the bottom to be closer to the Overdue Pet Tasks card
- Removes unneeded sex column from table

# 📷 Screenshots/Demos
image

### DIFF
--- a/app/views/organizations/staff/dashboard/index.html.erb
+++ b/app/views/organizations/staff/dashboard/index.html.erb
@@ -87,43 +87,7 @@
     </div>
 
     <div class="mt-5">
-      <turbo-frame id="tasks-frame">
-      <h3 class = "text-capitalize"><%= @header_title %></h3>
-        <div class="card">
-          <table class="table mb-0 text-nowrap table-hover table-centered">
-            <thead>
-              <tr>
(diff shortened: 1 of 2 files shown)


app/views/organizations/staff/dashboard/index.html.erb
```diff
@@ -103,7 +87,43 @@
     </div>
 
     <div class="mt-5">
-      <%= render 'pets_with_incomplete_or_overdue_tasks' %>
```
jasonwang7517, line 106: @kasugaijin Pulled out this raw HTML like you said to do in the ticket but wanted to ask if there was ever a situation in which we'd want the Sex column to appear in this table. If not, we should consider editing the partial directly since it's not really used anywhere else or we can add a param to determine rendering it.

kasugaijin: If that partial isn’t used anywhere else let’s just remove that column completely!

PR #1333, opened 2025-01-21
fix the param name and callbacks

kasugaijin:
# 🔗 Issue
Fixes a bug caused by `params[:files]` after updating the upload form to send `params[:csv]`. I prefer specifiying the file format because it helps separate from the other two attachment types we have: `files` and `images`.

image

# ✍️ Description

# 📷 Screenshots/Demos

### DIFF
--- a/app/controllers/concerns/attachment_manageable.rb
+++ b/app/controllers/concerns/attachment_manageable.rb
@@ -1,6 +1,6 @@
 module AttachmentManageable
-  def allow_only_one_attachment
-    return unless params[:files].is_a?(Array)
+  def allow_only_one_csv_attachment
+    return unless params[:csv].is_a?(Array)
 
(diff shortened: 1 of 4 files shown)


app/controllers/concerns/attachment_manageable.rb
```diff
@@ -1,6 +1,6 @@
 module AttachmentManageable
-  def allow_only_one_attachment
-    return unless params[:files].is_a?(Array)
+  def allow_only_one_csv_attachment
```
kasugaijin, line 2: This method could be made more expandable by declaring a parameter to define the file type when used, but given we don't need that anywhere, this is fine for now, I think.

jmilljr24 (approved):
@kasugaijin Awesome! I just noticed this bug but didn't see your PR. Thanks for the fix.

Issue #1238, opened 2024-12-11
Invited User: Force new user to choose to accept terms and conditions 


kasugaijin:
SuperAdmin have the ability to invite staff and fosterers to the application. These invite flows create a new User. We have a boolean on user `tos_agreement` that we use to track if the user accepts, or not, the terms and conditions and privacy policy. This value is `nil` when a user is invited. Adopter users have to select `true` in order to sign up with the application. 

We want to make sure that invited users (staff, fosterers) have to check true for these when they first log into the application. 

There are two possible ways I can see this working:

Add a filter on `ApplicationController` - we check the user's `tos_agreement` and if it is `nil`, or `false` we boot them to a page that asks them to agree to the T&C etc., and only when they do this, can they access other routes. 

You will also need to create a new controller to show this new page, and accept the submission of the response (to update the `tos_agreement` attribute.

mononoken:
@kasugaijin I'll work on this one since I've got it fresh on my mind from the meeting 🙂

kasugaijin:
@mononoken thanks Ken!

kasugaijin:
One small consideration we need to make is in the org create service we actually mark the tos_agreement on that user we create as true. So we might want to remove that so they actually have to go through this flow as well.

mononoken:
Good thinking

mononoken:
## Acceptance Criteria
- [ ] Add an acceptance criteria page and route
- [ ] Require all users to accept TOS before interacting with private pages and routes
  - [ ] Adopters
  - [ ] Fosterers
  - [ ] Staff
  - [ ] Staff admin
- [ ] Make sure the original org user also has to accept TOS

kasugaijin:
> ## Acceptance Criteria
> * [ ]  Add an acceptance criteria page and route
> * [ ]  Require all users to accept TOS before interacting with private pages and routes
>   
>   * [ ]  Adopters
>   * [ ]  Fosterers
>   * [ ]  Staff
>   * [ ]  Staff admin
> * [ ]  If invited, user should be prompted to accept TOS when setting their password
> * [ ]  Make sure the original org user also has to accept TOS

[...]

PR #1341, opened 2025-01-22
1238 require tos

mononoken:
# 🔗 Issue
#1238 

# ✍️ Description
Users who have had accounts created through some method that did not involve accepting the TOS agreements will be forwarded to a form to accept the agreements.

My local environment isn't working after merging with main. If it's working for others, please confirm flows and pages still work.

If you have any other changes you'd prefer made, let me know. Thanks, y'all!

# 📷 Screenshots/Demos

SCR-20250121-ryox


app/services/organizations/create_service.rb
```diff
@@ -73,8 +73,7 @@ def create_user(email, first_name, last_name)
         email: email,
         first_name: first_name,
         last_name: last_name,
-        password: SecureRandom.hex(3)[0, 6],
-        tos_agreement: 1
+        password: SecureRandom.hex(3)[0, 6]
```
kasugaijin, line 76: Is it possible to create this user now we have the validation on User? Or does this create_user fail? Did you test creating an Organization?

app/controllers/agreements_controller.rb
```diff
@@ -0,0 +1,26 @@
+class AgreementsController < ApplicationController
```
kasugaijin, line 1: What do you think of the name `TosAgreements`? I find it a bit more explicit in what it's for.

app/controllers/organizations/staff/invitations_controller.rb
kasugaijin, line 1: Same on this controller, when we sent the invite we create the user record - does this work now we have the validation for `:tos_agreement` required to be true?

MooseCowBear: What if instead of requiring the tos agreement as a condition on a user model, we added it to the authorization policy? So basically if the user hasn't agreed to anything they can't do anything (other than go to the page where they can accept the agreement)?

kasugaijin: Yeh that’s not a bad idea in theory as I can see how it’s tied to what you’re allowed to do. We’d have to have a policy check at a high enough level I.e where we have the current check, so that it’s always hit when a user access an org scope route. It’s basically saying you’re not allowed to do anything unless you have accepted the TOS.,right?

MooseCowBear: Right that's how I was thinking about it

mononoken: See here for the validation mechanics: https://api.rubyonrails.org/classes/ActiveModel/Validations/HelperMethods.html#method-i-validates_acceptance_of

One way to quickly test and verify it is working fine is check out the branch and run the below in `bin/rails c --sandbox`:
```ruby
u = User.new(organization: Organization.first, email: "foo@example.com", first_name: "Foo", last_name: "Bar", password: "123baz")
u.save # true
```

For the model, I would prefer to keep the validation. This Rails validation was created specifically for this purpose.

We can move the controller check to a policy instead, and that may be the more pure approach. It will be many more lines of code than the 10 lines we have in one spot here though.

Solution I'm thinking of:
1. Create new parent policy for organization scoped policies
2. Change all organization scoped policies to inherit the new parent.
4. Add precheck in new parent policy to check user tos_agreement status
5. If tos_agreement `false`, add a "failure reason" to the error signifying tos_agreement
6. Instead of the rescue logic above, we would check every authorization failure to see if it has this failure reason. If it does, we would forward them to the tos form.
7. Check every organization scoped controller action is using an organization scoped policy (not 100% sure this is the case currently)

I am happy to work on it if that is preferred. We could also merge this solution and refactor it later. But it also may be worth it doing this work now, so that there is precedence for using "failure reasons" in policies, which I think we currently do not use at all. What do y'all think? @kasugaijin @MooseCowBear

app/services/organizations/create_service.rb
kasugaijin: Ah yes, per this line `This check is performed only if terms_of_service is not nil.`

app/controllers/organizations/staff/invitations_controller.rb
kasugaijin: @mononoken I think that is a sound approach. 👍

app/controllers/agreements_controller.rb
mononoken: Updated.

app/controllers/organizations/staff/invitations_controller.rb
kasugaijin: One thought. Couldn’t this policy be global rather than org scoped? If ever a user is signed in they need to accept TOS regardless or organization.

mononoken: Sounds good! I'll get started on that solution then and let y'all know if I encounter any problems with it.

kasugaijin: To my comment above whether or not it could be global rather than org scoped, I think it depends on whether we want to apply this at the root level, which I think we should? We can probably add an exceptions list in application controller that will permit access to certain routes like user registration edit.

kasugaijin: > One thought. Couldn’t this policy be global rather than org scoped? If ever a user is signed in they need to accept TOS regardless of organization.
> 
> It also shouldn’t prevent someone from setting a password if invited for example.

Not sure if you saw this comment, @mononoken ^ ? I think it could be a reasonable, and more simple approach?

mononoken: Made these changes to use the policy. Making it scoped broadly was a good idea!

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -5,8 +5,13 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
...
       @user = User.from_omniauth(request.env["omniauth.auth"])
       if @user.persisted?
-        sign_in_and_redirect @user, event: :authentication
+        sign_in @user, event: :authentication
         set_flash_message(:notice, :success, kind: "Google") if is_navigational_format?
+        if !@user.tos_agreement
+          redirect_to edit_tos_agreement_path
+        else
+          redirect_to root_path
+        end
```
mononoken: We redirect oauth users straight to tos agreement if they don't have it yet, since they won't by default.

app/controllers/application_controller.rb
```diff
@@ -7,9 +7,15 @@ class ApplicationController < ActionController::Base
   KNOWN_ERRORS = [ActionPolicy::Unauthorized]
   rescue_from StandardError, with: :log_and_reraise
   rescue_from ActionPolicy::Unauthorized do |ex|
-    flash[:alert] = t("errors.authorization_error")
+    if ex.result.reasons.details.values.flatten.any?(:no_tos_accepted)
+      flash[:alert] = "You must accept the Terms of Service before continuing."
+      session[:original_url] = request.original_url
 
-    redirect_back_or_to root_path
```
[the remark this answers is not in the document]
kasugaijin: Nice!

app/controllers/tos_agreements_controller.rb
```diff
@@ -0,0 +1,26 @@
+class TosAgreementsController < ApplicationController
+  before_action :set_user, only: %i[edit update]
+
+  def edit
+  end
+
+  def update
+    if @user.update(user_params)
+      redirect_to session.delete(:original_url) || root_path, notice: "TOS accepted"
```
kasugaijin, line 9: Huh I never knew you could do it this way, I guess `session.delete(:xyz)` returns the value? That's nifty.

app/policies/application_policy.rb
kasugaijin, line 1: This is nice and simple :)

app/controllers/application_controller.rb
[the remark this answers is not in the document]
kasugaijin: In future iterations this would be nice to handle with a modal, too.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -5,8 +5,13 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
     def google_oauth2
       @user = User.from_omniauth(request.env["omniauth.auth"])
       if @user.persisted?
-        sign_in_and_redirect @user, event: :authentication
+        sign_in @user, event: :authentication
         set_flash_message(:notice, :success, kind: "Google") if is_navigational_format?
+        if !@user.tos_agreement?
```
kasugaijin, line 10: Did you test this? I think it will always force a user who signs up/in via Google Oauth to the tos path for the first time, right? I wonder if we could automatically set such user's tos to `true` when created because there is already a disclaimer pointing to the tos and privacy policy. 

Alternatively, we always force their sign up/first sign in to check this box off. It's a bit of friction but then we have for certain validated in our own data they have accepted.

app/policies/user_policy.rb
```diff
@@ -0,0 +1,7 @@
+class UserPolicy < ApplicationPolicy
+  skip_pre_check :verify_tos_agreement!
+
+  def manage?
```
kasugaijin, line 4: I think you have said this before, but we don't have to explicitly set the alias rule for an action, then?

e.g. ` alias_rule :edit?, to: :manage?`

app/controllers/tos_agreements_controller.rb
mononoken: Yup. Most Rails built in destroy commands return the object destroyed if successful.

app/controllers/users/omniauth_callbacks_controller.rb
mononoken: I was intending to always send them to the form if not completed, since we opted to not let users do much without accepting.

Not a fan of auto-setting TOS to true. NAL but am dubious of just showing links being **clear consent**. If it's not, it would make this field pretty useless if some users have clear consent and others do not. 

I think it's especially not consent with that verbage. Maybe you could claim some kind of implied consent with `By joining, you agree to our [Terms of Service] and [Privacy Policy]` or something like that. Maybe varying levels based on jurisdiction.

app/policies/user_policy.rb
mononoken: This is an Action Policy default for Rails.

`edit? update? delete? show?` are aliased to `manage?` by default. Other actions need to be explicitly aliased. These ones are specifically set as defaults since they're the ones that typically handle existing records. I have some notes on this at `app/policies/application_policy.rb:12`

kasugaijin: Ah yes. Thank you.

app/controllers/users/omniauth_callbacks_controller.rb
kasugaijin: OK that sounds reasonable to me. It's not something that is common with OAuth in all the times I have used it from memory (to have to accept ToS after authenticating), but, I think it's a worthy trade off, especially given this app is living under the RfG umbrella.

kasugaijin (approved):
LGTM!

PR #1345, opened 2025-01-25
Add browse species button unconditionally

jasonwang7517:
# 🔗 Issue
[Original issue](https://github.com/rubyforgood/homeward-tails/issues/1081)

# ✍️ Description
- Changes logic for creation of **Browse** buttons on the root page to be programmatic based on unique species 

# 📷 Screenshots/Demos


jasonwang7517:
@kasugaijin I know there was a conversation on the original PR about whether we should only display the button if there are adoptable pets of that given species but I know we also added [an empty state](https://github.com/rubyforgood/homeward-tails/issues/1062). 

Thus, I kept the scope as is to just pull the unique species of all pets regardless of `placement_type` value but if you would like me to specify the scope to factor that in as well such that, for example, **Browse Dogs** only appears if there is at least one record with `species: "Dog"` and `placement_type: "Adoptable"` just let me know.

kasugaijin:
> @kasugaijin I know there was a conversation on the original PR about whether we should only display the button if there are adoptable pets of that given species but I know we also added [an empty state](https://github.com/rubyforgood/homeward-tails/issues/1062).
> 
> Thus, I kept the scope as is to just pull the unique species of all pets regardless of `placement_type` value but if you would like me to specify the scope to factor that in as well such that, for example, **Browse Dogs** only appears if there is at least one record with `species: "Dog"` and `placement_type: "Adoptable"` just let me know.

The simplest approach is to keep as you suggest. I just wonder though, why take people to an empty page?

jasonwang7517:
> > @kasugaijin I know there was a conversation on the original PR about whether we should only display the button if there are adoptable pets of that given species but I know we also added [an empty state](https://github.com/rubyforgood/homeward-tails/issues/1062).
> > Thus, I kept the scope as is to just pull the unique species of all pets regardless of `placement_type` value but if you would like me to specify the scope to factor that in as well such that, for example, **Browse Dogs** only appears if there is at least one record with `species: "Dog"` and `placement_type: "Adoptable"` just let me know.
> 
> The simplest approach is to keep as you suggest. I just wonder though, why take people to an empty page?

To get their hopes of adopting their dream pet as high as we can before we shatter them mercilessly.

Alternatively, we can add some more logic in the partial to disable the button if there aren't any adoptable pets of that species. Just an initial suggestion. Or we can keep it as is, totally up to you boss.

kasugaijin:
> > > @kasugaijin I know there was a conversation on the original PR about whether we should only display the button if there are adoptable pets of that given species but I know we also added [an empty state](https://github.com/rubyforgood/homeward-tails/issues/1062).
> > > Thus, I kept the scope as is to just pull the unique species of all pets regardless of `placement_type` value but if you would like me to specify the scope to factor that in as well such that, for example, **Browse Dogs** only appears if there is at least one record with `species: "Dog"` and `placement_type: "Adoptable"` just let me know.
> > 
> > 
> > The simplest approach is to keep as you suggest. I just wonder though, why take people to an empty page?
> 
> To get their hopes of adopting their dream pet as high as we can before we shatter them mercilessly.
> 
> Alternatively, we can add some more logic in the partial to disable the button if there aren't any adoptable pets of that species. Just an initial suggestion. Or we can keep it as is, totally up to you boss.

Can’t we just chain two scopes together? I think we have one to get the adoptable pets, then one to get the species? Then the whole thing is controlled by one query without need for view logic. Does that make sense?

app/models/pet.rb
```diff
@@ -95,6 +95,9 @@ class Pet < ApplicationRecord
     query.where(published: true).order(:breed).distinct.pluck(:breed)
   }
 
+  # Returns an array of unique species of an organization
+  scope :unique_species, -> { where(placement_type: ["Adoptable", "Adoptable and Fosterable"]).distinct.pluck(:species) }
```
jasonwang7517, line 99: I think this is what we want right? Both of these placement types can be considered "adoptable"?

kasugaijin: Yup!

test/models/pet_test.rb
```diff
@@ -83,6 +83,28 @@ class PetTest < ActiveSupport::TestCase
...
+      should "returns an array of unique species" do
+        create(:pet, species: "Dog", placement_type: "Adoptable")
+        create(:pet, species: "Cat", placement_type: "Adoptable")
+        create(:pet, species: "Dog", placement_type: "Adoptable")
+
+        unique_species = Pet.unique_species
+
+        assert_equal ["Cat", "Dog"], unique_species.sort
+      end
+
```
kasugaijin, line 97: We should also add a test here that handles the case where the pet is adopted i.e., has a `match` with `match_type` of `adoption`

app/models/pet.rb
jasonwang7517: I _think_ so. Let me know if what I did in my most recent commit isn't in line with what you were looking for.

kasugaijin: Yeh that looks good!

kasugaijin:
Looking good. There’s one more place we use these buttons - log in as an adopter or fosterer and in your dashboard index you will see them. Can you use this feature there as well?

also if you want to add in this PR we also need to add the logic to the top navbar dropdown that lets your browse species.

jasonwang7517:
Gotchu

app/views/layouts/shared/_navbar.html.erb
```diff
@@ -44,15 +44,11 @@
...
-            <li>
-              <%= link_to t("general.dogs"),
-              adoptable_pets_path(species: "dog"),
-              class: "dropdown-item" %></li>
-
-            <li>
-              <%= link_to t("general.cats"),
-              adoptable_pets_path(species: "cat"),
-              class: "dropdown-item" %></li>
+            <% Pet.adoptable_unique_species.each do |species| %>
```
kasugaijin, line 47: Ah so for this one I’m 99% sure that this will not be scoped to the tenant (organization) and will actually query all pet records across all orgs because it is in the layouts and not handled by a specific org scoped controller. So you will need to scope this to the tenant. See how we handle things in the footer partial in the layouts for example, where we set the org tenant. If it’s easier it might be worth removing this here and making a new PR. Up to you.

you can manually test this by making one org have only dogs and the having dogs and cats in another org and then see if the first org with only dogs shows the cats as well in the navbar dropdown.

does this make sense?

jasonwang7517: Ah yeah, I totally forgot about the scoping issue on the controller. I'll fix it.

kasugaijin: Yeh it won’t be in a controller either because this is rendered across loads of controllers so it will have to be scoped in the view query. Might be better to move the query to a helper and call it something like ‘pet browse links for tenant’ or the likes

jasonwang7517: big brain I'll make those edits my good sir

app/helpers/adoptable_unique_species_for_tenant_helper.rb
jasonwang7517, line 1: Tested this logic by nuking all the poor cats for Alta on my local and the dropdown only showed dogs.
image

app/helpers/adoptable_unique_species_for_tenant_helper.rb
```diff
@@ -0,0 +1,9 @@
+module AdoptableUniqueSpeciesForTenantHelper
+  def adoptable_unique_species(organization)
+    organization.pets
```
kasugaijin, line 3: Why not use the new scope here?
`:adoptable_unique_species`

jasonwang7517: I think I hallucinated while testing last night. For some reason I thought I tried to run `Organization.first.pets.adoptable_unique_species` in the Rails console and got an error but that's not the case.

kasugaijin: Whoo lay off the good stuff!

app/helpers/adoptable_unique_species_for_tenant_helper.rb
```diff
@@ -0,0 +1,5 @@
+module AdoptableUniqueSpeciesForTenantHelper
```
kasugaijin, line 1: Ah I hate to be a pain in the butt, but can you add a comment here to state why we do this? I.e. this is used in the navbar and therefore it is not scoped to a tenant, so we have to set the tenant before making the query.

app/views/organizations/shared/_unique_species.html.erb
```diff
@@ -0,0 +1,3 @@
+<% @adoptable_unique_species.each do |species| %>
```
kasugaijin, line 1: And here, the rails way is to never use instance variables in partials. It should just be `adoptable_unique_species` and then you can pass that to the partial when you call the partial in 
app/views/organizations/home/index.html.erb
app/views/organizations/adopter_fosterer/dashboard/index.html.erb

kasugaijin (commented):
It's looking good! Sorry for my back and forth on it. These last two small changes should do it for approval.

kasugaijin (approved):
Nice!

Issue #1317, opened 2025-01-06
Optimize asset compilation workflow


jasonwang7517:
Currently, changes to assets (images, Stimulus controllers, etc.) do not take effect in the app unless one runs:

1. `bin/rails assets:clobber`
2. `bin/rails assets:precompile`

This is time-consuming and also requires a full restart of the server. To save time, it would be nice if we could automatically update these assets in development without manual actions.

jasonwang7517:
@kasugaijin can you assign me to this when you get a chance? I don't think I have the right GitHub permissions but I'll start work on it.

jasonwang7517:
Just a general note, might be a little while before I can work on this. Currently spending most of my time preparing for some upcoming interviews but will get back to it as soon as I can!

kasugaijin:
@jasonwang7517 no problem! Good luck

jasonwang7517:
I'm going to put this on ice for now. I can't seem to figure out what's going on as, based on my understanding, the whole purpose of Propshaft is to streamline asset delivery without having to manually precompile them.

jmilljr24:
@kasugaijin @jasonwang7517 How are things working now? Can we close this?

kasugaijin:
Yup I think so.

PR #1346, opened 2025-01-25
Fix: docker databases

### LINKED ISSUE #1334: Get solid queue working in Docker
I am running Docker and tried to import a CSV in the external form upload tool (staff dashboard > external form upload) and the job does not get performed in Docker with solid queue. This could be troubleshot and fixed!

jmilljr24:
# 🔗 Issue
Resolves #1334 

# ✍️ Description

The queue and cable database's tables were not being created when initially running db:migrate. The default install of Solid Cable and Solid queue do not add a migration file. The tables are built from the schema files. Running db:migrate seems to looks for the migration files and when there is none, it overwrites the schema files and we end up with an database with no tables.

This change adds db:prepare to the entry point file which properly creates the tables and also seeds the db. I also removed asset precompile, maybe that will help with the assets issues in development?? #1317 

I'd suggest nuking your database, then run docker compose build.

# 📷 Screenshots/Demos

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -50,6 +50,8 @@ services:
     environment:
       DATABASE_URL: postgres://${DATABASE_USERNAME:-postgres}:${DATABASE_PASSWORD:-password}@postgres:5432/petrescue_development
       RAILS_ENV: development
+      SERVICE_NAME: app
+      SOLID_QUEUE_IN_PUMA: true
     stdin_open: true
     tty: true
 
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -4,5 +4,10 @@ set -e
 # Remove a potentially pre-existing server.pid for Rails.
 rm -f /petrescue/tmp/pids/server.pid
 
+if [ "$SERVICE_NAME" = "app" ]; then
+  echo "Running db:prepare"
+  /petrescue/bin/rails db:prepare
+fi
+
 # Then exec the container's main process (what's set as CMD in the Dockerfile).
 exec "$@"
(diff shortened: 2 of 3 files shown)


docker-compose.yml
```diff
@@ -50,6 +50,8 @@ services:
     environment:
       DATABASE_URL: postgres://${DATABASE_USERNAME:-postgres}:${DATABASE_PASSWORD:-password}@postgres:5432/petrescue_development
       RAILS_ENV: development
+      SERVICE_NAME: app
+      SOLID_QUEUE_IN_PUMA: true
```
jmilljr24, line 54: When true, runs solid queue job service via puma.

entrypoint.sh
```diff
@@ -4,5 +4,10 @@ set -e
 # Remove a potentially pre-existing server.pid for Rails.
 rm -f /petrescue/tmp/pids/server.pid
 
+if [ "$SERVICE_NAME" = "app" ]; then
```
jmilljr24, line 7: We only want this to run on that app service not sass.

jasonwang7517:
> I also removed asset precompile, maybe that will help with the assets issues in development?? https://github.com/rubyforgood/homeward-tails/issues/1317

Big if true

kasugaijin:
Nice! It looks good. I just want to test locally before approving. I should get to that this week.

kasugaijin (approved):
@jmilljr24 looking good so far! It has fixed some nagging issues I had. One being it would kill the app service if I got any 500 errors, but now that doesn't happen. It's also processing background jobs properly locally which is awesome! I was able to upload via the external form upload, for example. 

I also tested adding a new image asset and it worked great. No need to recompile.

Great job!

PR #1353, opened 2025-02-02
Add option to customize ports for cable and queue

mononoken:
# 🔗 Issue

n/a

# ✍️ Description

Right now, our `docker-compose.yml` has these ports mapped:
```yml
  postgres:
    ports:
      - 54320:5432
  ...
  cable:
    ports:
      - 54321:5432
  ...
  queue:
    ports:
      - 54322:5432
```
However, our database configs do not use these local ports nor have a way to allow their use. This config change would allow a user to set env variables to pick what ports they want locally instead. 

One use case would be to set these local ports so that you can run postgresql, cable, and queue in Docker compose but run the app locally with these set:
```sh
export DATABASE_PORT=54320
export CABLE_HOST=localhost
export CABLE_PORT=54321
export QUEUE_HOST=localhost
export QUEUE_PORT=54322
```

I prefer a setup like this, since I like running the database in the Docker container, but I prefer just running the app locally.

This won't affect running the whole app in docker compose, since it keeps the same default ports.

# 📷 Screenshots/Demos

n/a


kasugaijin (commented):
Thanks @mononoken 
Should be update the Readme accordingly?

config/database.yml
```diff
@@ -43,20 +36,19 @@ development:
...
+    port: <%= ENV.fetch("CABLE_PORT") { "5432" } %>
     migrations_paths: db/cable_migrate
     gssencmode: disable
 
   queue:
     <<: *default
     database: queue
-    host: <%= ENV.fetch('QUEUE_HOST') { 'queue' } %>
+    host: <%= ENV.fetch("QUEUE_HOST") { "localhost" } %>
+    port: <%= ENV.fetch("QUEUE_PORT") { "5432" } %>
```
mononoken: I think these defaults should be optimized for local development without Docker. Docker is another dependency for a dev that they may not know or may not want to use.

That said, I think we should set the hosts to localhost as the default, and we can use the environment variables to set them in the compose for Docker.

docker-compose.yml
```diff
@@ -49,6 +49,10 @@ services:
         condition: service_healthy
     environment:
       DATABASE_URL: postgres://${DATABASE_USERNAME:-postgres}:${DATABASE_PASSWORD:-password}@postgres:5432/petrescue_development
+      DATABASE_USERNAME: ${DATABASE_USERNAME:-postgres}
+      DATABASE_PASSWORD: ${DATABASE_PASSWORD:-password}
+      CABLE_HOST: cable
+      QUEUE_HOST: queue
```
mononoken, line 55: Set defaults for Docker in here instead of in database.yml.

.github/workflows/rubyonrails.yml
```diff
@@ -27,6 +27,8 @@ jobs:
     env:
       RAILS_ENV: test
       DATABASE_URL: "postgres://rails:password@localhost:5432/rails_test"
+      DATABASE_USERNAME: postgres
+      DATABASE_PASSWORD: password
```
mononoken, line 31: Think I may need to push this first before I can merge this branch or it won't be able to pass the workflows, since these currently don't exist.

config/database.yml
```diff
@@ -18,16 +18,9 @@ default: &default
...
-  port: <%= ENV.fetch("DATABASE_PORT") do
-          begin
-            docker_port = `docker compose port postgres 5432 2> /dev/null`.split(":").last
-            Integer(docker_port.strip)
-          rescue
-            5432
-          end
-        end %>
+  username: <%= ENV.fetch("DATABASE_USERNAME") %>
+  password: <%= ENV.fetch("DATABASE_PASSWORD") %>
```
mononoken, line 22: I am suggesting we remove these defaults for a better developer error experience. These environment variables need to be setup and PostgreSQL setup for a dev to run the app locally. If we remove the defaults, it's going to yell at you for not having "DATABASE_USERNAME" set.

If we have defaults, it's going to tell them that the role "postgres" does not exist. I think that's a little more confusing for new devs.

config/database.yml
```diff
@@ -18,16 +18,9 @@ default: &default
...
-  username: <%= ENV.fetch('DATABASE_USERNAME') { 'postgres' } %>
-  password: <%= ENV.fetch('DATABASE_PASSWORD') { 'password' }  %>
-  port: <%= ENV.fetch("DATABASE_PORT") do
-          begin
-            docker_port = `docker compose port postgres 5432 2> /dev/null`.split(":").last
-            Integer(docker_port.strip)
-          rescue
-            5432
-          end
-        end %>
```
mononoken, line 30: I'm not sure why this was necessary at some point, but I think we can get rid of it. Devs should be able to use the default or set an environment variable for customization.

mononoken (commented):
@kasugaijin Thanks for taking a look! I noticed the local development without Docker required more configuration than it used to. I fiddled with some other configs so that I think it should be fairly seamless for both use cases. With these changes, if you aren't using docker, you just need to setup your environment variables for your PostgreSQL role, and you should be good to go. 

I'm not sure we need to add any comments for the initial change I proposed. We have docs for both Docker and for pure local. I added a little customization to allow someone to do a hybrid approach, but I think that should not be something we document and recommend to do but allow people to do it if they know how to and want to.

Let me know what you think, or if you have any objections or suggestions to any of these changes. Thank you!

jmilljr24 (commented):
@mononoken I'm on mobile for this week so I can't test anything out but I'm just a bit confused about some of the changes. Forgive me if any of my questions are dumb.

You mentioned the ports in your description but there's no changes to those lines?

For context, in docker-compose.yml, 54321:5432 was required because we are setting up three different containers each with their own postgres database. They each need a different port on your local machine (54321, 54322 etc.) Those ports are mapped to the containers port of 5432.

I don't see a problem with allowing a dev to set a different port but I'm not sure of the use case for postgres running in a container. Nothing else should be running in the container that would conflict with the default and most commonly used port of 5432.

~~Did you confirm changing the hosts for cable and queue didn't affect running a full docker setup (vs. your hybrid approach). For a new local setup the host variable is being set in application.yml to localhost. If I remember correctly, doing a full docker setup required having the host set different for each container to connect properly.~~
I just saw how you switched it. That makes sense to have local development as the default and then define the different hosts for docker compose!

.github/workflows/rubyonrails.yml
jmilljr24: Looking at that database url, I think the user name is rails?

mononoken: Woops, you're right! thank you I'll update that

config/database.yml
```diff
@@ -43,20 +36,19 @@ development:
...
+    port: <%= ENV.fetch("CABLE_PORT", "5432") %>
     migrations_paths: db/cable_migrate
     gssencmode: disable
 
   queue:
     <<: *default
     database: queue
-    host: <%= ENV.fetch('QUEUE_HOST') { 'queue' } %>
+    host: <%= ENV.fetch("QUEUE_HOST", "localhost") %>
+    port: <%= ENV.fetch("QUEUE_PORT", "5432") %>
```
mononoken, line 48: > You mentioned the ports in your description but there's no changes to those lines?

@jmilljr24 These are the ports I mentioned. We have 54320, 54321, 54322 mapped locally in the compose but since we do not allow the user to select the ports in the **Rails db config**, they cannot actually hook up the app locally to these ports. Adding the env variable fetch here means you could set variables to those ports to allow a local Rails app to utilize the containerized databases.

mononoken: Thinking about this more just now, we probably actually want to use these `_PORT` variables in the compose too. For example, in the unlikely case that a dev has these ports locally occupied, they should be able to set their own ports in one spot that does not require them to manually update the compose file. We can set the defaults again in the compose file. I'll look at this next break I get.

jmilljr24: Ah gotcha. For some reason I thought that was what you were originally intending to resolving. Thanks for clarifying.

config/database.yml
kasugaijin: This is for anyone running the app on WSL

config/database.yml
kasugaijin: I think this is the only thing we'd want to update the readme for? To ensure both setups (local/docker) set these values.

mononoken: Ah, gotcha! I assumed that was already in the README, since it already is necessary to set them. I'll take a look.

To clarify too, you do not have to set these with Docker. It sets these for you.

config/database.yml
mononoken: I am very skeptical this is actually necessary for WSL. Are you running on Windows? I don't have a Windows machine to test but am fairly confident this new config will work fine for WSL.

mononoken:
@kasugaijin I made some updates to the README. Let me know if you have any other suggestions. I also opened #1358. We will need to merge that first for the Standard job to run on this branch. Hoping to make tomorrow's meeting too if you would prefer to discuss any of this real time.

.github/workflows/rubyonrails.yml
```diff
@@ -27,6 +27,8 @@ jobs:
     env:
       RAILS_ENV: test
       DATABASE_URL: "postgres://rails:password@localhost:5432/rails_test"
+      DATABASE_USERNAME: rails
+      DATABASE_PASSWORD: password
```
jmilljr24, line 31: I merged the workflow changes but the "standard" job is failing. We might need to add these under "standard" in this workflow as well.

mononoken: Thanks, @jmilljr24 !

That seems suspect. Why is the standard job accessing the db at all? Looks like it is because it is using the Rails runner: https://github.com/rubyforgood/homeward-tails/blob/9c2652599da400799ed6751d136776da382762d4/bin/style#L7C1-L7C19

I think it could just run with bundler:
```
bundle exec standardrb
```

I can take a further look after work.

jmilljr24: Good question. I took a quick look at the docs for Standard but nothing stood out. If you get more insight I'd love to know!

mononoken: Opened #1360. Hopefully that will get this ready to merge too.

mononoken:
Rebased to remove the workflow changes from this one, since those changes are in separate PRs now. I also did a little more cleaning up the README. I considered moving Docker below the local install stuff, but since it's a lot less steps to just download Docker and use docker compose, I think it makes sense to keep it up top.

jmilljr24 (approved):
@mononoken  lgtm

kasugaijin (approved):
LGTM.

PR #1354, opened 2025-02-02
Initial commit: update User and Person models

BurkeMtn:
# 🔗 Issue
https://github.com/rubyforgood/homeward-tails/issues/1324

# ✍️ Description
https://github.com/rubyforgood/homeward-tails/issues/1324

# Testing

**Login as Adopter/Fosterer**
- Login to alta
- Verify user is automatically logged in to baja
- Verify user can visit all public routes on baja
- Verify Person does not exist for baja
- Visit private route
- Verify user is prompted to join the organization
- Join organization
- Verify Person is created for baja
- Verify adopter role for baja has been added for the user
- Verify user is able to visit private routes

**Sign-up as Adopter/Fosterer**
- Logout
- Sign up as adopter
- Verify Person is created in context

**Login as Staff**
- Login to alta
- Verify user is automatically logged in to baja
- Verify user can visit all public routes on baja
- Verify Person does not exist for baja
- Visit private route
- Verify user is not authorized
- Have baja staff invite user to baja organization as staff
- Verify user is able to visit private routes

# 📷 Screenshots/Demos
https://www.loom.com/share/7dafc5b11cde49a79174112648350292?sid=da32fc33-9ee1-4d26-8b77-f2a6bb116d3b
https://www.loom.com/share/7dafc5b11cde49a79174112648350292?sid=f9a2143f-bf98-4ab7-9309-34a9413d86c5


app/models/user.rb
```diff
@@ -111,15 +95,9 @@ def inactive_message
   end
 
   def ensure_person_exists
```
kasugaijin, line 97: I know this is WIP, but I couldn't help but take a peak. 

Is this callback still here just to get the authentication working? If so, forget the rest of this comment. 

In the issue I describe an alternative process where we get rid of this callback and instead create the person contextually (adopter sign up, staff invite, fosterer invite) so then we know what roles to add to the person.

app/models/user.rb
```diff
@@ -71,15 +52,18 @@ class User < ApplicationRecord
...
+  has_many :organizations, through: :people
+  accepts_nested_attributes_for :people
 
   before_validation :ensure_person_exists, on: :create
 
   before_save :downcase_email
 
   delegate :latest_form_submission, to: :person
 
+  self.ignored_columns += ["person_id"]
```
kasugaijin, line 65: do we need this? `person_id` was removed from users table

app/models/user.rb
BurkeMtn: Looks like it. It runs every time a User is created. Unless there are other flows that create User objects. This doesn't seem to be the case (since I didn't see any overrides for devise registration controllers).

app/models/user.rb
BurkeMtn: Rails recommend this as part of DB migration (related to a bunch of the safety checks that are run). Suspect this can be remove though.

app/models/user.rb
BurkeMtn: Looks like the invitation controller is also creating Users.

kasugaijin: Yeah we should create the Person in those flows I think I rather than on the model

app/policies/organizations/adopter_fosterer/match_policy.rb
```diff
@@ -2,7 +2,7 @@ module Organizations
   module AdopterFosterer
     class MatchPolicy < ApplicationPolicy
       relation_scope do |relation|
-        relation.where(person_id: user.person.id)
+        relation.where(person_id: Person.where(user_id: user.id).first)
```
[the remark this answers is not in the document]
BurkeMtn: Good catch. Fixed.

app/views/organizations/staff/fosterer_invitations/new.html.erb
```diff
@@ -2,7 +2,7 @@
   <% c.with_header_title { t("invite_fosterer") } %>
   <% c.with_body do |b| %>
     <div class='mx-auto' style='max-width: 600px'>
-      <%= render "organizations/staff/fosterer_invitations/form", user: @user %>
+      <%= render "organizations/staff/fosterer_invitations/form", user: @user, person: @person %>
```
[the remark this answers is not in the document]
BurkeMtn: It is used to generate the URL for the form: https://github.com/rubyforgood/homeward-tails/blob/f6f20558b77cd16da34786d1f614645042a09df3/app/views/organizations/staff/fosterer_invitations/_form.html.erb#L1

app/controllers/organizations/staff/invitations_controller.rb
```diff
@@ -83,9 +83,8 @@ def after_accept_path_for(_resource)
     end
   end
 
-  # TODO: We should consider removing this duplicated logic, so we don't have to do this.
   def assign_person_attributes(user)
-    person = user.person || user.build_person
+    person = user.person || user.people.build(organization: Current.organization)
```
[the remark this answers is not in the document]
BurkeMtn: > Currently, it looks like we're creating a new user for every invitation. But I believe we will want to only create a new user if one doesn't exist. If that's right, then this controller will need more revisions.

This is a good point but valid regardless of whether the user is scoped to an organization (i.e. a user with the same email may already exist).

The focus of this PR is to make user global scope keeping everything else unchanged (to the extent possible). Given this seems like a separate bug should we fix it here or in a follow up PR? Happy either way... Just don't want this PR to bloat too much.

test/models/user_test.rb
```diff
@@ -59,8 +59,7 @@ class UserTest < ActiveSupport::TestCase
       user = create(:admin)
       assert_includes User.staff, user
 
-      user.destroy
-      assert_not_includes User.staff, user
+      user.people.destroy
```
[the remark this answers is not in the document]
BurkeMtn: Good catch. Updated. This was failing due to a FK constraint on User from Person. All people need to be deleted before the User can.

app/models/user.rb
```diff
@@ -110,16 +89,20 @@ def inactive_message
     deactivated? ? :deactivated : super
   end
 
-  def ensure_person_exists
-    return if person.present?
-
-    existing = Person.find_by(organization: organization, email: email)
+  def person
+    people.where(user_id: id).first
```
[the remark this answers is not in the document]
MooseCowBear: @BurkeMtn I understand there can't be more than one person per organization, but what does `user.people` return if the current tenant is not set? Maybe this method isn't intended to ever be called when there is no current tenant, but if the behavior of this method depends on whether or not a current tenant exists, then there's a brittleness we might want to think about trying to avoid (maybe adding a guard clause?) or at least document it with some test coverage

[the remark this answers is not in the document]
jmilljr24: I can see the ambiguity where a tenant wouldn't be set (especially for new contributors not familiar with acts_as_tenant).

 I've looked through the code and it appears 99% of the places this method is used is on current_user. One suggestion would be to remove this method and instead rely on using a Current Attribute for Person. I'm not sure of any major downsides. 

Anywhere `current_user.person` is used could be replaced with `Current.person`.  We would set `Current.person` most likely in `ApplicationController` else it would be nil. This would behave very similar to current_user and be pretty obvious to anyone new to the code base. Since most things are now revolving around the Person model (within an organization) this seems to make sense in my mind.

Action Policy also relies on the current user which may need to be tweaked a bit but similar outcome. The invitations controller is the only other location I found that calls `user.person` (not on current_user) but this needs to be refactored after this PR anyways.

Just a suggestion! If it's dumb feel free to ignore :smiley:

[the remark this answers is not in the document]
BurkeMtn: > @BurkeMtn I understand there can't be more than one person per organization, but what does user.people return if the current tenant is not set?

Good question. User and Person where scoped to Organization prior to this change so there shouldn't be any cases where this happens IIUC. For _potential_ future use-cases using User.person will not make sense. The lookup will need to be for a specific User which will depend on the context.

> Maybe this method isn't intended to ever be called when there is no current tenant, but if the behavior of this method depends on whether or not a current tenant exists, then there's a brittleness we might want to think about trying to avoid (maybe adding a guard clause?) or at least document it with some test coverage

Agree but would also argue this implicit behaviour exists throughout the code given our use of ActAsTenant on numerous models. That said, agree on the guard clause.

> Anywhere current_user.person is used could be replaced with Current.person. We would set Current.person most likely in ApplicationController else it would be nil.

Interesting idea. I agree it would make the Organization scope more explicit. A few questions come up: How does this work when no organization is present? Would this create similar confusion (in the other direction - sort of)? Is this a pattern we want to apply to other object? IMO, I think it will be hard to avoid getting familiar with the ActAsTenant concept when working on this app :)...

app/controllers/organizations/staff/invitations_controller.rb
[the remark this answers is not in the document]
BurkeMtn: Tested this. The current invitation form checks if the email is already registered.

app/controllers/application_controller.rb
```diff
@@ -16,6 +17,16 @@ def set_current_user
     Current.user = current_user
   end
 
+  def verify_and_set_current_person
```
kasugaijin, line 20: I like the idea of setting a person context to use in the request, rather than having to use `current_user.person` everywhere. 

@jmilljr24 and I discussed on the call today the best place for it. Here in `ApplicationController` this callback will be hit on every request, and some of these requests are going to be on routes outside the organization context. WE could handle this conditionally like you have, but I think it might be even better to only set the person when we know there is an organization to worry about.

We have the `OrganizationScopable` module that is included in the `OrganizationBaseController`, which every controller in the `Organization` directory inherits from. In short, all routes where we need to worry about an organization will call this module. It's also the same place where we set the `current_tenant` which is used by the `ActsAsTenant` gem to, among other things, scope queries to that tenant. I think it would make more sense to include there? What do you think?

I did wonder whether we would prefer to set an instance variable `@person` instead of `Current.person`. The former makes sense if we are only refering to the person object in controllers and views. But, if we are needing it in other places, like models, then we'd have to pass the object as an argument. I think `Current.person` alleviates this and is available globally.

app/controllers/organizations/home_controller.rb
```diff
@@ -1,5 +1,6 @@
 class Organizations::HomeController < Organizations::BaseController
   skip_before_action :authenticate_user!
+  skip_before_action :verify_and_set_current_person
```
kasugaijin, line 3: If you implement [this](https://github.com/rubyforgood/homeward-tails/pull/1354/files#r2023859335) you can remove all of these `skip_before_action` calls on the non-org scoped routes.

app/models/user.rb
```diff
@@ -110,15 +88,11 @@ def inactive_message
...
+    raise StandardError, "Organization not set" unless Current.organization
 
-    existing = Person.find_by(organization: organization, email: email)
-
-    if existing
-      self.person = existing
-    else
-      build_person(first_name:, last_name:, email:, organization:)
+    ActsAsTenant.with_tenant(Current.organization) do
+      people.where(user_id: id).first
```
kasugaijin, line 95: `people.find_by(user_id: id)` will do the same thing more performantly.

app/controllers/organizations/staff/invitations_controller.rb
```diff
@@ -83,11 +97,9 @@ def after_accept_path_for(_resource)
...
   end
 
-  # TODO: We should consider removing this duplicated logic, so we don't have to do this.
-  def assign_person_attributes(user)
-    person = user.person || user.build_person
-    person.first_name = user.first_name
-    person.last_name = user.last_name
-    person.email = user.email
+  def create_person(user)
+    unless (Person.where(email: user.email).any?)
```
kasugaijin, line 101:
```suggestion
    unless (Person.exists?(email: user.email))
```

app/controllers/organizations/staff/invitations_controller.rb
```diff
@@ -17,13 +17,28 @@ def create
...
+      # TODO: The user should probably get an email letting them know they
+      # were added to a new organization.
+      if User.where(email: user_params[:email]).any?
+        @user = User.find_by(email: user_params[:email])
+        if !@user.has_role?(user_params[:roles], Current.organization)
+          @user.add_role(user_params[:roles], Current.organization)
+          redirect_to staff_staff_index_path, notice: t(".success")
+          return
+        end
+      end
```
kasugaijin, line 32: Thoughts on this flow instead?

Keep the comments as is 
```suggestion
     @user = User.find_by(email: user_params[:email]
     
     if @user.present?
        if !@user.has_role?(user_params[:roles], Current.organization)
          @user.add_role(user_params[:roles], Current.organization)
          redirect_to staff_staff_index_path, notice: t(".success")
          return
        end
     else
     .
     .
     .
   end
```

app/controllers/organizations/staff/people_controller.rb
```diff
@@ -0,0 +1,40 @@
+class Organizations::Staff::PeopleController < Organizations::BaseController
```
kasugaijin, line 1: Nit: could we reformat to use nested modules/classes eg

```
module Organizations
  module Staff
    Class PeopleController
```

app/controllers/organizations/staff/people_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  skip_before_action :verify_and_set_current_person
```
kasugaijin, line 2: Update this callback if you do implement this https://github.com/rubyforgood/homeward-tails/pull/1354/files#r2023859335

app/controllers/organizations/staff/people_controller.rb
kasugaijin, line 1: If I understand correctly, this controller will handle the cases where I am, say, an adopter who has already got a Person for Organization A, but then I go and check out Organization B, and now I need to complete an action to get my person created for Organization B, right? If so, why is this scoped under `Organizations::Staff`? I think it would better live under `Organizations`?

app/controllers/organizations/staff/people_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  before_action :validate_person_does_not_exist
+  
+  def new
+    @person = Person.new
+  end
+
+  def create
+    @person = Person.new(person_params)
+    if @person.save
+      Current.user.add_role(:adopter, Current.organization)
```
kasugaijin, line 13: Yes this makes sense to only allow someone to create a person with an adopter role, since staff and fosterers need to be invited.

app/controllers/organizations/staff/people_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    else
+      flash.now[:alert] = "There was an unexpected error. Please try again soon."
+      render :new
+    end
+  end
+
+  private
+  
+  def validate_person_does_not_exist
+    if Person.where(user_id: Current.user.id).exists?
```
kasugaijin, line 24:
```suggestion
    if Person.find_by(user_id: current_user.id).present?
```

app/controllers/organizations/staff/people_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+      flash[:notice] = "You are already a member of this organization."
+      redirect_to root_path
```
kasugaijin, line 26: I think we shold redirect to the adopter foster dashboard here.

app/controllers/organizations/staff/people_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+  end
+  
+  def validate_inputs
```
kasugaijin, line 30: Does this get used?
One way we prevent them from using different emails is to pre-populate the email field with the `current_user.email` and disable it so they cannot edit it. Thought?

app/controllers/registrations_controller.rb
```diff
@@ -16,6 +16,10 @@ def create
     super do |resource|
       if resource.persisted?
         resource.add_role(:adopter, Current.organization)
+        # Create a new person unless it already exists
+        unless (Person.where(email: resource.email).any?)
```
kasugaijin, line 20:
```suggestion
        unless Person.exists?(email: resource.email)
```

app/controllers/root_controller.rb
```diff
@@ -1,5 +1,6 @@
 class RootController < ApplicationController
   skip_before_action :authenticate_user!
+  skip_before_action :verify_and_set_current_person
```
kasugaijin, line 3: You can remove this if you implement https://github.com/rubyforgood/homeward-tails/pull/1354/files#r2023859335

app/controllers/static_pages_controller.rb
```diff
@@ -1,5 +1,6 @@
 class StaticPagesController < ApplicationController
   skip_before_action :authenticate_user!
+  skip_before_action :verify_and_set_current_person
```
kasugaijin, line 3: Same here if you implement https://github.com/rubyforgood/homeward-tails/pull/1354/files#r2023859335

app/policies/adopter_application_policy.rb
```diff
@@ -21,7 +21,7 @@ def create?
   private
 
   def applicant?
-    user.person_id == record.person.id
+    user.person.id == record.person.id
```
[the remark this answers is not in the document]
kasugaijin: So `user.person` will return the person based on the tenant (organization), right? 
Makes me think we should add a validation to Person model to make sure we never duplicate a person record with the same `user_id` for a given org.

app/models/person.rb
```diff
@@ -36,8 +39,7 @@ class Person < ApplicationRecord
   accepts_nested_attributes_for :location,
     reject_if: ->(attributes) { attributes["city_town"].blank? }
   has_many :matches # , dependent: :destroy
-
-  has_one :user, dependent: :destroy
+  belongs_to :user, optional: true
```
kasugaijin, line 43: I think we need to add a validation like this to prevent multiple person records under a given Org scope having the same `user_id`

`validates :user_id, uniqueness: { scope: :organization_id }, allow_nil: true`

app/policies/organizations/like_policy.rb
kasugaijin, line 1: I think we can keep this method e.g.

`deny! if pet.organization_id != user.person.organization_id`

app/views/organizations/staff/people/_form.html.erb
kasugaijin, line 1: Same question as the associated controller - does it need to be scoped under staff? It's only giving the adopter role.

app/views/organizations/staff/people/_form.html.erb
```diff
@@ -0,0 +1,27 @@
+<div class="row">
+  <div class="col-md-8 mx-auto card p-5 bigger">
+    <%= bootstrap_form_with model: [:staff, person] do |form| %>
+
+      <%= form.hidden_field :user_id,
+                            value: current_user.id %>
+
+       <div class='form-group'>
+        <%= form.text_field :email, value: Current.user.email, readonly: true, class: 'form-control' %>
```
kasugaijin, line 9: for all cases we use `Current.user` in a controller or view, we can change to `current_user` as it's more conventional in Rails. Devise is so ubiquitous it's a common method to refer to.

app/views/organizations/staff/people/new.html.erb
```diff
@@ -0,0 +1,15 @@
...
+      <h1 class="h4 mb-5">
+        <%= "Join #{Current.organization.name}!" %>
+      </h1>
+    </div>
+
+    <!--main section-->
+    <%= render 'form', person: @person %>
+
+  </div>
+</section>
```
kasugaijin, line 15: Nit: new line

app/views/organizations/staff/staff/_role.html.erb
```diff
@@ -4,11 +4,12 @@
   </button>
 
   <div class="dropdown-menu"  aria-labelledby="dropdownMenuButton" >
-    <% if staff.has_role?(:super_admin, current_user.organization)%>
+    <!-- TODO: How to hanlde this? -->
```
kasugaijin, line 7: Is this comment still relevant? What's the question?

app/policies/adopter_application_policy.rb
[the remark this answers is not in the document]
BurkeMtn: Added.

app/controllers/application_controller.rb
BurkeMtn: Makes sense. Updated.

app/controllers/organizations/home_controller.rb
BurkeMtn: Yea, its required since the symbol is not defined after the refactor.

app/models/user.rb
BurkeMtn: Good call. Updated.

app/controllers/organizations/staff/invitations_controller.rb
BurkeMtn: Duh. Still getting used to the idioms.

app/controllers/organizations/staff/invitations_controller.rb
BurkeMtn: I don't think the else is needed. I prefer returning vs wrapping the rest of the function in a large else block. I get the following error if I don't include the return statement so will keep it for now (unless you are aware of another work around).

> AbstractController::DoubleRenderError: Render and/or redirect were called multiple times in this action. Please note that you may only call render OR redirect, and at most once per action. Also note that neither redirect nor render terminate execution of the action, so if you want to exit an action after redirecting, you need to do something like "redirect_to(...); return".

app/controllers/organizations/staff/people_controller.rb
BurkeMtn: Good catch. Removed. This is already happening in the view.

app/views/organizations/staff/staff/_role.html.erb
BurkeMtn: No longer needed.

app/policies/organizations/like_policy.rb
BurkeMtn: The roles are scoped to an organization so this should already be handled by the permission checks `permission?(:manage_likes)` right?

app/controllers/organizations/staff/people_controller.rb
BurkeMtn: Yea, that's right. I believe I just used the existing route and implemented the corresponding view & controller.

https://github.com/rubyforgood/homeward-tails/blob/main/config/routes.rb#L68-L70

I'm OK to change it. Directly under Organization does seem like a better fit.

app/controllers/organizations/adoptable_pets_controller.rb
```diff
@@ -3,6 +3,7 @@ class AdoptablePetsController < Organizations::BaseController
     include ::Pagy::Backend
 
     skip_before_action :authenticate_user!
+    skip_before_action :verify_and_set_current_person, only: %i[index]
```
BurkeMtn, line 6: The show action requires the person when the current_user despite the route being public hence only skip for the index action.

app/controllers/organizations/staff/people_controller.rb
kasugaijin: I think it's more intuitive to just be under `Organizations` otherwise I will think this has something to do with a staff action

app/policies/organizations/like_policy.rb
kasugaijin: Ok makes sense.

app/controllers/organizations/adoptable_pets_controller.rb
kasugaijin: Ah gotcha.

kasugaijin:
Looking good. It looks like just one test context needs a tweak. And `rails standard:fix` should fix the standard error in CI.

jmilljr24:
> An alternative option to allowing multiple roles, is that we only allow one role. If I have an adopter role for Org B, but later I get invited to either be staff or fosterer, we delete the adopter role. Thoughts? Tagging @jmilljr24 and @MooseCowBear here, too.

I think deleting the adopter role could cause more confusion and may have undesirable effects for the end user. For example, what happens to the adopters pending adoptions? 

> Another thought is that multiple roles is no issue. A user with e.g., adopter and staff roles can just access both sets of routes. Is this a problem? Off the top of my head I think, no. The only things to consider is in some places we have redirects based on role e.g. to dashboards, and we'd need to decide which dashboard we'd redirect to. In this case, redirecting to staff over adopter or fosterer makes sense.

Just poking around a bit, I don't see any major issues with allowing multiple roles. The "Dashboard" link on the org home page would need some thought if the user has both adopter and staff role. Defaulting to staff makes sense but we would need an easy way for the user to view their adopter dashboard.

On the adopter dashboard, the only link that would direct to the staff dashboard is edit settings which I think is fine. The rest stay within the adopter dashboard. The only thing I noticed in the public facing areas is the pet show page. We conditionally show the button to adopt vs manage the pet.

The biggest changes would be with the invitation flow but that needs to be fleshed out anyways. Some thoughts on that...

- User/Person has adopter role in org. Staff can click a button to add whatever new role to their account. They get email notification.
- User exists / No person in this org. Staff enters email to send invite token. User clicks link and creates their person in this org. Redirect to new role dashboard.
- No user / No person. Staff sends email invite token. Link to create user/person.
- We currently are not handling a person with no user so no problem there for now.

kasugaijin:
> I think deleting the adopter role could cause more confusion and may have undesirable effects for the end user. For example, what happens to the adopters pending adoptions?

Yeah you are right. I think multiple roles is the way to go. 

> The "Dashboard" link on the org home page would need some thought if the user has both adopter and staff role. 

Yeah this will need a refactor. I think it's fine to show two separate links e.g., `Dashboard` for adopters and fosteres, and `Admin Panel` for the staff dashboard. We can get fancy in the future and make it a dropdown or whatever UX makes sense, but for now two separate links should solve this. Now that we have `admin` and `superadmin` roles, I prefer the name `Admin Panel`, `Admin Dashboard`, or something along those lines to `Staff Dashboard`.

> On the adopter dashboard, the only link that would direct to the staff dashboard is edit settings which I think is fine. The rest stay within the adopter dashboard. 

Agreed. Not a problem, I think.

> The only thing I noticed in the public facing areas is the pet show page. We conditionally show the button to adopt vs manage the pet.

I guess, technically we can show both buttons at the same time, couldn't we?


> The biggest changes would be with the invitation flow but that needs to be fleshed out anyways. Some thoughts on that...
> * User/Person has adopter role in org. Staff can click a button to add whatever new role to their account. They get email notification.
> * User exists / No person in this org. Staff enters email to send invite token. User clicks link and creates their person in this org. Redirect to new role dashboard.
> * No user / No person. Staff sends email invite token. Link to create user/person.

Yeah I like where this is going.

kasugaijin:
@BurkeMtn just one failing test to review, and to run `rails standard:fix`. 
When they are done, @jmilljr24 and I (and @MooseCowBear if available) can approve. 
BEFORE MERGING - please set the base branch to user-person-refactor-base. You can set this by clicking edit at the top of this PR and selecting that branch. Do not merge into main branch. Let me know if you have questions.

image

BurkeMtn:
Ran `rails standard:fix` and set the base to `user-person-refactor-base`. Ran `bin/rails test` and were passing. Was there a particular test you wanted me to take a look at?

jmilljr24:
@BurkeMtn Can you run `bin/rails test:all`? Looks like we only have our github actions set to run the test's for merges on the main branch. [This](https://github.com/rubyforgood/homeward-tails/actions/runs/14296707344/job/40064807491) was what @kasugaijin was referring to but running test:all should* show some failing test's locally as well.

Also if you don't mind, can you merge the latest `upstream/main` to your branch and resolve any conflicts (if there are any) before merging?

kasugaijin (approved):
LGTM

PR #1362, opened 2025-02-09
Docker Compose Sass Error

jmilljr24:
# 🔗 Issue

# ✍️ Description

The sass service in docker was getting an error due to missing missing env variables. Updated accordingly.

I've also removed the `DATABASE_URL` variable. `ENV['DATABASE_URL']` is merged with and/or takes precedence over what is defined in `database.yml` ( [connection preference](https://guides.rubyonrails.org/configuring.html#connection-preference) ). 

Prior to #1353 , the primary database `postgres` was using `DATABASE_URL` to set the port and the db's `cable` and `queue` were using the default port 5432 from the postgres adapter. We are now explicitly defining ports etc. in `database.yml`  so I feel it make sense to keep in all on one place. This required adding a default host variable for docker to connect correctly.

# 📷 Screenshots/Demos

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -17,7 +17,7 @@
 default: &default
   adapter: postgresql
   encoding: unicode
-  host: localhost
+  host: <%= ENV.fetch("DEFAULT_HOST", "localhost") %>
   username: <%= ENV.fetch("DATABASE_USERNAME") %>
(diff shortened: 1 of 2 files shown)


kasugaijin (approved):
Lgtm

Issue #1365, opened 2025-02-10
Active Storage Image Errors in development


jmilljr24:
In development, we seed the db with pets. Each pet has an image attached. Some images will not display for the pet and produce a 404 not found error. Currently the same image file `hero.jpg` is uploaded and attached to each pet. No issues have been observed in production.

![Image](https://github.com/user-attachments/assets/e876ae28-d2be-49ea-a46b-87e598c32cd6)

**Investigative steps:**
- I wasn't able to reproduce the error when I seeded via the console.
- I do not observe any errors prior to the Rails 8 upgrade #1241 
- In rails console, a pet with an image error will show that is has a blob attached and a valid checksum but if you attempt to download it you will receive `ActiveStorage::FileNotFoundError`

![Image](https://github.com/user-attachments/assets/50dad37f-566f-4ab7-aae3-b4a6be02c2c7)

Creating a single blob and having all pets associated with it does resolve the issue but doesn't explain the the problem. I think its a better way to handle the images in development so we are not duplicating the same file hundreds of time, but my concern is there may be some config issue we may not know about and it might crop up in production.

PR #1366, opened 2025-02-10
Development Image Errors: Attach same blob to each pet

jmilljr24:
# 🔗 Issue

#1365 

# ✍️ Description

This eliminates the random broken pet images in development however it does not address the root cause. I created an issue which we might consider keeping open in case someone can address the real issue and/or we get the same issue in production when more data is being added.

Overall I still think this is a good change for the seed data. This creates a single blob from a single image file. Each pet then associates with that one blob. Currently we take the single `hero.jpg` image and create/upload a blob and new file on disk over and over again for each pet even though it is the same image. Of course this is not what we do in production per say, but I don't think we need 200 different dummy images just for seed data. 

# 📷 Screenshots/Demos

### DIFF
--- a/db/seeds/01_alta.rb
+++ b/db/seeds/01_alta.rb
@@ -192,6 +192,12 @@ ActsAsTenant.with_tenant(@organization) do
   )
 
   path = Rails.root.join("app", "assets", "images", "hero.jpg")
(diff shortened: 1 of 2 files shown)


kasugaijin (approved):
yeah this is a good start and way more efficient!

PR #1371, opened 2025-02-12
Add dummy variables for production assets precompile

jmilljr24:
# 🔗 Issue

# ✍️ Description

This was causing the build to fail. The previous changes made are fine to keep.

# 📷 Screenshots/Demos


kasugaijin (approved):
Nice catch

PR #1373, opened 2025-02-12
Add default db username and password back

jmilljr24:
# 🔗 Issue

# ✍️ Description

I am unable to figure out where in the kamal deployment `bin/rails` is being run which causes the failure when the env variables are not set. 

The assets:precompile was one location which I was able to replicate locally and resolve. I was able to deploy to production from my machine successfully. However, somewhere in the deployment flow, my local `DATABASE_USERNAME` was being read. When I removed it, the deployment failed the same as from the github [action.](https://github.com/rubyforgood/homeward-tails/actions/runs/13285027555/job/37091740951#step:7:8)

Adding the default username and password back to `database.yml` is the simplest solution at this time and removes the need to add dummy values in multiple places. This also removes the risk of someone mistakenly updating a dummy value with a secret and committing it. 

# 📷 Screenshots/Demos

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,7 +51,7 @@ COPY . .
 RUN bundle exec bootsnap precompile app/ lib/
 
 # Precompiling assets for production without requiring secret RAILS_MASTER_KEY
-RUN SECRET_KEY_BASE_DUMMY=1 PRECOMPILING=1 ./bin/rails assets:precompile
+RUN SECRET_KEY_BASE_DUMMY=1 PRECOMPILING=1 DATABASE_USERNAME=dummy ./bin/rails assets:precompile
 
 # Final stage for app image
 FROM base
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,7 +51,7 @@ COPY . .
 RUN bundle exec bootsnap precompile app/ lib/
 
 # Precompiling assets for production without requiring secret RAILS_MASTER_KEY
-RUN SECRET_KEY_BASE_DUMMY=1 PRECOMPILING=1 DATABASE_USERNAME=dummy ./bin/rails assets:precompile
+RUN SECRET_KEY_BASE_DUMMY=1 PRECOMPILING=1 DATABASE_USERNAME=dummy DATABASE_PASSWORD=dummy ./bin/rails assets:precompile
 
 # Final stage for app image
 FROM base
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,7 +51,7 @@ COPY . .
 RUN bundle exec bootsnap precompile app/ lib/
 
 # Precompiling assets for production without requiring secret RAILS_MASTER_KEY
-RUN SECRET_KEY_BASE_DUMMY=1 PRECOMPILING=1 DATABASE_USERNAME=dummy DATABASE_PASSWORD=dummy ./bin/rails assets:precompile
+RUN SECRET_KEY_BASE_DUMMY=1 PRECOMPILING=1 ./bin/rails assets:precompile
 
 # Final stage for app image
 FROM base
--- a/config/database.yml
+++ b/config/database.yml
@@ -18,8 +18,8 @@ default: &default
   adapter: postgresql
   encoding: unicode
   host: <%= ENV.fetch("DEFAULT_HOST", "localhost") %>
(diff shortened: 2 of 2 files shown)


kasugaijin (approved):

mononoken:
@jmilljr24 @kasugaijin I don't use Kamal, so I apologize if I have any misunderstandings. However, this seems like a big red flag to me, and it highlights another good reason to not have the default. It sounds like y'all are using these defaults on production. I may just be confused though.

I think there is some confusion happening here with the dev setup vs the prod setup. I see the following and hope it helps. Let me know if I can help dig more:

1. I see Kamal is setting env variables names in `﻿﻿.kamal/secrets` that do not match the variable names being handled in `database.yml`
2. Kamal is using `Rails.application.credentials` while the `database.yml `is using `ENV` (they are not the same thing)
3. `database.yml` has development and production sharing the logic of setting `DATABASE_USERNAME` and `DATABASE_PASSWORD` despite it looking like we are storing these variables in separate methodologies.
4. I don't think dummy settings for these should be necessary. The production environment should have database env variables for things to work in the first place. 

I think coming up with a solution to clean this situation up will help solve this issue. My gut tells me we should pick one or the other for secret handling, not use both and clean up the other inconsistencies mentioned. Perhaps there is a reason for all of this though.

jmilljr24:
@mononoken Thanks for taking a look!! I could very well be mistaken on some things so bear with me. I did recently update the rails production credentials and `.kamal/secrets` to be consistent and use `DATABASE_USERNAME` and `DATABASE_PASSWORD`. 

1. & 2. `Rails.application.credentials` have been updated to use `database_username` for consistency but it is inconsequential as long as its referenced correctly. Kamal will set these in `.kamal/secrets` for example `DATABASE_USERNAME=$(RAILS_ENV=production KEY=deploy,database_username, bin/rails credentials:read)` . This variable `DATABASE_USERNAME` must be set in `config/deploy.yml` for it to be used in the docker container that is deployed. `deploy.yml > .kamal/secrets > Rails.application.credentials`

3. Since we updated kamal to use `DATABASE_USERNAME` for production, we can now use the default anchor in `database.yml` for production for username and password instead of overriding with the previous used variables `DB_USER` and `DB_PASSWORD`. The `ENV` is set via `.kamal/secrets` during deployment.

4. The reason I believe a dummy variable is required is due to the fact `database.yml` is read any time `bin/rails` is run. Even if `production` doesn't use `default` there will still be an error because the entire file is read. Kamal does not set the `ENV` variable until later in the deployment flow. For example in `Dockerfile` (used for production) we have `RUN SECRET_KEY_BASE_DUMMY=1 PRECOMPILING=1 ./bin/rails assets:precompile
`. The `SECRET_KEY_BASE_DUMMY` is a special `ENV` that allows precompiling assets for production without requiring secret RAILS_MASTER_KEY. If I build the production container locally and set the `ENV` for username and password on my machine, the precompile will still fail without dummy values. (real values could be passed but is not convention or required)

So I have confirmed the dummy variable is required specifically for the precompile if we do not have a default fallback for the `ENV` in database.yml. The issue I cannot figure out is where in the `kamal deploy` flow that `bin/rails` is called. It is before the precompile. This if from the github workflow and I replicated locally while attempting to deploy to production from my machine. Setting verbose logging does not yield more information.

Screenshot from 2025-02-12 17-21-26


So at minimum a dummy value would be required for precompile. My suspicous is that it is required in the deployment flow prior to that as well. I do agree with your prior point about the potential for new developers to be confused if defaults as set but I feel production takes priority. Also you did a great job updated the doc's so headaches should be minimal!! 

I'd love to dive into this deeper but I have already skimmed through the source for kamal and my next step would be to debug in their code directly.

mononoken:
Thank you for the thorough write up, @jmilljr24 !

Sorry about the mixup with the environments. I must not have pulled your latest changes when I checked during lunch. Thanks for fixing those.

To be clear, the main thing I dislike with the defaults is it is a critical config variable and think it should be very explicit. But the issue of these bin/rails being called and crashing does sound very troublesome. It does not sound like it is worth the effort to remove them. Not like they are the actual value being used on the db, and it sounds like the db is working fine in prod.

Thank you for your hard work and taking care of many of these deploy tasks!

PR #1376, opened 2025-02-13
Export fosterer emails to CSV

caitlynl22:
# 🔗 Issue

https://github.com/rubyforgood/homeward-tails/issues/1338

# ✍️ Description

Adds a button allowing users to export fosterer emails, preserving search terms or filters on the manage fosters dashboard. Also adds tests for filters on manage fosters dashboard.

# 📷 Screenshots/Demos

Screenshot 2025-02-13 at 8 55 44 PM

### DIFF
--- a/app/views/organizations/staff/fosterers/index.html.erb
+++ b/app/views/organizations/staff/fosterers/index.html.erb
@@ -6,6 +6,37 @@
     </div>
   <% end %>
   <% c.with_body do %>
+    <!-- Ransack Search Form -->
+    <div class="mx-1 row p-4 border rounded-5 mb-5 bg-white">
+      <div class="col-xl-12">
+        <%= search_form_for @q, url: staff_fosterers_path do |f| %>
+          <div class="row">
+            <div class="form-group mb-3 col-md-4">
+              <%= f.label :email_cont, "Email" %>
+              <%= f.search_field :email_cont,
+                class: "form-control",
+                placeholder: "Enter Email"
+              %>
+            </div>
+            <div class="form-group mb-3 col-md-4">
+              <%= f.label :first_name_or_last_name_cont, "Fosterer Name" %>
+              <%= f.search_field :first_name_or_last_name_cont,
+                class: "form-control",
+                placeholder: "Enter Fosterer Name"
+              %>
+            </div>
+          </div>
+          <div class="row">
+            <div class="col-md-12 text-right">
+              <%= f.submit "Search", class: "btn btn-primary" %>
+              <%= link_to "Clear filters", staff_fosterers_path,
+                class: "btn btn-default"
+              %>
+            </div>
+          </div>
+        <% end %>
+      </div>
+    </div>
     <div class="d-flex justify-content-center align-items-center mt-2">
       <%== pagy_bootstrap_nav(@pagy) if @pagy.pages > 1 %>
     </div>
--- a/app/views/organizations/staff/fosterers/index.html.erb
+++ b/app/views/organizations/staff/fosterers/index.html.erb
@@ -7,9 +7,9 @@
   <% end %>
   <% c.with_body do %>
     <% if params[:q] %>
-      <%= link_to 'Export CSV', staff_fosterers_path(request.parameters.merge({format: 'csv'})), class: "btn btn-secondary mb-4" %>
+      <%= link_to t('.export_emails'), staff_fosterers_path(request.parameters.merge({format: 'csv'})), class: "btn btn-secondary mb-4", data: {turbo: false} %>
     <% else %>
-      <%= link_to "Export CSV", staff_fosterers_path(format: "csv"),
+      <%= link_to t('.export_emails'), staff_fosterers_path(format: "csv"),
         class: "btn btn-secondary mb-4"
       %>
     <% end %>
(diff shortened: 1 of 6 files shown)


app/controllers/organizations/staff/manage_fosters_controller.rb
jmilljr24, line 1: Changes should be in `Organizations::Staff::FosterersController`

app/models/match.rb
```diff
@@ -106,6 +108,16 @@ def retire_applications(application_class: AdopterApplication)
...
 
+  def self.to_csv(matches)
+    CSV.generate(headers: true) do |csv|
+      csv << ["Email"]
+
+      matches.each do |match|
+        csv << [match.person.email]
+      end
+    end
+  end
```
jmilljr24, line 119: This can probably just be in the controller. Although maybe a concern would be better as we may use it other place in the future. Thoughts? Something like `exportable`

jmilljr24 (changes requested):
@caitlynl22 This works great! However, it looks like there was a little mix up on which page we wanted this feature. It's a little confusing but we have `Fosterers` and `In Foster`. The page you made the change on deals with when a "Fosterer" is matched with a pet and placed "in foster".

We are looking to have this feature on the `Fosterers` page which lists all "Fosterers" in the organization, regardless of if they are currently fostering a pet. 

I really like the idea of exporting the filtered results. Currently the `Fosterers` page does not have a filter. `Adopters` page does have a filter so it is probably worth while to add that to `Fosterers` as well. That is outside the scope of this issue but if you are interested, feel free to add a filter with your changes in this PR and continue to export the filtered results. If you do not want to, then export all "Fosterers" and we will open a separate issue for a filter.

app/controllers/organizations/staff/manage_fosters_controller.rb
caitlynl22: Oh yeah, that was my bad, sorry about that! I'll make that change.

app/models/match.rb
caitlynl22: Yeah, that makes sense to make it more general and put it in a concern.

caitlynl22:
> @caitlynl22 This works great! However, it looks like there was a little mix up on which page we wanted this feature. It's a little confusing but we have `Fosterers` and `In Foster`. The page you made the change on deals with when a "Fosterer" is matched with a pet and placed "in foster".
> 
> We are looking to have this feature on the `Fosterers` page which lists all "Fosterers" in the organization, regardless of if they are currently fostering a pet.
> 
> I really like the idea of exporting the filtered results. Currently the `Fosterers` page does not have a filter. `Adopters` page does have a filter so it is probably worth while to add that to `Fosterers` as well. That is outside the scope of this issue but if you are interested, feel free to add a filter with your changes in this PR and continue to export the filtered results. If you do not want to, then export all "Fosterers" and we will open a separate issue for a filter.

@jmilljr24 I'm fine with adding the filters in this PR, I assume we'd filter on the same attributes? Name and email?

jmilljr24:
> > @caitlynl22 This works great! However, it looks like there was a little mix up on which page we wanted this feature. It's a little confusing but we have `Fosterers` and `In Foster`. The page you made the change on deals with when a "Fosterer" is matched with a pet and placed "in foster".
> > We are looking to have this feature on the `Fosterers` page which lists all "Fosterers" in the organization, regardless of if they are currently fostering a pet.
> > I really like the idea of exporting the filtered results. Currently the `Fosterers` page does not have a filter. `Adopters` page does have a filter so it is probably worth while to add that to `Fosterers` as well. That is outside the scope of this issue but if you are interested, feel free to add a filter with your changes in this PR and continue to export the filtered results. If you do not want to, then export all "Fosterers" and we will open a separate issue for a filter.
> 
> @jmilljr24 I'm fine with adding the filters in this PR, I assume we'd filter on the same attributes? Name and email?

Yes! Thank you

caitlynl22:
@jmilljr24 this should be all set now!

app/views/organizations/staff/fosterers/index.html.erb
```diff
@@ -6,6 +6,44 @@
     </div>
   <% end %>
   <% c.with_body do %>
+    <% if params[:q] %>
+      <%= link_to 'Export CSV', staff_fosterers_path(request.parameters.merge({format: 'csv'})), class: "btn btn-secondary mb-4" %>
```
jmilljr24, line 10:
```suggestion
      <%= link_to 'Export CSV', staff_fosterers_path(request.parameters.merge({format: 'csv'})), class: "btn btn-secondary mb-4",
      data: {turbo: false} %>
```

jmilljr24 (approved):

jmilljr24:
@caitlynl22 Awesome!! The only minor thing I noticed is that when you click export for a filtered list, the turbo progress bar displays and doesn't go away. It should be as simple as adding `data: {turbo: false}` to the link. No sure why it's only on the filtered link not all. If you don't mind fixing that, feel free to merge this after you push the change.

[Screencast from 2025-02-14 08-57-17.webm](https://github.com/user-attachments/assets/1e36cfcb-623b-43f9-b044-2e8c4fea5c78)

app/views/organizations/staff/fosterers/index.html.erb
kasugaijin, line 10: I think we can be a bit more specific in the button text to say `Export Emails` instead. Otherwise, looks good.

jmilljr24: Good idea. Also we should use i18n while we're at it!

caitlynl22:
@jmilljr24 Looks like I don't have write access, but this should be good to be merged.

Issue #1402, opened 2025-04-06
Relocate the Google Authentication option in Log In and Sign Up forms


kasugaijin:
Currently, the google authentication option is located below the form inputs to log in or sign up. This means a user could fill out the fields and then see the Google Auth option, which is a little frustrating if you didn't want to fill out the fields.

So, to fix this, we can move the Contine with Google button above the form fields. We should make this change on all log in and sign up gotms, and place them under the Log In and Sign Up headers so it's something like:

```
Sign Up

Continue with Google

<horizontal rule>or</hr>

Form Fields
```

<img width="485" alt="Image" src="https://github.com/user-attachments/assets/e0b9caa3-4970-4eb9-95f9-6392b394a9b0" />

Gabe-Torres:
I'll take this one! 🐶

PR #1405, opened 2025-04-11
1402 relocate google auth option

Gabe-Torres:
# 🔗 Issue
#1402

# ✍️ Description
- Moves the Google Auth button in both "app/views/devise/sessions/new.html.erb" and "app/views/devise/registrations/new.html.erb" under the Log in and Sign Up headers

- Move hortizontal rule and "or" divider under Google Auth Button

- removes Google Auth button from devise shared links partial ("app/views/devise/shared/_links.html.erb")

# 📷 Screenshots/Demos

<h3> Before </h3>

Screenshot 2025-04-10 at 10 35 44 PM
Screenshot 2025-04-10 at 10 35 54 PM

<h3> After </h3>

Screenshot 2025-04-10 at 11 17 03 PM
Screenshot 2025-04-10 at 11 16 47 PM

### DIFF
--- a/app/views/devise/registrations/new.html.erb
+++ b/app/views/devise/registrations/new.html.erb
@@ -17,6 +17,8 @@
             <h1 class="mb-1 fw-bold">Sign up</h1>
           </div>
 
+          <%= render "devise/shared/google_oauth_button" %>
+
           <%= bootstrap_form_for(resource, as: resource_name, url: registration_path(resource_name)) do |f| %>
             <div class="form-group mb-3 bigger">
               <%= f.email_field :email,
--- a/app/views/devise/shared/_google_oauth_button.html.erb
+++ b/app/views/devise/shared/_google_oauth_button.html.erb
@@ -1,0 +1,21 @@
+<%- if devise_mapping.omniauthable? %>
+    <%- resource_class.omniauth_providers.each do |provider| %>
+      <div class="d-grid gap-1 mb-1">
+        <%= link_to user_google_oauth2_omniauth_authorize_path, 
+          method: :post, 
+          data: { turbo: 'false' },
+          class: "btn border w-100 d-flex align-items-center justify-content-center gap-2",
+          onclick: "this.innerHTML='<span class=\"spinner-border spinner-border-sm me-2\" role=\"status\" aria-hidden=\"true\"></span>Signing in...'" do %>
+          <%= image_tag "google.svg", size: "20x20", alt: "Google" %>
+          Continue with Google
+        <% end %>
+        <div class="position-relative mt-4">
+          <hr class="bg-gray-300">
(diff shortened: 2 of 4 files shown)


app/views/devise/registrations/new.html.erb
```diff
@@ -16,6 +16,26 @@
             </div>
             <h1 class="mb-1 fw-bold">Sign up</h1>
           </div>
+          <%- if devise_mapping.omniauthable? %>
+            <%- resource_class.omniauth_providers.each do |provider| %>
+              <div class="d-grid gap-1 mb-1">
```
kasugaijin, line 21: I see why you moved this google oauth button from the `app/views/devise/shared/_links.html.erb` because we render that at the bottom of this form. However we do have some duplication going on here for the button itself, so I think a partial would be beneficial here. 

Could we move the repeated logic to a partial? I think we can create the partial in `app/views/devise/shared` and call it something like `_google_oauth.html.erb` or the likes. 

Thoughts?

Gabe-Torres: I agree, a partial here is the way to go. Keeps it DRY. I did notice that one of the buttons is wrapped in a conditional and using a `devise_mapping.omniauthable` helper method. I don't think it makes a difference and will work anywhere. I'll go ahead and use that format as it seems to add another layer of security unless you think otherwise. 👍🏽

app/views/devise/shared/_google_oauth_button.html.erb
```diff
@@ -0,0 +1,20 @@
...
+        <% end %>
+        <div class="position-relative mt-4">
+          <hr class="bg-gray-300">
+            <p class="position-absolute top-50 start-50 translate-middle bg-white px-3 text-muted small">
+              or
+            </p>
+        </div>
+      </div>
+    <% end %>
+  <% end %>
```
kasugaijin, line 20: new line :)

kasugaijin (approved):

app/views/devise/shared/_google_oauth_button.html.erb
Gabe-Torres: Added the line! I ran a `git rebase -1 HEAD~4` and picked my head commit to squash the others into it. For some reason, one of the commits failed to squash and couldn't rebase how I intended. I noticed that the commit message references a pr from 2023? I'm not sure where that came from

Gabe-Torres: HMM seems I've done it again. In my attempts to clean the commit history up, I've made a mess. Apologies. Thought I had the method down! Whelp. I created a new pr and linked it in the issue ticket.

kasugaijin:
Haha no worries. I will close this PR - the new PR is merged.

PR #1412, opened 2025-04-24
Person group with deactivate

jmilljr24:
# 🔗 Issue

# ✍️ Description

# 📷 Screenshots/Demos

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -14,7 +14,7 @@ class ApplicationController < ActionController::Base
       redirect_to edit_tos_agreement_path
     else
       flash[:alert] = t("errors.authorization_error")
-      redirect_back_or_to root_path
+      redirect_to root_path
     end
   end
 
--- a/app/controllers/organizations/activations_controller.rb
+++ b/app/controllers/organizations/activations_controller.rb
@@ -1,23 +1,36 @@
 module Organizations
   class ActivationsController < Organizations::BaseController
+    before_action :set_pg
+
     def update
-      @user = User.find(params[:user_id])
-
-      authorize! @user, with: ActivationsPolicy
-
-      if @user.deactivated_at
-        @user.activate
+      if params[:person_group][:deactivated] == "true"
+        @pg.update!(deactivated_at: Time.now)
+      elsif params[:person_group][:deactivated] == "false"
+        @pg.update!(deactivated_at: nil)
       else
-        @user.deactivate
+        respond_to do |format|
+          format.html { redirect_back_or_to staff_staff_index_path, alert: t("errors.try_again)") }
+          format.turbo_stream { flash.now[:alert] = t("errors.try_again") and return }
+        end
       end
 
       respond_to do |format|
-        success = @user.deactivated_at.nil? ?
-          t(".activated", staff: @user.full_name) :
-          t(".deactivated", staff: @user.full_name)
+        success = @person.active_in_group?(@group.name) ?
+          t(".activated", staff: @person.full_name) :
+          t(".deactivated", staff: @person.full_name)
         format.html { redirect_to staff_staff_index_path, notice: success }
         format.turbo_stream { flash.now[:notice] = success }
       end
+    end
+
+    private
+
+    def set_pg
+      @pg = PersonGroup.find(params[:id])
+      @group = Group.find(@pg.group_id)
+      @person = Person.find(@pg.person_id)
+
+      authorize! @person, with: ActivationsPolicy, context: {group: @group.name}
     end
   end
 end
--- a/app/controllers/organizations/activations_controller.rb
+++ b/app/controllers/organizations/activations_controller.rb
@@ -5,8 +5,10 @@ module Organizations
     def update
       if params[:person_group][:deactivated] == "true"
         @pg.update!(deactivated_at: Time.now)
+        @person.user.remove_role(@group.name, Current.organization)
       elsif params[:person_group][:deactivated] == "false"
         @pg.update!(deactivated_at: nil)
+        @person.user.add_role(@group.name, Current.organization)
       else
         respond_to do |format|
           format.html { redirect_back_or_to staff_staff_index_path, alert: t("errors.try_again)") }
--- a/app/controllers/organizations/adopter_fosterer/adopter_applications_controller.rb
+++ b/app/controllers/organizations/adopter_fosterer/adopter_applications_controller.rb
@@ -19,6 +19,7 @@ class Organizations::AdopterFosterer::AdopterApplicationsController < Organizati
         notice: t(".success", message: MessagesHelper.affirmations.sample)
 
       # mailer
+      # TODO change to Person.staff (or active staff)
       @org_staff = User.staff
       StaffApplicationNotificationMailer.with(pet: @pet,
         organization_staff: @org_staff)
--- a/app/controllers/organizations/adopter_fosterer/fostered_pets_controller.rb
+++ b/app/controllers/organizations/adopter_fosterer/fostered_pets_controller.rb
@@ -5,6 +5,7 @@ module Organizations
       layout "adopter_foster_dashboard"
 
       def index
+        # TODO adopter can visit this page
         @fostered_pets = authorized_scope(Match.fosters.current.or(Match.fosters.upcoming), with: Organizations::AdopterFosterer::MatchPolicy)
       end
 
--- a/app/controllers/organizations/staff/invitations_controller.rb
+++ b/app/controllers/organizations/staff/invitations_controller.rb
@@ -54,20 +54,26 @@ class Organizations::Staff::InvitationsController < Devise::InvitationsControlle
       authorize! User, context: {organization: Current.organization},
         with: Organizations::FostererInvitationPolicy
 
-      @user = User.new(
-        user_params.merge(password: SecureRandom.hex(8)).except(:roles)
-      )
-      @user.add_role("adopter", Current.organization)
-      @user.add_role("fosterer", Current.organization)
+      @user = User.find_by(email: user_params[:email])
 
-      if @user.save
-        @user.invite!(current_user)
-        create_person(@user)
-        @person.add_group(:adopter, :fosterer)
-        redirect_to staff_fosterers_path, notice: t(".success")
-      else
-        render "organizations/staff/fosterer_invitations/new", status: :unprocessable_entity
+      if @user.nil?
+        @user = User.new(
+          user_params.merge(password: SecureRandom.hex(8)).except(:roles)
+        )
+        if @user.save
+          @user.invite!(current_user)
+        else
+          render "organizations/staff/fosterer_invitations/new", status: :unprocessable_entity
+        end
       end
+
+      if @user.person.nil?
+        create_person(@user)
+      end
+
+      @user.person.add_role_and_group(:adopter, :fosterer)
+      redirect_to staff_fosterers_path, notice: t(".success")
+
     else
       authorize! User, context: {organization: Current.organization},
         with: Organizations::InvitationPolicy
--- a/app/controllers/organizations/staff/invitations_controller.rb
+++ b/app/controllers/organizations/staff/invitations_controller.rb
@@ -25,26 +25,31 @@ class Organizations::Staff::InvitationsController < Devise::InvitationsControlle
       # TODO: The user should probably get an email letting them know they
       # were added to a new organization.
       @user = User.find_by(email: user_params[:email])
-      if @user.present?
-        if !@user.has_role?(user_params[:roles], Current.organization)
-          @user.add_role(user_params[:roles], Current.organization)
-          redirect_to staff_staff_index_path, notice: t(".success")
-          return
+
+      if @user.nil?
+        @user = User.new(
+          user_params.merge(password: SecureRandom.hex(8)).except(:roles)
+        )
+        if @user.save
+          @user.invite!(current_user)
+        else
+          render "organizations/staff/staff_invitations/new", status: :unprocessable_entity
         end
       end
 
-      @user = User.new(
-        user_params.merge(password: SecureRandom.hex(8)).except(:roles)
-      )
-      @user.add_role(user_params[:roles], Current.organization)
-
-      if @user.save
-        @user.invite!(current_user)
+      if @user.person.nil?
         create_person(@user)
-        redirect_to staff_staff_index_path, notice: t(".success")
-      else
-        render "organizations/staff/staff_invitations/new", status: :unprocessable_entity
       end
+
+      if @user.person.is_staff?
+        redirect_to staff_staff_index_path, notice: "User is already Staff in this organization"
+      else
+        @user.person.add_or_change_staff_role_and_group(user_params[:roles])
+        # TODO send notice to email
+        redirect_to staff_staff_index_path, notice: t(".success")
+
+      end
+
     when "fosterer"
       authorize! User, context: {organization: Current.organization},
         with: Organizations::FostererInvitationPolicy
@@ -58,6 +63,7 @@ class Organizations::Staff::InvitationsController < Devise::InvitationsControlle
       if @user.save
         @user.invite!(current_user)
         create_person(@user)
+        @person.add_group(:adopter, :fosterer)
         redirect_to staff_fosterers_path, notice: t(".success")
       else
         render "organizations/staff/fosterer_invitations/new", status: :unprocessable_entity
@@ -104,8 +110,9 @@ class Organizations::Staff::InvitationsController < Devise::InvitationsControlle
   end
 
   def create_person(user)
+    # TODO need to handle this error
     unless Person.exists?(email: user.email)
-      Person.create!(user_id: user.id, first_name: user.first_name, last_name: user.last_name, email: user.email)
+      @person = Person.create!(user_id: user.id, first_name: user.first_name, last_name: user.last_name, email: user.email)
     end
   end
 end
--- a/app/controllers/organizations/staff/invitations_controller.rb
+++ b/app/controllers/organizations/staff/invitations_controller.rb
@@ -51,6 +51,7 @@ class Organizations::Staff::InvitationsController < Devise::InvitationsControlle
       end
 
     when "fosterer"
+      # TODO upremitted param
       authorize! User, context: {organization: Current.organization},
         with: Organizations::FostererInvitationPolicy
 
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -11,13 +11,15 @@ class RegistrationsController < Devise::RegistrationsController
     respond_with resource
   end
 
-  # MARK: only adopters are created through this route. Adopters need both the adoper role and a form submission to attach their applications to
+  # MARK: only adopters are created through this route.
   def create
     super do |resource|
       if resource.persisted?
-        resource.add_role(:adopter, Current.organization)
+        # TODO Currently a person shouldn't exist without a user with the same email. If the person exists (but no user),
+        # how should we be handling this newly created user?
         unless Person.exists?(email: resource.email)
-          Person.create!(user_id: resource.id, first_name: resource.first_name, last_name: resource.last_name, email: resource.email)
+          person = Person.create!(user_id: resource.id, first_name: resource.first_name, last_name: resource.last_name, email: resource.email)
+          person.add_role_and_group(:adopter)
         end
       end
     end
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -1,3 +1,22 @@
+# == Schema Information
+#
+# Table name: groups
+#
+#  id              :bigint           not null, primary key
+#  name            :integer          not null
+#  created_at      :datetime         not null
+#  updated_at      :datetime         not null
+#  organization_id :bigint           not null
+#
+# Indexes
+#
+#  index_groups_on_name_and_organization_id  (name,organization_id) UNIQUE
+#  index_groups_on_organization_id           (organization_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (organization_id => organizations.id)
+#
 class Group < ApplicationRecord
   acts_as_tenant(:organization)
 
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -1,0 +1,7 @@
+class Group < ApplicationRecord
+  acts_as_tenant(:organization)
+
+  has_many :person_groups
+  has_many :people, through: :person_groups
+  enum :name, %i[adopter fosterer admin super_admin]
+end
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -76,4 +76,53 @@ class Person < ApplicationRecord
       raise ArgumentError, "Unsupported format: #{format}"
     end
   end
+
+  def add_role_and_group(name)
+    transaction do
+      user.add_role(name, Current.organization)
+      add_group(name)
+    end
+  end
+
+  def add_group(*args)
+    args.map(&:to_s).uniq.each do |name|
+      next unless Group.names.key?(name)
+      group = Group.find_or_create_by!(name: name)
+
+      unless groups.exists?(id: group.id)
+        person_groups.create!(group: group)
+      end
+    end
+  end
+
+  def is_staff?
+    groups.exists?(name: %i[admin super_admin])
+  end
+
+  def add_or_change_staff_role_and_group(new_group, prev_group = nil)
+    transaction do
+      user.change_role(prev_group, new_group)
+
+      # Remove existing admin/super_admin groups
+      groups.where(name: [:admin, :super_admin]).each do |group|
+        groups.destroy(group)
+      end
+
+      add_group(new_group)
+    end
+  end
+
+  def deactivate_group(group_name)
+    person_groups
+      .joins(:group)
+      .find_by(groups: {name: group_name.to_s})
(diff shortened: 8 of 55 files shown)


app/controllers/application_controller.rb
```diff
@@ -14,7 +14,7 @@ class ApplicationController < ActionController::Base
       redirect_to edit_tos_agreement_path
     else
       flash[:alert] = t("errors.authorization_error")
-      redirect_back_or_to root_path
```
jmilljr24, line 17: This had the potential to create an infinite loop. 

For example, a staff user is deactivated while they are on the staff dashboard. If they click any link or action requiring "active staff" the authorization will fail and they will keep getting redirected back to the staff dashboard which they do not have authorization for.

jmilljr24:
@kasugaijin or anyone else that want's to take a look. I think this is a solid start. No major issues (other than the 500+ tests I have to fix)

I wrote down or put TODO's on anything that seems note worthy. Some things can be addressed in a separate PR. 

Let me know your thoughts before I tackle the tests!!

kasugaijin:
Nice @jmilljr24 I will take a look this weekend!

app/controllers/organizations/activations_controller.rb
```diff
@@ -1,23 +1,38 @@
...
+        @person.user.remove_role(@group.name, Current.organization)
+      elsif params[:person_group][:deactivated] == "false"
+        @pg.update!(deactivated_at: nil)
+        @person.user.add_role(@group.name, Current.organization)
       else
-        @user.deactivate
+        respond_to do |format|
+          format.html { redirect_back_or_to staff_staff_index_path, alert: t("errors.try_again)") }
+          format.turbo_stream { flash.now[:alert] = t("errors.try_again") and return }
+        end
```
jmilljr24, line 16: This really is only handling a invalid param, there is not a check if the group/role didn't change. I could do it with repeating multiple respond_to's but there should be a better way that's DRY.

app/controllers/organizations/adopter_fosterer/adopter_applications_controller.rb
jmilljr24, line 1: I changed this to use nested modules and it broke some things so that can be revisited at a later date.

test/factories/users.rb
jmilljr24, line 1: This will change the way we test users/people. Previously we would call `create(:adopter)` which would create a user with the adopter role. This change allows a more generic create/build of user and person. A user isn't just an adopter, they can be a user with the adopter role in Org A and a user with the admin role in Org B. This also simplifies the creation of person with groups/roles. A person (with user role adopter) can have a group (adopter).

If we are simply testing a user's role we create the user with the role. If we need to handle groups/roles/deactivations, we need to create a person with a user (and role) with a group (person_group deactivated: Time.now)

db/seeds/01_alta.rb
kasugaijin, line 1: I am guessing you had to manually add the groups (adopter, fosterer, admin, super admin) in the console in your local environment? I think it would be good to have a separate seed file that we run under `bin/setup` that will add these to the dev's local db.

app/models/group.rb
```diff
@@ -0,0 +1,26 @@
...
+#
+#  index_groups_on_name_and_organization_id  (name,organization_id) UNIQUE
+#  index_groups_on_organization_id           (organization_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (organization_id => organizations.id)
+#
+class Group < ApplicationRecord
+  acts_as_tenant(:organization)
```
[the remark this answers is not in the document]
kasugaijin: Yeah I think scoping to Org works.

app/models/person.rb
```diff
@@ -74,4 +86,91 @@ def full_name(format = :default)
...
+        user.add_role(name, Current.organization)
+      end
+      add_group(*names)
+    end
+  end
+
+  def add_group(*names)
+    names.map(&:to_s).uniq.each do |name|
+      next unless Group.names.key?(name)
+      group = Group.find_or_create_by!(name: name)
```
kasugaijin, line 102: You'll have to remind me, if I call this within the Org scoped controllers where we set a tenant context, this will add the `org_id` to the `group` record? But, if I call it outside of that context e.g., in the console, it won't have an org scope and won't know what `org_id` to use, right? So, do we need to handle that ability?
`Group.find_or_create_by!(name: name)`

I am looking at this late on Saturday night, so I could be derping here.

app/models/person.rb
kasugaijin, line 1: What do you think about creating a concern, or one could argue a service object, to put all the group and role related logic? I think concerns work, but open to other suggestions.

app/models/person.rb
kasugaijin: Yeah I think I am just reminding myself of a fundamental approach of using ActsAsTenant...it has been a while.

app/controllers/organizations/adopter_fosterer/fostered_pets_controller.rb
```diff
@@ -5,6 +5,7 @@ class FosteredPetsController < Organizations::BaseController
       layout "adopter_foster_dashboard"
 
       def index
+        # TODO adopter can visit this page
```
kasugaijin, line 8: would be good to add an issue for the to dos

app/views/organizations/staff/fosterer_invitations/_form.html.erb
```diff
@@ -32,6 +32,7 @@
             %>
           </div>
 
+          <%# TODO Why are we adding a loaction to person here and nowhere else? %>
```
kasugaijin, line 35: I believe the thinking here is that for fosterers, who will physically possess a pet, staff need to ensure they have an address. Whereas, knowing the address of invited staff is less of a priority. Could be argued to justify always asking for these details.

kasugaijin (commented):
@jmilljr24 I like the direction this is heading. My main comments are around code organization. I think we can find a better place to handle roles, groups and activations. Might be a good thing to discuss on Tue.

app/models/person.rb
jmilljr24: 100% - I was definitionally questioning where to put methods after I was going through more and more flows.

app/views/organizations/staff/fosterer_invitations/_form.html.erb
jmilljr24: Hmm. That makes sense to have a way to add an address to a person for staff info purposes. I'm kind of leaning the other direction and give the option to add a location after the fact(not in the invitation flow) instead of requiring it. 

I'm sure there is a way around it but the invitation flow already has quite a few different scenarios and adding the location adds more complexity. Specifically handling user/person's already in the DB.

Also Location is pretty basic as the moment with just Country/State/Town and the only way to view a fosterers location is via edit so that could use some discussion.

kasugaijin: Yeah I’m fine with that approach also to keep invitations lean.

app/models/person.rb
kasugaijin: Cool. Happy to spitball ideas if you want to. But I think services make most sense.

app/models/person.rb
kasugaijin: I agree. I like the idea explicit validation here rather than relying on and gems. Devs could make mistakes and in this case it could be very costly eg making someone a staff in a different organization.

app/models/person.rb
jmilljr24: @kasugaijin continuing from our meeting on organization of these methods, I made some tweaks after seeing where all the methods would go and what was being used in multiple places. 

I kept `activation` it's own model.

 Because `staff` uses multiple method's that are applicable to all people/users, I decided to have a model that encapsulates all an people/user group and have `staff` inherit from that. 

We discussed called it `adopter_fosterer_staff` but I don't love it have `Staff < AdopterFostererStaff`. Open to discussion but I went with `GroupMember` (i.e. Member within this organization). This is namespaced to `GroupRoleManagement` so group may be redundant. We also discussed `Participant` (same idea, Participant within an org) but that didn't see to be a clear/descriptive term to you. I'm out of naming idea's haha.

I've added comments on the delegation to Person but I like how it turned out. Couple of examples:

```
@person.add_role_and_group(:adopter)
@person.staff?
```

This isn't delegated but I kind of prefer going through `.staff` for the specific `Staff` methods.
``` 
@person.staff.active?
@person.staff.change_role_and_group(:admin, :super_admin)
```

app/models/group_role_management/group_member.rb
```diff
@@ -0,0 +1,50 @@
...
+        .where(deactivated_at: nil, groups: {name: name})
+        .exists?
+    end
+
+    def deactivated_in_org?
+      !person.person_groups
+        .joins(:group)
+        .where(deactivated_at: nil)
+        .exists?
+    end
```
jmilljr24, line 35: These two methods could potentially go in `activation.rb` but I'm not sure what makes more sense. These are not manipulating an activation, more so to check on the person_group of a person. But that names lend themselves to the `Activation` model.

kasugaijin: Yeah I think it depends on for what purposes we might use them. If they are involved in activations i.e. making someone active, or not, then I'd move them, otherwise they could stay here just fine.

app/models/person.rb
kasugaijin: I think `GroupMember` works, even under `GrouRoleManagement` because the latter is encapsulating two distinct things, so the former is still adding value telling you which 'thing' is the focus i.e., Group. 

> This isn't delegated but I kind of prefer going through .staff for the specific Staff methods.

Hmmm I would personally prefer methods with staff within them, than having to go through an intermediary. e.g `person.active_staff?`

jmilljr24: > I think `GroupMember` works, even under `GrouRoleManagement` because the latter is encapsulating two distinct things, so the former is still adding value telling you which 'thing' is the focus i.e., Group.
> 
> > This isn't delegated but I kind of prefer going through .staff for the specific Staff methods.
> 
> Hmmm I would personally prefer methods with staff within them, than having to go through an intermediary. e.g `person.active_staff?`

Makes sense to me. A third option if you don't care exactly how the method is named, is to use a prefix on the delegation. So instead of having to add "staff" in every staff method name, the delegation would add the prefix `person.staff_active?` or `person.staff_change_role_and_group`

app/controllers/contacts_controller.rb
jmilljr24, line 1: moved under organizations

app/controllers/organizations/adopter_fosterer/likes_controller.rb
```diff
@@ -7,15 +7,15 @@ class Organizations::AdopterFosterer::LikesController < Organizations::BaseContr
...
-    likes = current_user.person.likes
+    likes = Current.person.likes
     @likes_by_id = likes.index_by(&:pet_id)
-    @pets = current_user.person.liked_pets.includes(:matches, :adopter_applications, images_attachments: :blob)
+    @pets = Current.person.liked_pets.includes(:matches, :adopter_applications, images_attachments: :blob)
   end
 
   def create
-    authorize! context: {pet: @pet}
+    authorize! @pet, with: Organizations::LikePolicy
```
jmilljr24, line 16: We are adding a like to the "pet"  so this is more direct by passing the pet as the record being acted on for the authorization check.

app/controllers/organizations/adopter_fosterer/likes_controller.rb
```diff
@@ -31,7 +31,7 @@ def create
 
   def destroy
     @pet = @like.pet
-    authorize! context: {pet: @pet}, with: Organizations::LikePolicy
+    authorize! @like
```
jmilljr24, line 34: We are destroying the `@like` so send the `@like` as the record. Then we don't need to worry about the pet.

app/policies/organizations/staff_policy.rb
jmilljr24, line 1: Renamed from `Organizations::UserPolicy` to Staff. This is a bit clearer as we are managing staff, not all users.

app/views/organizations/staff/people/_form.html.erb
jmilljr24, line 1: Moved out of staff namespace.

app/models/adopter_application.rb
```diff
@@ -24,9 +24,9 @@
 #  fk_rails_...  (pet_id => pets.id)
 #
 class AdopterApplication < ApplicationRecord
-  acts_as_tenant(:organization)
   belongs_to :pet, touch: true
   belongs_to :person
+  acts_as_tenant(:organization)
```
jmilljr24, line 29: Per Acts As Tenant docs, this should come after `belongs_to`.

kasugaijin:
> The invitation flow's are addressed but still have some minor tweaks required. When inviting a foster we have location on the form but not for inviting staff? 

Yeah, thinking there was that the physical address would be required for fosterers because the pet would be located at that address. But...this can be re-designed. It would just be nice for staff to be able to attach an address to whoever they want to e.g., staff, or fosterers, but not require it for an invite. A discussion for another day.

app/controllers/concerns/organization_scopable.rb
```diff
@@ -44,4 +51,12 @@ def after_sign_in_path_for(resource_or_scope)
   def after_sign_out_path_for(resource_or_scope)
     root_path
   end
+
+  def new_person_after_sign_up_path
+    if Current.organization.external_form_url
```
kasugaijin, line 56: let's add `.present` here just to be safe. I believe we have a validation to prevent users from saving an empty string, but it might still be possible. So, this will make sure there's actually a string/url

app/controllers/feedback_controller.rb
kasugaijin, line 1: I see what you mean where this gets funky when someone is in multiple groups. Would be good to just streamline this to a single layout, or use a modal, in the future.

app/controllers/organizations/activations_controller.rb
```diff
@@ -1,23 +1,37 @@
...
+        respond_to do |format|
+          format.html { redirect_back_or_to staff_staff_index_path, alert: @result[0] }
+          format.turbo_stream { flash.now[:alert] = @result[0] }
+        end
       end
     end
+
+    private
+
+    def set_pg
```
kasugaijin, line 29: At the risk of being verbose, can we use full names instead of abbrev? (pun intended). 
At first glance it's not obvious and I have to spend a few seconds figuring out what pg is.

app/controllers/organizations/activations_controller.rb
```diff
@@ -1,23 +1,37 @@
...
+      @pg = PersonGroup.find(params[:id])
+      @person = @pg.person
```
kasugaijin, line 31: I believe we can clean this a tad with 
`@person, @group = @pg.person, @pg.group`

app/controllers/organizations/activations_controller.rb
```diff
@@ -1,23 +1,37 @@
...
-    def update
-      @user = User.find(params[:user_id])
-
-      authorize! @user, with: ActivationsPolicy
+    before_action :set_pg
 
-      if @user.deactivated_at
-        @user.activate
+    def update
+      @result = if params[:person_group][:deactivated] == "true"
```
kasugaijin, line 6: I don't think `@result` needs to be an instance variable here? I don't see it used in the view.

app/models/group_role_management/activation.rb
kasugaijin, line 1: I think the description of how we use Groups, Person groups to manage user roles you provided in the PR description would be a useful bit of context to include in the app at a high level. The question is where to put it. I was thinking we could add a markdown file in the `GroupRoleManagement` scope directory?
 Unless you think it best lives on one of the models (Group, PersonGroup)?

app/controllers/organizations/staff/matches_controller.rb
```diff
@@ -19,6 +19,7 @@ def create
     end
   end
 
+  # TODO: Do we use this anywhere?
```
kasugaijin, line 22: No I don't think it is feasible via the UI. It's likely an edge where someone might return a pet after adopting...

app/controllers/organizations/staff/invitations_controller.rb
```diff
@@ -1,111 +1,126 @@
...
-      :index?, with: Organizations::AdopterFosterDashboardPolicy,
-      context: {organization: Current.organization}
-    )
-      adopter_fosterer_dashboard_index_path
-    else
-      root_path
-    end
-  end
+      def create_person(user)
+        # TODO: do we need this condition?
```
kasugaijin, line 115: we could probably get rid of this if we used `find_or_create_by` in the `set_person` method in place of this method call `create_person`?

app/controllers/registrations_controller.rb
```diff
@@ -11,13 +11,15 @@ def new
     respond_with resource
   end
 
-  # MARK: only adopters are created through this route. Adopters need both the adoper role and a form submission to attach their applications to
+  # MARK: only adopters are created through this route.
   def create
     super do |resource|
       if resource.persisted?
-        resource.add_role(:adopter, Current.organization)
+        # TODO: Currently a person shouldn't exist without a user with the same email. If the person exists (but no user),
```
kasugaijin, line 18: This is referring to a case where in the future staff might create a person and then that real person comes along and signs up? TBD. We might silently create a user all the time? Not sure.

app/models/person.rb
```diff
@@ -40,20 +43,41 @@ class Person < ApplicationRecord
...
 
+  delegate :activate!, :deactivate!,
+    to: :activation
+  delegate :add_group, :active_in_group?, :deactivated_in_org?,
+    to: :group_member
+  delegate :staff?,
+    to: :staff
+
+  delegate :active?, :current_group, :change_group,
+    to: :staff, prefix: :staff
```
kasugaijin, line 63: These could be moved into the `Authorizable` concern in an `included do` block

app/models/person_group.rb
```diff
@@ -0,0 +1,45 @@
...
+#
+#  fk_rails_...  (group_id => groups.id)
+#  fk_rails_...  (person_id => people.id) ON DELETE => cascade
+#
+class PersonGroup < ApplicationRecord
+  belongs_to :person
+  belongs_to :group
+
+  validates :person_id, uniqueness: {scope: :group_id}
+  validate :organization_ids
```
kasugaijin, line 28: I guess we could maje this singular, as they should be the same `id` ?

app/services/organizations/create_service.rb
```diff
@@ -75,14 +75,22 @@ def create_user(email, first_name, last_name)
...
+      @person = Person.create!(
+        user: @user,
+        first_name: first_name,
+        last_name: last_name,
+        email: email
+      )
     end
   end
 
   def add_super_admin_role_to_user
```
kasugaijin, line 88: Shall we update this method name to use `person`?

jmilljr24:
> > The invitation flow's are addressed but still have some minor tweaks required. When inviting a foster we have location on the form but not for inviting staff?
> 
> Yeah, thinking there was that the physical address would be required for fosterers because the pet would be located at that address. But...this can be re-designed. It would just be nice for staff to be able to attach an address to whoever they want to e.g., staff, or fosterers, but not require it for an invite. A discussion for another day.

I forgot to update that sentence after our meeting where we discussed it briefly . I did remove location from the invite form. You can still add a location via the UI under `edit` for fosterer's. That could use some changes but at least it is there for now.

app/controllers/organizations/activations_controller.rb
jmilljr24: I went back and forth on the name as well. Whatever you think is best for clarity as you are reading it from an outside perspective. 

Realisticly is it actually `set_person_and_group_and_persongroup` but that is kinda of ridiculous haha .

 (`@pg` is used in the view)

app/controllers/organizations/staff/adopters_controller.rb
```diff
@@ -3,12 +3,13 @@ class Organizations::Staff::AdoptersController < Organizations::BaseController
...
-    authorize! Person, context: {organization: Current.organization}
+    authorize! Person
 
-    @q = authorized_scope(Person.adopters).ransack(params[:q])
+    @q = authorized_scope(Person.adopters.includes(person_groups: :group)).ransack(params[:q])
     @pagy, @adopter_accounts = pagy(
       @q.result,
       limit: 10
     )
+    @group_id = Group.find_by(name: "adopter")&.id
```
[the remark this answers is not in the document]
jmilljr24: Good catch. I think it should just be a symbol as we are using an enum in Group.

app/controllers/organizations/staff/matches_controller.rb
jmilljr24: Should we remove it until it is needed?

app/models/person_group.rb
jmilljr24: Hmm, they should be the same value but my if I was to say "validate the organization **id** on PersonGroup, I initially would think is a validation for PersonGroup.organization_id (which we don't have).

Plural tells me it is looking at two different things. I could be convinced otherwise.

app/services/organizations/create_service.rb
jmilljr24: I would like to discuss naming conventions at the next meeting. (but yes in this specific case!)

app/models/group_role_management/activation.rb
```diff
@@ -0,0 +1,40 @@
+# This is a PORO Model that encapsulates logic for modifying associations on Person/User models
+
+# Methods in this class are delegated to the Person model via the `activation` association.
+# This allows calls like `@person.activate!(:adopter)` to invoke logic defined here.
+#
+module GroupRoleManagement
```
kasugaijin, line 6: I was also thinking about how to define these methods, right now it takes a single argument that accounts for the group, or the name, how about splitting these into two separate possible arguments, group, name? I think that's more conventional. Then check at least one is provided.

app/controllers/organizations/staff/matches_controller.rb
kasugaijin: Yeah may as well.

app/controllers/registrations_controller.rb
jmilljr24: Yes, that is the scenario I was thinking of.  I wanted to make note because there are multiple ways we could handle it with varying consequences. It doesn't need to be figured out right now but I also want to make sure what we implement now doesn't cause future issues or get forgotten. 

For example, we could simply link the existing person to a new user. This doesn't guarantee the "Person" is the same as the "User"  that is signing up other than the emails match. (maybe the existing persons email was never verified and is wrong). Highly unlikely but possible.

app/controllers/organizations/activations_controller.rb
jmilljr24: I'd prefer to keep the assignments on separate lines for readability. I don't believe we've generally been using parallel assignments for defining single variables in this code base. 

Is that something you were thinking of moving towards? Not a big deal but it does go against the [Ruby Style Guide](https://rubystyle.guide/#assignment-comparison).

test/controllers/country_states_controller_test.rb
```diff
@@ -2,7 +2,7 @@
 
 class CountriesStatesControllerTest < ActionDispatch::IntegrationTest
   test "should return turbo stream with the states in it" do
-    adopter = create(:adopter)
+    adopter = create(:person, :adopter).user
```
kasugaijin, line 5: Another one for the naming - variable is `adopter` but object is `user`

kasugaijin (approved):
Phew...what a lot of work. Great job. I think this looks really good. Given the size I wasn't going through with a fine toothcomb on the views, but this is stuff we should see, and find easily if anything is awry.

app/controllers/organizations/activations_controller.rb
kasugaijin: Yeah let's leave as is. I think this is a good exception to the rule.

app/controllers/organizations/activations_controller.rb
kasugaijin: Yeah that's fine. I am not married to either.

app/models/person.rb
kasugaijin: We could make a separate concern just for this. They feel to me like they are more to do with groups and person groups than the actual Person model. It's quite normal for this kind of stuff to be located in an included concern.

app/models/person_group.rb
kasugaijin: Yeah you are right, let's leave as is.

app/models/group_role_management/activation.rb
kasugaijin: Nice.

app/models/person.rb
jmilljr24: Good article. Thank you. I'm going to leave this as I have it for now.

If you would like to put up a PR after this one is finished with the structure you are think of, I'd be happy to look it over and chat more about it. My brain is just not connecting the dot's on how this would fit in the "good concern" pattern.

kasugaijin: Sounds good! 

It’s ok where it is for now. But if the person model starts to grow, I think it would make sense to move it. 

The way I see it is that this logic is peripheral to the person itself, and isn’t core person logic. After all, it is getting and setting other models for the most part, through other classes no less. It’s like an add on to the person model.

kasugaijin:
> @kasugaijin Thanks for the review! I've pushed up the changes. If you are happy with things, I will merge the latest changes from main and handle any conflicts. Then we should be set for the next phase.
> 
> I held off on the markdown file. I'd like to do that in conjunction with updating any other doc's regarding roles that we already have in place but will need to look through things.

Sounds good. Merge away!

Issue #1421, opened 2025-05-25
Verify Omniauth Sign-Up flow


jmilljr24:
A User can sign up to an Organization via Google OAuth. The first time joining any Homeward Tails Organization an unscoped `user` is created for the associated email address and a `person` scoped to the specific organization they are joining is created.

Verify the person is created with the group `:adopter` for the initial organization. Sign out. Attempt to sign in/join while scoped to a different organization and verify a new person (with adopter group) is created for this separate organization.

jmilljr24:
@sarvaiyanidhi Not sure if you are interested or available for this, but I know you did the initial work with adding the google sign-up. 

We made some larger changes to Users/Persons/Roles/Groups so this may need to some tweaking.

jmilljr24:
@sarvaiyanidhi Yes! I've assigned you. If there is anything that comes up that is outside of the scope of what you would like to do, just comment and I'll open it back up to other contributors. I just recall you had to set some extra things up specifically to work on this in development so I want to give you the first option to take a look.

PR #1436, opened 2025-05-29
1421 - Verify and fix Google Omniauth flow

sarvaiyanidhi:
# 🔗 Issue
#1421 

# ✍️ Description
- Current implementation was giving issue when we try to sign in with same google user for different organization as user_id was not getting mapped in database table for person table so flow was getting stuck in new person page. It was not allowing user to even logout due to error.
- Have fixed issue by moving people creation flow outside user loop as issue with previous implementation was that person and its group was getting created in database before user was created when user first time sign up using google login.

# 📷 Screenshots/Demos

[Alta Pet Rescue _ google_onmiauth.webm](https://github.com/user-attachments/assets/0ae2ba42-7b65-4977-9724-1a7e2301ced3)

### DIFF
--- a/app/models/concerns/omniauthable.rb
+++ b/app/models/concerns/omniauthable.rb
@@ -7,10 +7,12 @@ module Omniauthable
 
   class_methods do
     def from_omniauth(auth)
-      where(provider: auth.provider, uid: auth.uid).first_or_create do |user|
-        user.assign_attributes_from_auth(auth)
-        user.set_adopter_role
+      user = where(provider: auth.provider, uid: auth.uid).first_or_create do |u|
+        u.assign_attributes_from_auth(auth)
       end
+
+      user.set_adopter_role if user.persisted?
+      user
     end
   end
 


sarvaiyanidhi:
@jmilljr24 I verified Google Omniauth flow and fixed issue when same google user tries to sign in with different organization but there is one more gap in this flow when we compare it with usual sign up flow.

- In normal sign-up flow when same user tries to login with different organization, it is redirected to people/new page to join the organization
- In Google login flow, user is not getting directed to people/new page though person and its group is created in database and he is logged into the system

Only issue I see is that below page is not shown in google flow so user cannot update those details.. Even after login I could not find option to update phone number for person in any screen

image_2025-05-29_142127700

jmilljr24:
@sarvaiyanidhi Thanks for looking this over!

Good catch on the phone number. I see we don't actually collect the phone number on the initial sign up with manual entry (not google), only when a user joins a second org on that new person sign up form. I think that was added by mistakes but I will confirm and create a separate issue. I believe we are only using phone numbers, for example, when staff want to add a phone number to a fosterer via that staff dashboard.

> In Google login flow, user is not getting directed to people/new page though person and its group is created in database and he is logged into the system

-  If a user has joined Org A, then signs out, visits Org B and signs in (there by joining Org b) without hitting the new person form is totally acceptable. 
- The problem would be where they have joined Org A and then visit Org B while still authenticated. If I understand correctly, you are saying their new "person" for Org B would be created with the adopter group without hitting the new person form. One of main reasons we want the new person form for Org B is so the person is actually choosing to join the new org. If they are authenticated and visit Org B by accident, it wouldn't be their choice on joining if the person is created automatically.


So to reiterate the ideal flows:

Scenario 1:
-Unauthenticated: User joins Org  A with google, person is created for Org A with details from google
-Unauthenticated: User then joins Org B with google, person is created for Org B with details from google

Scenario 2:
-Unauthenticated: User joins Org A with google as above.
-Still authenticated: User visits Org B, prompt to join Org B (could be the same new person from, or a simple confirmation)

Let me know your thoughts. Tagging @kasugaijin for visibility/input.

kasugaijin:
@jmilljr24 yes this is consistent with how I understand it. 

I agree that asking for a phone number is not necessary on the person form.

jmilljr24 (approved):
@sarvaiyanidhi Thank you so much for taking care of this!!

PR #1448, opened 2025-06-11
Production db backup to local

jmilljr24:
# 🔗 Issue

# ✍️ Description

This allows backing up the production database to your local machine and restoring it to the dev db. I have not tested this using docker.

 This could be altered easily in the future to create the backup and store it in an azure blob. [Reference human-essentials bakup](https://github.com/rubyforgood/human-essentials/blob/main/lib/tasks/backup_db.rake) and [restore](https://github.com/rubyforgood/human-essentials/blob/main/lib/tasks/fetch_latest_db.rake). We currently rely on Azures built in db backup but I do not believe that we can access the actual dump file. Creating a backup blob would be an alternate/additional method. Something to discuss in the future.

# 📷 Screenshots/Demos

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
-ruby 3.3.0
-postgres 12.12
+ruby     3.3.0
+postgres 16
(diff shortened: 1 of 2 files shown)


kasugaijin (approved):
great idea!

PR #1469, opened 2025-06-28
#1465 add powered by homeward tails links

### LINKED ISSUE #1465: As a user I can see a 'powered by Homeward Tails' link on the log in/sign in pages
We display the org name (and logo if present) on the log in and sign up pages. We should add a small link below this `Powered by Homeward Tails` that will open a new tab to the app root page. 

Note, because this app is multi-tenanted and we have middleware that sets the URL base on the slug, directing to `root_path` will take the user to the org home page, not the application route. So, you will need to figure out how we can cleanly direct the user to the application root page (dev = localhost:3000, production = www.homewardtails.org).

jmilljr24:
# 🔗 Issue

Resolves #1465 

# ✍️ Description

Adds powered by Homeward Tails links on sign_in/sign_up pages. Link opens global root page in a new tab.

I think having the link at the bottom of the card is unobtrusive and keeps the top of the form clean. Open to suggestions on styling/placement.

# 📷 Screenshots/Demos

[Screencast from 2025-06-27 20-25-04.webm](https://github.com/user-attachments/assets/8f117a68-dc7e-47ed-bcf0-5f3113e0f5b7)

### DIFF
--- a/app/views/devise/registrations/new.html.erb
+++ b/app/views/devise/registrations/new.html.erb
@@ -9,7 +9,10 @@
             <div class='d-flex justify-content-center'>
               <div class='d-flex flex-column align-items-center mb-3'>
                 <% if Current.organization.avatar.attached? %>
-                  <%= image_tag Current.organization.avatar, alt: current_organization_name, title: current_organization_name, height: 75 %>
+                  <%= image_tag Current.organization.avatar,
+                  alt: current_organization_name,
+                  title: current_organization_name,
+                  height: 75 %>
                 <% end %>
                 <h1><%= Current.tenant.name %></h1>
               </div>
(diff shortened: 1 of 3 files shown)


kasugaijin (approved):
Looks good! I agree on the placement. I also think for consistency we should update the org footer where we have the ‘if you want a website like this one…’ bit and have this there as well instead.

Also, I think we can move this to a shared partial that lives in app/views/orgs/shared or wherever makes sense. It could be global scoped as well instead app/views/shared

jmilljr24:
> Looks good! I agree on the placement. I also think for consistency we should update the org footer where we have the ‘if you want a website like this one…’ bit and have this there as well instead.
> 
> Also, I think we can move this to a shared partial that lives in app/views/orgs/shared or wherever makes sense. It could be global scoped as well instead app/views/shared

Good idea! I will create an issue for that. I'm working on some other things right now. If no one picks it up, I'll knock it out later.

Issue #1456, opened 2025-06-18
Improve UI/UX for add/updating fosterer location information


jmilljr24:
Only staff can add location information to a fosterer but it is not obvious. We should improve this. 

In Staff Dashboard > Fosterers `/alta/staff/fosterers`  change the column header "Action" to "Contact Info". Change the "edit" button to "view/edit". If there is too much text for the table to display properly we can consider removing phone as that information is present in the contact info page. Make appropriate changes to the mobile view as well.

Add `street` to the location model.

Improve the look of the contact information form (spacing, alignment, etc.). Add the optional street attribute. The staff member should be able to submit the form without location data as not all fosterers may provide it to the staff member. Consider adding a note near the name attributes letting staff know they should only change the name if there is a mistake or request by the fosterers. Generally the fosterer should change there own name but it is possible that the fosterer does not interact with the website so staff should be able to handle it.

Change the title "Edit #{Fosterer Name} details" to "Contact Information - #{Fosterer Name}"

PR #1473, opened 2025-06-29
Add person profile page for Staff to have a single page with person information

jmilljr24:
# 🔗 Issue

# ✍️ Description

This builds out an initial profile view of a person available on the staff dashboard. The page currently implements a details (edit first/last name) tab and form submission tab. The tabs are built using bootstrap.js which we already use. 

I've creating a basic layout that is responsive. The initial page will load the basic details of the person. Each tab is lazy loading via a turbo frame once selected.

I've built upon the `PeopleController`. I've left it out of the staff namespace as we will still use `new` and `create` when an person joins a second organizations. For routing, I've added `index` and `show` under staff. `show` being the new profile page for staff to view a person. The url will be `staff/people/:id`. 

Let me know if this is generally what you had in mind. I can keep building it out as a draft, or do some final touches for it to be merged and then add new tabs later.

# 📷 Screenshots/Demos

[Screencast from 2025-06-29 12-34-16.webm](https://github.com/user-attachments/assets/00305fb8-4906-4926-be78-f0dbfa006660)


jmilljr24:
@kasugaijin How's this looking for a start? There are definitely  some discussion points on what information we want where but this is a relatively simple new page to start with.

kasugaijin:
Nice - taking shape and just how I was envisioning it, too. I checked it out and played with it (finally got the app running locally 😺 ).

Some small things (you may or may not be aware so mentioning eveything I see)
- the index table is not visible for mobile (disppears)
- would be nice to update the breadcrumb to link to the index so that if I view a given show, I can click back to index easily
image
- the link to the CSV upload page is broken - probably needs a `target: :top` 
image

I think it's a good start and I like the architecture of loading async and getting resources from existing controllers where it makes sense to. I can envision this evolving over time a fair bit e.g., we might want to combine all of the people pages into a single on, and/or start to show more info under a different layout for a person...but let's get feedback from a user before we dive into perfection.

I'll review the code now and make more comments if I need to.

app/controllers/organizations/people_controller.rb
```diff
@@ -1,8 +1,21 @@
...
+    skip_before_action :verify_person_in_org, only: %i[new create]
     skip_verify_authorized only: %i[new create]
-    before_action :validate_person_does_not_exist
+    before_action :validate_person_does_not_exist, only: %i[new create]
+
+    def index
+      authorize!
+
+      @q = authorized_scope(Person.all).ransack(params[:q])
+      @pagy, @people_accounts = pagy(
```
kasugaijin, line 14: nit: could just call this `@people` ? I think that's more idiomatic in Rails.

app/controllers/organizations/staff/user_roles_controller.rb
```diff
@@ -1,7 +1,12 @@
 module Organizations
   module Staff
     class UserRolesController < Organizations::BaseController
-      before_action :set_person
+      before_action :set_person, only: [:to_admin, :to_super_admin]
+
+      def index
```
kasugaijin, line 6: Assuming this is wired up for the Deactivate column on the index?
We can discuss what else to include on that table, too, on Tue.

app/views/organizations/people/_form.html.erb
kasugaijin, line 1: Something to discuss - should staff be able to edit the Person's direct attributes? Maybe some cases make sense? I just wouldn't want a staff to name someone `Cathy Walters (with annoying laugh)` because Cathy would see it and not be stoked (probably). I think we need to be careful letting staff edit personal data that the person can see.

app/views/organizations/people/show.html.erb
kasugaijin, line 1: We can discuss what else to show here - e.g. might be useful to see what groups the person belongs to etc. No need to perfect this...we can take a good guess, then let a real user tell us what is expected.

app/views/organizations/staff/form_submissions/index.html.erb
kasugaijin, line 1: Currently, staff can go to adopter application reviews and click on an adopter to see their form submissions. This path is now broken (see [this partial](https://github.com/rubyforgood/homeward-tails/blob/b7bd8a0dbc568a6ada03dcd0486176e18cc02566/app/views/organizations/staff/pets/tabs/partials/_application_info_card.html.erb)) 

I think we can update that partial to instead point to `/staff/people/:id` that you have built. I am not sure if we can pass a param to open up the form submissions tab specifically. I don't think that's a deal breaker if not simply do-able.

app/controllers/organizations/staff/user_roles_controller.rb
jmilljr24: Ah yeah, I started to look at adding a group page/index columns but there were too many questions on what we wanted where.

app/views/organizations/staff/form_submissions/index.html.erb
jmilljr24: I believe we would need some js. A simple stimulus controller would do it. The bootsrap js tabs are just adding and removing classes.

kasugaijin: Yeah let's not bother with that for now. It's not a big issue and probably not worth the extra code until we can test it out and get feedback.

app/views/organizations/people/show.html.erb
jmilljr24: Location would be good to have as well as an optional form. We could close #1456 as I think this profile is a more intuitive place and can be used on any person, not just a fosterer. Maybe list of all applications from the person?

app/views/organizations/people/_form.html.erb
kasugaijin: Yeah I see your point. I think we should keep it limited on fields for now, and explore further with a test user.

app/views/organizations/people/show.html.erb
kasugaijin: Agreed with location. I also think it would make sense to show adoption applications, fosters, adoptions

jmilljr24: Ok. I might add location in this pr since it is somewhat tied to person (nested attributes). The others could be built out in future issues. I'm guessing some things wont be as simple as slapping a turbo frame in.

jmilljr24:
I'm pretty happy with this for the initial layout. I feel good about how I have the `edit name` implemented. I've added a very explicit warning and only allow super_admin to edit. I created the following controller/policy conditions:

- Adopter/Fosterer can edit/update their own person attributes
- Admin can edit/update person attributes other than name ( phone_number, location)
- Super Admin can edit/update all person attributes

An admin cannot edit their own name in the profile page (or anyone elses). They should edit their name in their account settings. The account settings page should be updated to use this policy once we remove name for `User`.

If you strongly appose allowing super_admin to edit names, I'm happy to remove it.

[Screencast from 2025-07-04 07-52-11.webm](https://github.com/user-attachments/assets/f408a1df-2654-4244-8fbb-5e31f13ff4d8)

app/views/organizations/people/show.html.erb
```diff
@@ -0,0 +1,52 @@
+<%= render DashboardPageComponent.new(crumb: :person, crumb_options: @person) do |c| %>
+  <% c.with_header_title { t(:people) } %>
+  <% c.with_body do %>
+    <nav>
+      <div
+        class="nav nav-tabs d-flex-inline flex-column flex-sm-row"
+        id="nav-tab"
+        role="tablist"
+      >
+        <button
```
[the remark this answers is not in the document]
jmilljr24: Great point! I haven't done much with view components yet. I'll have to play around with it.

PR #1491, opened 2025-07-07
Show completed tasks on load and set toggle default to checked

caitlynl22:
# 🔗 Issue

https://github.com/rubyforgood/homeward-tails/issues/1481

# ✍️ Description

Removed code to hide completed tasks on load and set initial checkbox state to checked.

# 📷 Screenshots/Demos

https://github.com/user-attachments/assets/ead1afa0-a282-4b96-ab83-9872f4b081d8

### DIFF
--- a/test/system/tasks_test.rb
+++ b/test/system/tasks_test.rb
@@ -48,8 +48,6 @@ class TasksTest < ApplicationSystemTestCase
       check "task_completed"
     end
 
-    find("input#flexSwitchCheckChecked").click # Toggle to show completed tasks.
-
     assert_text("recurring task", count: 2)
     assert has_current_path?(staff_pet_path(@pet, active_tab: "tasks"))
   end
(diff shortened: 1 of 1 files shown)


jmilljr24:
@caitlynl22 The changes to stimulus look good, however when I run this locally I'm noticing two issues.

1. When initially loading the page with a completed task, the task does not show even though the toggle is correct for "show completed"
2. When I check a task as completed, the completed task is removed from the list instead of changing to green and showing completed.

In both cases, toggling the "show completed" of and back on resolves the issues. Maybe something to do with turbo? I can dig deeper if you need help.

[Screencast from 2025-07-08 06-47-18.webm](https://github.com/user-attachments/assets/f3e57f57-a329-4306-9868-993161a70ea5)

caitlynl22:
> @caitlynl22 The changes to stimulus look good, however when I run this locally I'm noticing two issues.
> 
> 1. When initially loading the page with a completed task, the task does not show even though the toggle is correct for "show completed"
> 2. When I check a task as completed, the completed task is removed from the list instead of changing to green and showing completed.
> 
> In both cases, toggling the "show completed" of and back on resolves the issues. Maybe something to do with turbo? I can dig deeper if you need help.
> 
>  Screencast.from.2025-07-08.06-47-18.webm

Hmmm I'm not seeing either of those issues. I was before I removed the line hiding the completed tasks on load. Maybe rebuilding assets or restarting the server?

jmilljr24 (approved):
@caitlynl22 you were correct! LGTM!!

Issue #1485, opened 2025-07-02
Pet tabs - add empty states for Applications and Tasks


kasugaijin:
Log in as staff and go to admin dashboard > pets > pet > Applications tab  & Tasks tab
There are no empty states - create empty states for these that say 'There are no <adoption applications/tasks> for this pet.'

crespire:
I can take this one!

PR #1493, opened 2025-07-09
Empty state for tasks and applications

crespire:
# 🔗 Issue
Issue #1485 

# ✍️ Description
We add an empty state for tasks and applications that primarily uses CSS to leverage the `:only-child` psuedo selector.

Additional changes were required for the toggling of empty state when toggling tasks, so changes were made to the `toggle-completed` Stimulus controller to add a hide class you can input from HTML, as well as checking whether you were toggling all tasks or not to decide if we needed the empty state to show up.

# 📷 Screenshots/Demos
Tasks New - Empty
image

Applications New - Empty
image

All Completed Tasks - Before Toggle
image

All Completed Tasks - After Toggle
image

### DIFF
--- a/app/assets/stylesheets/custom.scss
+++ b/app/assets/stylesheets/custom.scss
@@ -47,10 +47,10 @@ label.required:after {
 
 
 // Custom styling for only-child psuedo class
-.list-group .only-child-block:not(:only-child) {
+.only-child-block:not(:only-child) {
   display: none;
 }
 
-.list-group .only-child-block:only-child {
+.only-child-block:only-child {
   display: block;
 }
--- a/app/assets/stylesheets/custom.scss
+++ b/app/assets/stylesheets/custom.scss
@@ -44,3 +44,13 @@ label.required:after {
     }
   }
 }
+
+
+// Custom styling for only-child psuedo class
+.list-group .only-child-block:not(:only-child) {
+  display: none;
+}
+
+.list-group .only-child-block:only-child {
+  display: block;
+}
--- a/app/javascript/controllers/toggle_completed_controller.js
+++ b/app/javascript/controllers/toggle_completed_controller.js
@@ -1,17 +1,26 @@
 import { Controller } from "@hotwired/stimulus";
 
 export default class ToggleCompletedController extends Controller {
+  static classes = ["show"];
+
   connect() {
-    this.completedTasks = document.querySelectorAll('.bg-success-subtle');
+    this.completedTasks = document.querySelectorAll('.task-list .bg-success-subtle');
+    this.taskFrames = document.querySelectorAll('.task-list turbo-frame');
+    this.emptyState = document.querySelector('.task-list li.only-child-block')
   }
 
   toggleCompletedTasks() {
     const isChecked = this.element.checked;
     if (isChecked) {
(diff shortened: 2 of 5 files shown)


jmilljr24 (commented):
@crespire Thanks for working on this!

We've generally avoided adding custom css in this project unless absolutely necessary. What are you thoughts on handling this server side?

```
if applications.any? 
  applications.each ....
else
  <td> There are no applications....
end
```

Same for the tasks. We are also using [bootstrap list group](https://getbootstrap.com/docs/5.3/components/list-group/#basic-example) if you want to keep consistent styling for the `li`

crespire:
> @crespire Thanks for working on this!
> 
> We've generally avoided adding custom css in this project unless absolutely necessary. What are you thoughts on handling this server side?
> 
> ```
> if applications.any? 
>   applications.each ....
> else
>   <td> There are no applications....
> end
> ```
> 
> Same for the tasks. We are also using [bootstrap list group](https://getbootstrap.com/docs/5.3/components/list-group/#basic-example) if you want to keep consistent styling `li class="list-group-item"`

Thanks, will change the styling on the list-group-item!

In terms of doing branching in the template, if that's the way we'd like to go, I'm happy to do that, but I will have to re-think how we do the "All Completed Tasks" case, and so there will likely need to be a pre-rendered, hidden element that we then show via Javascript anyway, so we sort of end up back at what I've already implemented but with an extra branch in the view code.

Let me know if that really is the preferred method, and I can figure it out. At least for applications to adopt, we don't have that issue because either there are some or there aren't any, and there's no frontend toggle to modify the display.

jmilljr24:
> In terms of doing branching in the template, if that's the way we'd like to go, I'm happy to do that, but I will have to re-think how we do the "All Completed Tasks" case, and so there will likely need to be a pre-rendered, hidden element that we then show via Javascript anyway, so we sort of end up back at what I've already implemented but with an extra branch in the view code.
> 

Correct me if I am wrong, I believe you are describing the case where there is one or more completed tasks but "Show Completed" is unchecked (so completed tasks are hidden). If we stick the original text "'There are no tasks for this pet." having no empty state text is that case makes sense. There are tasks but they are hidden. We default to having "Show Completed" checked, so the user would have to see they have completed tasks and then click the "Show Completed"  toggle (to off) to get to the state were there is no empty state text. I'm not sure how much value adding text in that case gives us at the expense of complexity. Maybe @kasugaijin has some thoughts.

I'm not apposed to tweaking the displayed text but I'm not sure "active" is the best term. It's a bit ambiguous to me and seems to be driving this edge case.

> In terms of avoiding custom CSS, I don't think there's a way around it for psuedo-classes as I can't use them in line. I did see places in the app where we use inline `<element style="">` so I had considered that usage. I feel that this is a compelling case of custom CSS, as I'm not overriding any Bootstrap styling, the utility class is simply not available. That said, I am happy to defer to you and the other project maintainers.

If we want to hide an element via stimulus, I believe we can just add the class `d-none` (bootstrap).

jmilljr24:
@crespire just checking in on this.

crespire:
> @crespire just checking in on this.

Hey, sorry, just haven't had time recently, but it's on my radar!

crespire:
@jmilljr24 This should be ready for review!

PR #1497, opened 2025-07-12
tweak org footer

kasugaijin:
# 🔗 Issue

# ✍️ Description
Just proposing re-designing the org footer a bit, and the powered by logo to just smush it a bit closer together. Style could be finessed at smallest width, but this is a movement to what I think it a slightly better layout in the footer. Thoughts?
Lg
image

Md
image

Sm
image

# 📷 Screenshots/Demos

### DIFF
--- a/app/views/layouts/shared/_footer.html.erb
+++ b/app/views/layouts/shared/_footer.html.erb
@@ -8,9 +8,8 @@
           <h3>
             <%= t(".organization_name", name: Current.tenant.name) %>
           </h3>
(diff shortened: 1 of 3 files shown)


kasugaijin:
@jmilljr24 good feedback...I battled with bootstrap and think I got it to a good place :) I decided to just switch the order of the links and Copyright and think it looks good (saves switching order based on screen width).

Lg
image

Md
image

Sm
image

PR #1499, opened 2025-07-12
Add active to Org and small fixes

kasugaijin:
# 🔗 Issue
Adds active column to Org that is default to true. 
Also cleans up the static controller a bit. 
I also made the org top nav white because A) it's cleaner and B) logos are much easier to make look good on a white background.

image

@jmilljr24 if you have time could you set Baja and Petey to `active: false` in production please? Feel free to merge this if you are happy with it.

# ✍️ Description

# 📷 Screenshots/Demos

### DIFF
--- a/db/migrate/20250712211752_add_active_column_to_organizations.rb
+++ b/db/migrate/20250712211752_add_active_column_to_organizations.rb
@@ -1,0 +1,8 @@
+class AddActiveColumnToOrganizations < ActiveRecord::Migration[8.0]
+  def change
+    safety_assured do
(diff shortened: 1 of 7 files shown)


db/migrate/20250712211752_add_active_column_to_organizations.rb
kasugaijin, line 1: We shouldnt need to backfill any prod data given the default value.

jmilljr24 (approved):
lgtm! I'll deploy the changes and set the values soon.

PR #1512, opened 2025-07-19
update SMTP to use GMAIL and add credentials to production creds

kasugaijin:
# 🔗 Issue

# ✍️ Description
Sendgrid have borked their free tier. So, I want to test using Gmail SMTP in production. It's probably the simples approach for now, until more is needed.

# 📷 Screenshots/Demos


jmilljr24:
@kasugaijin lint failed so the deployment action probably won't work. If you have Kamal working locally you can always just deploy your working branch for now since we don't have users. Might be less painful if you need to do minor changes to make this work vs waiting on a PR review.

kasugaijin:
> @kasugaijin lint failed so the deployment action probably won't work. If you have Kamal working locally you can always just deploy your working branch for now since we don't have users. Might be less painful if you need to do minor changes to make this work vs waiting on a PR review.

That's something I still need to set up on this machine :)
This time was an easy lint fix.

Issue #1452, opened 2025-06-18
Fix policy for Fostered Pets page


jmilljr24:
If I am active in the group `:adopter` but not `:fosterer` and visit the adopter dashboard I do not have a link to `Fostered Pets` which is correct. However, I am able to view the fostered pets page if I type the url directly `alta/adopter_fosterer/fostered_pets`

Update the policy so only people with the active group `fosterer` can visit that page. Update the policy tests.

Aaryanpal:
@jmilljr24 Please Assign this!
Will get this done in Weekend

jmilljr24:
> @jmilljr24, I noticed that two controllers are using the same `app/policies/organizations/adopter_fosterer/match_policy.rb` file.
> There are a couple of ways we can handle this:
> 
> 1. Pass specific context from each controller to the `MatchPolicy`.
> 2. Alternatively, create a separate policy specifically for the foster controllers.
> 
> Let me know what you think.
> 

Let's go with #1

PR #1521, opened 2025-07-25
1452 - Fix policy for Fostered Pets page

Aaryanpal:
# 🔗 Issue
#1452

# ✍️ Description
- Updated `adopted_pets_controller.rb` and `fostered_pets_controller.rb` to include context in the authorization scope for better clarity.
- Modified `MatchPolicy` to handle page-specific permissions for adopted and fostered pets.
- Adjusted tests to reflect changes in authorisation context for both controllers.

# 📷 Screenshots/Demos

[Screencast from 26-07-25 01:19:40 AM IST.webm](https://github.com/user-attachments/assets/e90d0ca3-5426-47e4-b12e-761b71fd62f3)

### DIFF
--- a/test/controllers/organizations/adopter_fosterer/adopted_pets_controller_test.rb
+++ b/test/controllers/organizations/adopter_fosterer/adopted_pets_controller_test.rb
@@ -21,7 +21,7 @@ class Organizations::AdopterFosterer::AdoptedPetsControllerTest < ActionDispatch
         ActsAsTenant.with_tenant(@organization) do
           assert_have_authorized_scope(
             type: :active_record_relation,
-            with: Organizations::AdopterFosterer::MatchPolicy, context: {page: "adopted_pets"}
+            with: Organizations::AdopterFosterer::MatchPolicy, context: {type: "adoption"}
           ) do
             get adopter_fosterer_adopted_pets_url
           end


app/models/concerns/authorizable.rb
```diff
@@ -22,7 +22,7 @@ def permission?(name)
 
   FOSTERER_PERMISSIONS = %i[
     view_adopter_foster_dashboard
-    view_adopter_foster_matches
+    view_foster_adopter_matches
```
jmilljr24, line 25: We can probably just use `view_foster_matches` as that is more specific. It also make me think would should probably change the adopter permission `view_adopter_foster_matches` to just `view_adopter_matches`. I'm not sure if that is outside the scope of this PR.

app/policies/organizations/adopter_fosterer/match_policy.rb
```diff
@@ -1,12 +1,17 @@
 module Organizations
   module AdopterFosterer
     class MatchPolicy < ApplicationPolicy
+      authorize :page
```
jmilljr24, line 4: I'm wondering if we can make this a bit more reusable and change it to something like `:type`. This would align more with the match model of match_type. I could be convince of a better name but I don't think we should tie it directly to the page.

It would end of being `MatchPolicy, context: {type: :foster}` or `type: :adoption` which I think reads well and makes sense.

app/policies/organizations/adopter_fosterer/match_policy.rb
```diff
@@ -1,12 +1,17 @@
...
       relation_scope do |relation|
         relation.where(person_id: person.id)
       end
 
       def index?
-        permission?(:view_adopter_foster_matches)
+        if page == "fostered_pets"
+          person.active_in_group?(:fosterer) && permission?(:view_foster_adopter_matches)
```
jmilljr24, line 11: `active_in_group` is already checked in `authorizable.rb`

```suggestion
          permission?(:view_foster_adopter_matches)
```

app/controllers/organizations/adopter_fosterer/fostered_pets_controller.rb
```diff
@@ -6,13 +6,13 @@ class FosteredPetsController < Organizations::BaseController
 
       def index
         # TODO: adopter can visit this page
```
jmilljr24, line 8: We can remove this TODO.

jmilljr24 (changes requested):
@Aaryanpal Nice work! Just some comments on naming and a little clean up.

app/controllers/organizations/adopter_fosterer/fostered_pets_controller.rb
Aaryanpal: Done

app/models/concerns/authorizable.rb
Aaryanpal: Hey @jmilljr24,i have made the changes. In the Same PR.
Now it makes more sense with naming.

app/policies/organizations/adopter_fosterer/match_policy.rb
Aaryanpal: I have made the changes suggested.
Totally Agree what you suggested.

jmilljr24 (approved):
@Aaryanpal Awesome!! I appreciate it.

PR #1540, opened 2025-08-07
Admin Dashboard Person Index - display groups a person belongs to

jmilljr24:
# 🔗 Issue

# ✍️ Description

This adds to the index table the groups a person belongs to and displays if they are deactivated. If they do not belong to a specific group, a button to add the group is displayed. The button to add a staff group( currently defaults to `:admin`) is disabled if the current user is not a `:super_admin`. If the current user is an `:admin` they can only add the groups `:adopter` or `:fosterer`. 

This removes the friction of having to go through the invite flow of typing the persons email for an existing person in the org . This also resolves the issue we currently have of no UI option to add the `:adopter` group to an existing person.

Future issues:
- Update the invite flows to reject email for people within the org. Add notice stating to use the dashboard buttons
- The button could be updated to the check mark symbol when a group is added via turbo_frame vs the page reloading
-We could add the email notification here a lot easier than the current invite flow or just only do in app notification as previously discussed.
# 📷 Screenshots/Demos

[Screencast from 2025-08-07 12-22-02.webm](https://github.com/user-attachments/assets/c8a220c6-117f-4e93-a3d0-e41ebfe83dc9)

Screenshot from 2025-08-07 13-24-56


app/controllers/organizations/people_controller.rb
```diff
@@ -62,6 +67,15 @@ def update
       end
     end
 
+    def add_group
```
kasugaijin, line 70: The rails purist in me says we should use RESTful actions in controllers, so we'd probably want to scope groups under person and have something like `person/:id/groups` and use a create action on that controller to handle this. It's not a hill I'd die on by any means, but it is more Railsy.

kasugaijin: Actually, looking at the code, we have `Organizations::ActivationsController` which is the closest existing controller we have to handle this action on `PersonGroup`. What do you think of passing the person ID to a create action on there, which creates the `PersonGroup` accordingly?

jmilljr24:
> Nice! I like it. This is a handy add, for sure. I don't think we need to jump to turbo right away. The redirect is a solid approach and for now works just fine.

There a just a lot of queries for the full page reload because this is index page when all that needs to be reloaded is the status symbol for a single persons group. It can wait until we decide how the different tables and their associated information gets organized.

> It does make me think about how all of these people pages could be harmonized in future, as well. One thought is a single people page with tabs for each page. There's also the thought of what the pages handle. This new page allows you to view, and add a role, but the other pages are where you deactivate each role. Further, the staff page allows you to change between admin and super admin. There is a probably a version of this that could be a single table that does everything at once...as it isn't automatically intuitive that I can enable on one, and have to deactivate on another...but it would be good to get real user feedback before diving into any grand changes on it. I'll look at the code, now.

I'm 100% on the same train of thought. From the UX perspective, having all groups displayed in one sections but individually toggled in another is not intuitive. I think this is worth exploring even before getting a user base as it is already an identified problem. I'm sure we wont get the perfect solution but we can do better in the mean time.

I went with the simple solution of just displaying the group status all in one place to make it clear each group is individually controllable. The (wrong) assumption could be made when I visit say the "adopter" table and click deactivate that all groups would be deactivated. Without returning to the all person index you wouldn't realize that only the adopter group was deactivated. I do like having the button to add a group as that shows there is a difference between adding a group and activation/deactivation. We do not allow removing a group but we don't tell the user that anywhere.

My thoughts on tabs and filtering different groups:

I think no matter what, for an individual person we need to show all groups and their status in one view to make it clear each groups activation is different. If we decide to display different information from the "all" tab vs the "adopter" tab, the person groups should all be displayed or none displayed for clarity within each tab.

My opinion on the best "first" solution is to have the same person information displayed for all tabs. The tabs would just be a filter of which group of people are displayed. So going off the table for "all" in this PR a persons info would have name, joined at, adopter, fosterer, staff. This would be the same info displayed if you clicked the "adopter" tab.

I would change the status icon from a static check mark or x to a toggleable button. This would eliminate the single toggle under "adopter" we currently have. The staff column could display the drop down admin/super_admin select. The changes would be in a new issue.

app/controllers/organizations/people_controller.rb
jmilljr24: You went through the same circle of thoughts I had. I settled on `PeopleController` to avoid creating confusion with `ActivationsController` without moving/refactoring that as well. I agree standard CRUD actions are better if they fit (which in this case they can)

`PersonGroupController` is a good choice as it is directly named with the model being updated and I'm happy do that.

I wonder if a better solution is to create `GroupManagementController`. We model the groups and their associations with the PORO's `activation` `group_member` and `staff` which are namespaced under `GroupRoleManagement` (this might be a good time to remove the "role" naming convention in this spot). They shouldn't need individual controllers at this point and they reflect the real world "model" we are dealing with not the join table this is kind of in a void.

This also aligns with the naming convention I used for the policy `GroupManagementPolicy`, which are also generally named to reflect the model they are dealing with.

Controller Actions:
Create - creates person_group association ( and creates the group if it doesn't yet exist for the org) via add_group method

Update - update `deactivated_at` on person_group

For changing the staff group - not sure if that would be its own action or if we can cleanly add it to the update action.

kasugaijin:
> I think no matter what, for an individual person we need to show all groups and their status in one view to make it clear each groups activation is different.

Yes I agree. 

> My opinion on the best "first" solution is to have the same person information displayed for all tabs. The tabs would just be a filter of which group of people are displayed. So going off the table for "all" in this PR a persons info would have name, joined at, adopter, fosterer, staff. This would be the same info displayed if you clicked the "adopter" tab.

I would change the status icon from a static check mark or x to a toggleable button. This would eliminate the single toggle under "adopter" we currently have. The staff column could display the drop down admin/super_admin select. The changes would be in a new issue.

Yeah I think we can put these into a ticket to capture the thoughts, and run it past a user to see what they would expect/what makes sense to them. I think a single view would be the easiest to comprehend...but I may be wrong here.

PR #1605, opened 2025-12-22
Send new super admin for Orgs to Change Password instead of receiving temp password

hcogden:
# 🔗 Issue
https://github.com/rubyforgood/homeward-tails/issues/1428

# ✍️ Description
We needed to change existing fuctionality so that when an organization is created, the super admin of the organization receives an email with a link to change their password instead of receiving a temp password. Once they change their password, they are redirected back to staff dashboard.

# 📷 Screenshots/Demos
I'd love to provide a screenshot/demo but I'm currently unsure on how to receive the email that the super admin should be receiving. The only email I'm able to receive when I complete the form on the app for signing up an org is the email sent to the site admins informing them to set up the org or whatever.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -78,6 +78,16 @@ class User < ApplicationRecord
     provider == "google_oauth2" && uid.present?
   end
 
+  def generate_token(attempts = 10)
+    retries = 0
+    update!(token: SecureRandom.urlsafe_base64(nil, false))
+    self.token
+  rescue ActiveRecord::RecordNotUnique
+    raise if (retries += 1) > attempts
+    Rails.logger.warn("random token, unlikely collision number #{retries}")
+    retry
+  end
+
   # Generally we want person data to stay scoped within its associated organization
   # This is a rare exception where we want to reduce friction for a user when they
   # join a new organization by prefilling the name values in the form.
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -78,16 +78,6 @@ class User < ApplicationRecord
     provider == "google_oauth2" && uid.present?
   end
 
-  def generate_token(attempts = 10)
-    retries = 0
-    update!(token: SecureRandom.urlsafe_base64(nil, false))
-    self.token
-  rescue ActiveRecord::RecordNotUnique
-    raise if (retries += 1) > attempts
-    Rails.logger.warn("random token, unlikely collision number #{retries}")
-    retry
-  end
-
   # Generally we want person data to stay scoped within its associated organization
   # This is a rare exception where we want to reduce friction for a user when they
   # join a new organization by prefilling the name values in the form.
--- a/app/views/organization_setup_mailer/invite_super_admin.html.erb
+++ b/app/views/organization_setup_mailer/invite_super_admin.html.erb
@@ -1,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
+  </head>
+  <body>
+    <h1>Welcome to Homeward Tails, <%= @user.first_name || 'friend' %>!</h1>
+    
+    <p>Thank you for registering <%= @organization.name %>.</p>
+    
+    <p>
+      You can access your organization's site at: <%= link_to "#{root_url}#{@organization.slug}", "#{root_url}#{@organization.slug}", target: "_blank" %>
+    </p>
+
+    <p>To get started, please create your password by clicking the link below. Once complete, you will be taken to the staff dashboard.</p>
+
+    <p><%= link_to 'Create my password', accept_invitation_url(@user, invitation_token: @token, script_name: "/#{@organization.slug}") %></p>
+
+    <p>If you have any questions, feel free to reach out.</p>
+
+    <p><strong>Homeward Tails Team</strong></p>
+  </body>
+</html>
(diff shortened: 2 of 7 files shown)


jmilljr24:
@hcogden ~We have a gem setup to view emails at [http://localhost:3000/letters](http://localhost:3000/letters). Not sure if you saw that or were looking somewhere else. [View Mail](https://github.com/rubyforgood/homeward-tails?tab=readme-ov-file#viewing-mail)~ See info in next comment.

Can you run `rails standard:fix` to fix the failing tests. [Linting](https://github.com/rubyforgood/homeward-tails?tab=readme-ov-file#-linting)

jmilljr24:
To create a new organization you need to run the service in `app/services/organizations/create_service.rb`. This will generate the email which you can then view at `/letters`.

If this doesn't trigger the email running `Organizations::CreateService.new.signal(args)` in the rails console try setting the send email method in the service to `deliver_now` ( you can commit that in your PR if it helps)

app/models/user.rb
```diff
@@ -78,6 +78,16 @@ def google_oauth_user?
...
 
+  def generate_token(attempts = 10)
+    retries = 0
+    update!(token: SecureRandom.urlsafe_base64(nil, false))
+    self.token
+  rescue ActiveRecord::RecordNotUnique
+    raise if (retries += 1) > attempts
+    Rails.logger.warn("random token, unlikely collision number #{retries}")
+    retry
+  end
```
jmilljr24: We are using devise in this app should we should be able to use `@user.invite!`

```suggestion

```

jmilljr24 (changes requested):
@hcogden Nice thoughts on this one!

Some more context on the app should point you in the right direction going forward. If you take a look at our `invitations_controller` our fosterers use the change password flow for new invites. This uses devise which I think we should be able to use for this PR as well.

The process for creating a new organization (with its initial super_admin) is through `create_service.rb` from the rails console.

`Organizations::CreateService.new.signal(args)`

This is where we need to change what email gets sent to this new super admin `@user.invite!`. The one gotcha is that we normally use our `organization_middleware` to set the slug and current organization which devise uses to build the invitation link. That doesn't get run when we use this service from the console so I think we need to build the url in the mailer. 

I think it's fine to create a new mailer just for this use case (vs. additional conditions to the normal invite). It should keep things simpler at the expense of a little duplication.

Open to other suggestions. Let me know your thoughts!

hcogden (commented):
> @hcogden ~We have a gem setup to view emails at [http://localhost:3000/letters](http://localhost:3000/letters). Not sure if you saw that or were looking somewhere else. [View Mail](https://github.com/rubyforgood/homeward-tails?tab=readme-ov-file#viewing-mail)~ See info in next comment.
> 
> Can you run `rails standard:fix` to fix the failing tests. [Linting](https://github.com/rubyforgood/homeward-tails?tab=readme-ov-file#-linting) 

Okay @jmilljr24 I've pushed up changes and I'm not seeing failing tests locally

hcogden (commented):
> @hcogden Nice thoughts on this one!
> 
> Some more context on the app should point you in the right direction going forward. If you take a look at our `invitations_controller` our fosterers use the change password flow for new invites. This uses devise which I think we should be able to use for this PR as well.
> 
> The process for creating a new organization (with its initial super_admin) is through `create_service.rb` from the rails console.
> 
> `Organizations::CreateService.new.signal(args)`
> 
> This is where we need to change what email gets sent to this new super admin `@user.invite!`. The one gotcha is that we normally use our `organization_middleware` to set the slug and current organization which devise uses to build the invitation link. That doesn't get run when we use this service from the console so I think we need to build the url in the mailer. 
> 
> I think it's fine to create a new mailer just for this use case (vs. additional conditions to the normal invite). It should keep things simpler at the expense of a little duplication.
> 
> Open to other suggestions. Let me know your thoughts! 

Okay @jmilljr24 I've made some changes. So I looked at the controller you recommended and saw how we use `@user.invite`, but yes, since we needed to build the url in the mailer, I created a new mailer and corresponding view for the letter. I also removed the unnecessary `generate_token` method call and removed it from the `User` model as well.

Only thing I'm still having trouble with is local testing, but I'll dig deeper into that with some debugging.

app/views/organization_setup_mailer/invite_super_admin.html.erb
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
+  </head>
+  <body>
+    <h1>Welcome to Homeward Tails, <%= @user.first_name || 'friend' %>!</h1>
```
jmilljr24:
```suggestion
    <h1>Welcome to Homeward Tails!</h1>
```
We don't have `first_name` on user so this was causing an error. Lets keep it simple and remove it. You should be able to generate the email now and view it at `letters` to test the password reset link.

jmilljr24 (changes requested):
@hcogden Looking good! Using the new mailer is great.

Once you make the change I request, can you try out the password reset link? I'm getting an error on my end. `Invitation token provided is not valid`.

Thanks!!

https://github.com/user-attachments/assets/1adebab1-e605-444e-9b1e-31720bd66091

jmilljr24:
@hcogden If you end up not figuring out the local tests and its unrelated to this issue, jump in the slack and we can troubleshoot from there.

hcogden:
> @hcogden Looking good! Using the new mailer is great.
> 
> Once you make the change I request, can you try out the password reset link? I'm getting an error on my end. `Invitation token provided is not valid`.
> 
> Thanks!!
> 
>  screenrecording-2025-12-29_16-09-29.mp4

Thanks so much for the video and suggestion on removing `first_name`! I committed that and I'll get with you in slack because yes I am still getting issues with demoing this locally 👍

